PR #22, opened 2025-04-18
test: Drain tool E2Es, UTs

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/test/e2e/drain_tool_test.go
+++ b/test/e2e/drain_tool_test.go
@@ -23,12 +23,15 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
+	clusterv1beta1 "github.com/kubefleet-dev/kubefleet/apis/cluster/v1beta1"
 	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
 	"github.com/kubefleet-dev/kubefleet/test/e2e/framework"
 	testutilseviction "github.com/kubefleet-dev/kubefleet/test/utils/eviction"
+	toolsutils "github.com/kubefleet-dev/kubefleet/tools/utils"
 )
 
 var _ = Describe("Drain cluster successfully", Ordered, Serial, func() {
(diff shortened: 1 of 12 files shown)


michaelawyu (commented):
Just 2 files left for review; apologies for the delay.

test/e2e/drain_tool_test.go
```diff
@@ -0,0 +1,389 @@
...
+	It("drain cluster using binary, should succeed", func() { runDrainClusterBinary(hubClusterName, memberCluster1EastProdName) })
+
+	It("should update member cluster with cordon taint", func() {
+		taintAddedActual := memberClusterCordonTaintAddedActual(memberCluster1EastProdName)
+		Eventually(taintAddedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to add cordon taint to member cluster")
+	})
+
+	It("should update drain cluster resource placement evictions status as expected", func() {
+		var fetchError error
+		drainEvictions, fetchError = fetchDrainEvictions()
```
michaelawyu: Hi Arvind! This step seems to be checking all evictions left on the API server? This way it would be prone to flakiness induced failures I am afraid. Could it be a bit more specific?

michaelawyu: We need to do at least a cluster name check as I understand -> a timestamp based approach would be better, just my initial thought.

michaelawyu: The same applies to other test specs.

test/e2e/setup_test.go
```diff
@@ -163,6 +164,11 @@ var (
 	}
 )
 
+var (
+	drainBinaryPath    = filepath.Join("../../", "hack", "tools", "bin", "kubectl-draincluster")
```
michaelawyu, line 169: Hi Arvind! Just a nit: would you like to add a before check that verifies if the binaries have been built before running the test specs?

michaelawyu: It's just a nice-to-have -> if you would like to add a TODO comment instead that's also good with me.

test/e2e/drain_tool_test.go
```diff
@@ -0,0 +1,399 @@
...
+
+	It("should update member cluster with cordon taint", func() {
+		taintAddedActual := memberClusterCordonTaintAddedActual(memberCluster1EastProdName)
+		Eventually(taintAddedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to add cordon taint to member cluster")
+	})
+
+	It("should update drain cluster resource placement evictions status as expected", func() {
+		var fetchError error
+		drainEvictions, fetchError = fetchDrainEvictions(crpName, memberCluster1EastProdName)
+		Eventually(fetchError).Should(Succeed(), "Failed to fetch drain evictions")
```
michaelawyu: Hi Arvind! This should be an Expect, rather than Eventually I assume.

michaelawyu: The same issue appears in a few other places I think.

test/e2e/drain_tool_test.go
```diff
@@ -0,0 +1,399 @@
...
+	Expect(err).ToNot(HaveOccurred(), "Uncordon command failed with error: %v", err)
+}
+
+func fetchDrainEvictions(crpName, clusterName string) ([]placementv1beta1.ClusterResourcePlacementEviction, error) {
+	var evictionList placementv1beta1.ClusterResourcePlacementEvictionList
+	if err := hubClient.List(ctx, &evictionList); err != nil {
+		return nil, err
+	}
+	var filteredDrainEvictions []placementv1beta1.ClusterResourcePlacementEviction
+	thirtySecondsAgo := time.Now().Add(-30 * time.Second)
```
michaelawyu: Hi Arvind! This definitely does not help with the flakiness situation I am afraid. The prev. comment about timestamp-based approach is something like:

a) we mark a timestamp at the beginning of the test spec.
b) when doing the list we filter out all evictions created before the creation timestamp.

Therefore the left evictions can be guaranteed to come from the draining tool as long as all eviction related test specs are running in serial.

michaelawyu: If it's too much to do at this moment I am fine with leaving only the cluster name based check and adding a TO-DO comment.

michaelawyu (approved):
LGTM

PR #25, opened 2025-04-22
feat: add metrics for updaterun

jwtty:

1. When updateRun is stuck waiting for a cluster update to finish or waiting for after-stage tasks to complete, we mark updateRun Progressing condition to false and add corresponding reacon.
2. Add updateRun status metrics based on updateRun condition. The metric contains 5 labels: name, generation, condition, status and reason. The value is the timestamp when the status metric is emitted.

Generated metrics look like:
```
# TYPE fleet_workload_update_run_status_last_timestamp_seconds gauge
fleet_workload_update_run_status_last_timestamp_seconds{condition="Succeeded",generation="1",name="example-run",reason="UpdateRunSucceeded",status="True"} 1.7452803474895709e+09
```

Fixes #

I have:

### DIFF
--- a/pkg/controllers/updaterun/execution.go
+++ b/pkg/controllers/updaterun/execution.go
@@ -44,6 +44,10 @@ var (
 	// stageUpdatingWaitTime is the time to wait before rechecking the stage update status.
 	// Put it as a variable for convenient testing.
 	stageUpdatingWaitTime = 60 * time.Second
+
+	// updateRunStuckThreshold is the time to wait on a single cluster update before marking update run as stuck.
+	// TODO(wantjian): make this configurable
+	updateRunStuckThreshold = 5 * time.Minute
 )
 
 // execute executes the update run by updating the clusters in the updating stage specified by updatingStageIndex.
@@ -55,8 +59,9 @@ func (r *Reconciler) execute(
 	updatingStageIndex int,
 	toBeUpdatedBindings, toBeDeletedBindings []*placementv1beta1.ClusterResourceBinding,
 ) (bool, time.Duration, error) {
-	// Mark the update run as started regardless of whether it's already marked.
-	markUpdateRunStarted(updateRun)
+	// Before the execute returns, mark the update run as started in memory if it's not marked as waiting or stuck.
+	// If the update run actually succeeds or fails, it will be remarked later in controller.go.
+	defer markUpdateRunStartedIfNotWaitingOrStuck(updateRun)
 
 	if updatingStageIndex < len(updateRun.Status.StagesStatus) {
 		updatingStage := &updateRun.Status.StagesStatus[updatingStageIndex]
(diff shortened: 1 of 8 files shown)


pkg/controllers/updaterun/execution.go
```diff
@@ -44,6 +44,10 @@ var (
 	// stageUpdatingWaitTime is the time to wait before rechecking the stage update status.
 	// Put it as a variable for convenient testing.
 	stageUpdatingWaitTime = 60 * time.Second
+
+	// updateRunStuckThreshold is the time to wait on a single cluster update before marking update run as stuck.
+	// TODO(wantjian): make this configurable
+	updateRunStuckThreshold = 5 * time.Minute
```
zhiying-lin, line 50: should we keep the default value consistent with the rollingUpdate?

pkg/controllers/updaterun/execution.go
```diff
@@ -55,8 +59,9 @@ func (r *Reconciler) execute(
 	updatingStageIndex int,
 	toBeUpdatedBindings, toBeDeletedBindings []*placementv1beta1.ClusterResourceBinding,
 ) (bool, time.Duration, error) {
-	// Mark the update run as started regardless of whether it's already marked.
-	markUpdateRunStarted(updateRun)
+	// Before the execute returns, mark the update run as started in memory if it's not marked as waiting or stuck.
+	// If the update run actually succeeds or fails, it will be remarked later in controller.go.
+	defer markUpdateRunStartedIfNotWaitingOrStuck(updateRun)
```
zhiying-lin: using "defer" statement looks weird to me. Marking updateRun as started is more natural at the beginning if it's not marked as waiting or stuck.

If error happens, the condition will be reset.

pkg/controllers/updaterun/execution.go
```diff
@@ -172,17 +177,27 @@ func (r *Reconciler) executeUpdatingStage(
 			markClusterUpdatingFailed(clusterStatus, updateRun.Generation, unexpectedErr.Error())
 			return 0, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
 		}
+
 		finished, updateErr := checkClusterUpdateResult(binding, clusterStatus, updatingStageStatus, updateRun)
 		if finished {
 			finishedClusterCount++
+			unmarkUpdateRunStuck(updateRun)
```
zhiying-lin: why not setting the progressing as false and saying that it's completed in the reason?

removing the condition looks weird here.

pkg/controllers/updaterun/execution.go
```diff
@@ -191,6 +206,7 @@ func (r *Reconciler) executeUpdatingStage(
 			return 0, err
 		}
 		if approved {
+			unmarkUpdateRunWaiting(updateRun)
```
zhiying-lin: ditto

pkg/controllers/updaterun/execution.go
jwtty: The default 1 min value is for the availability check of one single resource, while this one is for one cluster as a whole.

pkg/controllers/updaterun/execution.go
jwtty: The idea is, during the exection, we may mark it as stuck or waiting. After one round of execution, if it's not marked as stuck or waiting, that means the updateRun is still progressing and this call ensures that.

pkg/controllers/updaterun/execution.go
jwtty: Only sinlge cluster completes, not the updateRun. The updateRun should still be in Progressing status. Here if the cluster update completes, we are sure that it's not stuck, and we remove that.

pkg/controllers/updaterun/execution.go
jwtty: Same as above, the after-stage task completes, we are sure that it's not waiting anymore, and we remove that.

zhiying-lin (approved):

PR #29, opened 2025-04-22
feat: skip creation of empty work object

britaniar:

Fixes #

I have:


pkg/controllers/workgenerator/controller.go
```diff
@@ -520,13 +520,11 @@ func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding *fleetv1be
 		}
 		if len(simpleManifests) == 0 {
 			klog.V(2).InfoS("the snapshot contains no resource to apply either because of override or enveloped resources", "snapshot", klog.KObj(snapshot))
+		} else {
```
jwtty: A bit confused: from the comments, it seems that crp always depends on work to populate the status of the placement. And now, for CRP w/o any resources, we do not generate a work obj, what's gonna happen on CRP status?

test/e2e/placement_selecting_resources_test.go
```diff
@@ -46,55 +46,108 @@ var (
...
-	It("should remove controller finalizers from CRP", func() {
-		finalizerRemovedActual := allFinalizersExceptForCustomDeletionBlockerRemovedFromCRPActual(crpName)
-		Eventually(finalizerRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove controller finalizers from CRP %s", crpName)
+		AfterAll(func() {
+			By(fmt.Sprintf("garbage all things related to placement %s", crpName))
+			ensureCRPAndRelatedResourcesDeleted(crpName, allMemberClusters)
+		})
+
+		It("should update CRP status as expected", func() {
+			crpStatusUpdatedActual := crpStatusUpdatedActual(nil, allMemberClusterNames, nil, "0")
```
jwtty: Could you please add a comment here, what's the expected crp status? The CRP status is still available?

pkg/controllers/workgenerator/controller.go
britaniar: CRP Status will still be available. We are thinking of updating UX that will change CRP status.

test/e2e/placement_selecting_resources_test.go
ryanzhang-oss: same, I am not sure what to expect

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: without a work, the binding status will not be updated thus CRP reconcile not triggered, I am not entirely sure how it will work

britaniar: With the changes I implemented, `workUpdated` will be false [here](https://github.com/kubefleet-dev/kubefleet/blob/62f0d28bcefba3646155f039f059fc78993199d0/pkg/controllers/workgenerator/controller.go#L162C3-L162C79). 
Since it is false, we will [set the binding status](https://github.com/kubefleet-dev/kubefleet/blob/62f0d28bcefba3646155f039f059fc78993199d0/pkg/controllers/workgenerator/controller.go#L223C3-L226C45). 
Within the function `setBindingStatus`, we check to see the apply strategy and set the conditions [here](https://github.com/kubefleet-dev/kubefleet/blob/62f0d28bcefba3646155f039f059fc78993199d0/pkg/controllers/workgenerator/controller.go#L871C4-L878C3).
Since there are no works, the default for the condition will be set which is `True`. This is where we can update the UX interface to show whether the resources were selected.

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -162,6 +162,39 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+				ResourceSnapshotName: masterSnapshot.Name,
+				TargetCluster:        memberClusterName,
+			}
+			createClusterResourceBinding(&binding, spec)
+			// check the work is not created since the binding state is not bound
+			work := placementv1beta1.Work{}
+			Consistently(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: fmt.Sprintf(placementv1beta1.FirstWorkNameFmt, testCRPName), Namespace: memberClusterNamespaceName}, &work)
+				return errors.IsNotFound(err)
+			}, duration, interval).Should(BeTrue(), "controller should not create work in hub cluster until all resources are created")
```
jwtty: What does this message mean?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -162,6 +162,39 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+			Expect(len(binding.Finalizers)).Should(Equal(0))
+			// flip the binding state to bound
+			binding.Spec.State = placementv1beta1.BindingStateBound
+			Expect(k8sClient.Update(ctx, binding)).Should(Succeed())
+			updateRolloutStartedGeneration(&binding)
+			// check the work is not created since no resources are selected
+			Consistently(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: fmt.Sprintf(placementv1beta1.FirstWorkNameFmt, testCRPName), Namespace: memberClusterNamespaceName}, &work)
+				return errors.IsNotFound(err)
+			}, duration, interval).Should(BeTrue(), "controller should not create work in hub cluster until all resources are created")
```
jwtty: ditto

pkg/controllers/workgenerator/controller.go
jwtty: I can see two scenarios where `workUpdated` can be true in this case: 
1. With an older version of hub-agent without your change, an empty work will be generated and after upgrading the image with your change, that empty work object will be deleted by code below on line 546-562 and workUpdated becomes true.
2. When there are multiple resourceSnapshots generated or there's a configmap-wrapped resource, there will be other work obj generated and thus workUpdated also becomes true.
Could you please verify these two cases and add tests?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -162,6 +162,39 @@ var _ = Describe("Test Work Generator Controller", func() {
 			}, timeout, interval).Should(BeTrue(), "Get() member cluster, want not found")
 		})
 
+		It("Should not create work when no resources are selected", func() {
+			// create master resource snapshot with 0 resources
```
jwtty:
```suggestion
			// Create master resource snapshot with 0 resources.
```

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -162,6 +162,39 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+			masterSnapshot := generateResourceSnapshot(1, 1, 0, [][]byte{})
+			Expect(k8sClient.Create(ctx, masterSnapshot)).Should(Succeed())
+			// create a scheduled binding
```
jwtty:
```suggestion
			// Create a scheduled binding.
```

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -162,6 +162,39 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+			spec := placementv1beta1.ResourceBindingSpec{
+				State:                placementv1beta1.BindingStateScheduled,
+				ResourceSnapshotName: masterSnapshot.Name,
+				TargetCluster:        memberClusterName,
+			}
+			createClusterResourceBinding(&binding, spec)
+			// check the work is not created since the binding state is not bound
```
jwtty:
```suggestion
			// Check the work is not created since the binding state is not bound.
```

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -162,6 +162,39 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+			work := placementv1beta1.Work{}
+			Consistently(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: fmt.Sprintf(placementv1beta1.FirstWorkNameFmt, testCRPName), Namespace: memberClusterNamespaceName}, &work)
+				return errors.IsNotFound(err)
+			}, duration, interval).Should(BeTrue(), "controller should not create work in hub cluster until all resources are created")
+			// binding should not have any finalizers
```
jwtty:
```suggestion
			// Binding should not have any finalizers.
```

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -162,6 +162,39 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)).Should(Succeed())
+			Expect(len(binding.Finalizers)).Should(Equal(0))
+			// flip the binding state to bound
```
jwtty:
```suggestion
			// Flip the binding state to bound
```

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -162,6 +162,39 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+			binding.Spec.State = placementv1beta1.BindingStateBound
+			Expect(k8sClient.Update(ctx, binding)).Should(Succeed())
+			updateRolloutStartedGeneration(&binding)
+			// check the work is not created since no resources are selected
```
jwtty:
```suggestion
			// Check the work is not created since no resources are selected.
```

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -162,6 +162,39 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+			Consistently(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: fmt.Sprintf(placementv1beta1.FirstWorkNameFmt, testCRPName), Namespace: memberClusterNamespaceName}, &work)
+				return errors.IsNotFound(err)
+			}, duration, interval).Should(BeTrue(), "controller should not create work in hub cluster until all resources are created")
+			// check the binding status is available
```
jwtty:
```suggestion
			// Check the binding status is available.
```

zhiying-lin (commented):
@britaniar could you please update the PR description to show me the final crp status condition when no resources are selected? We need to agree on the user experience/api first . I remember ryan mentioned that cx is confused by the applied/available true status.  btw, wantong is working on updating the CRP conditions and code may have conflicts.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -232,7 +232,18 @@ func setCRPConditions(
 			shouldSkipRestCondTypes = true
 		case rpsSetCondTypeCounter[i][condition.FalseConditionStatus] > 0:
 			// There is at least one False condition of the given type being set on the per cluster placement statuses.
-			crp.SetConditions(i.FalseClusterResourcePlacementCondition(crp.Generation, rpsSetCondTypeCounter[i][condition.FalseConditionStatus]))
+			if len(crp.Status.SelectedResources) == 0 && i == condition.AppliedCondition {
```
jwtty: The current behavior is when no resource is selected, the work obj still gets created, and all clusters as well as the crp shows Applied/Available to true. We are changing this behavior. Is this what we want?

pkg/controllers/workgenerator/controller.go
```diff
@@ -593,6 +590,17 @@ func (r *Reconciler) syncApplyStrategy(
 	return true, nil
 }
 
+func workObjectCreationNeeded(snapshot *fleetv1beta1.ClusterResourceSnapshot, simpleManifests []fleetv1beta1.Manifest, newWork []*fleetv1beta1.Work) bool {
+	switch {
+	case len(simpleManifests) == 0 && len(newWork) == 0:
+		klog.V(2).InfoS("No work to create for this snapshot", "snapshot", klog.KObj(snapshot))
+		return false
+	case len(simpleManifests) == 0:
+		klog.V(2).InfoS("The snapshot contains no resource to apply due to overrides or enveloped resources", "snapshot", klog.KObj(snapshot))
```
jwtty, line 612: If we want to report error instead of reporting allAvailable, can we return error here?

pkg/controllers/clusterresourceplacement/placement_status.go
britaniar: Yes, we want to change it from it being Applied/Available is true because the cx gets confused.

zhiying-lin (commented):
i recall that why we treat the applied & available originally as true. There will be a case, cx first applied a namespace, and then he/she wants to delete this namespace in all members but keeps the crp. 

CRP still needs to apply the resources (delete the resources). I'm thinking of changing the applied & available reason to something else, instead of marking them as false, and it also simplifies our alert.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -249,6 +250,15 @@ func setCRPConditions(
 					cond.Reason = condition.OverrideNotSpecifiedReason
 					cond.Message = "No override rules are configured for the selected resources"
 				}
+			case condition.AppliedCondition:
+				if len(crp.Status.SelectedResources) == 0 {
+					cond.Message = "Previous resources are deleted. No new resources are selected for the placement"
```
zhiying-lin, line 255: ``
```suggestion
					cond.Message = "No new resources are selected for the placement and deleted applied resources if exists"
```

zhiying-lin: change the reason to "NoResourcesSelectedReason" too?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -249,6 +250,15 @@ func setCRPConditions(
...
+				}
+			case condition.AvailableCondition:
+				if len(crp.Status.SelectedResources) == 0 {
+					cond.Reason = condition.NoResourcesSelectedReason
+					cond.Message = "No new resources are selected for the placement. Please update crp to select new resources"
```
zhiying-lin, line 260:
```suggestion
					cond.Message = "No new resources are applied for the placement"
```
cx can create the resources instead of updating crp.

pkg/controllers/workgenerator/controller.go
```diff
@@ -155,11 +155,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 
 	workUpdated := false
 	overrideSucceeded := false
+	resourcesSelected := true
 	// list all the corresponding works
 	works, syncErr := r.listAllWorksAssociated(ctx, &resourceBinding)
 	if syncErr == nil {
 		// generate and apply the workUpdated works if we have all the works
-		overrideSucceeded, workUpdated, syncErr = r.syncAllWork(ctx, &resourceBinding, works, &cluster)
+		overrideSucceeded, workUpdated, resourcesSelected, syncErr = r.syncAllWork(ctx, &resourceBinding, works, &cluster)
```
zhiying-lin, line 163: Could you please test drift detection/ report diff feature when no resources are selected? i'm not sure if we reply on this empty work.

zhiying-lin (commented):
Let's split the PRs into 6 smaller PRs and add tests to catch the regression:

1. Fix the bug to track the deleting works, set the binding applied condition is false when work is still in the deleting state. https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/controllers/workgenerator/controller.go#L1010
2. Test to see what's the behavior our drift/reportDiff feature when no resource are selected. Add tests if missing.
3. Add e2e tests when resource is not selected
4. update code to not create work when resource is not selected
5. update the work binding condition message & reason when resource is not selected
6. update the crp condition message & reason when resource is not selected

PR #32, opened 2025-04-25
fix: block cluster-scoped resources from being wrapped in an envelope

michaelawyu:

This PR blocks cluster-scoped resources from being wrapped in an envelope to address a security concern.

I have:

- [x] Unit tests

### DIFF
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -1280,8 +1280,11 @@ func extractResFromConfigMap(uConfigMap *unstructured.Unstructured) ([]fleetv1be
 			klog.ErrorS(unMarshallErr, "manifest has invalid content", "manifestKey", key, "envelopeResource", klog.KObj(uConfigMap))
 			return nil, fmt.Errorf("the object with manifest key `%s` in envelope config `%s` is malformatted, err: %w", key, klog.KObj(uConfigMap), unMarshallErr)
 		}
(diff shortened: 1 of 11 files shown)


pkg/controllers/workgenerator/controller.go
```diff
@@ -1280,8 +1280,11 @@ func extractResFromConfigMap(uConfigMap *unstructured.Unstructured) ([]fleetv1be
 			klog.ErrorS(unMarshallErr, "manifest has invalid content", "manifestKey", key, "envelopeResource", klog.KObj(uConfigMap))
 			return nil, fmt.Errorf("the object with manifest key `%s` in envelope config `%s` is malformatted, err: %w", key, klog.KObj(uConfigMap), unMarshallErr)
 		}
+		if len(uManifest.GetNamespace()) == 0 {
+			// Block cluster-scoped resources.
+			return nil, fmt.Errorf("cannot wrap cluster-scoped resource %s in the envelope %s", uManifest.GetName(), klog.KObj(uConfigMap))
+		}
 		if len(uManifest.GetNamespace()) != 0 && uManifest.GetNamespace() != configMap.Namespace {
```
ryanzhang-oss, line 1287: with the new condition added, this is just
```suggestion
		if  uManifest.GetNamespace() != configMap.Namespace {
```

ryanzhang-oss (approved):

PR #33, opened 2025-04-25
interface: support multiple resourceSnapshot versions across clusters

jwtty:

This PR is the interface update to support showing multiple resourceSnapshot versions across targeted clusters on CRP.

Current behavior:
In current CRP status, there's `crp.status.ObservedResourceIndex` that is always set to the index of the latest resourceSnapshot currently exists on the hub cluster as current rollingUpdate rollout strategy only supports rolling to the latest resourceSnapshot version. The `crp.status.Conditions` (CRP status as a whole) as well as `crp.status.placementStatuses` (each targeted member clsuter status) are both based on this index.

This has several cons:
1. With staged update run enabled, users are able to specify the resourceSnapshot version to rollout. When the version is not the latest, CRP status always shows as "RolloutNotStarted", because resourceSnapshot version does not match.
2. During the rollout process, either with updateRun or rollingUpdate, different clusters may observe different resourceSnapshot versions, e.g. when rolling is waiting for resource to be available or just gets stuck due to some failure, current CRP always shows as "RolloutNotStarted", which does not accurately reflect the status. An example with manifestNotAvailable yet:
```
status:
  conditions:
  - lastTransitionTime: "2025-04-25T21:07:55Z"
    message: found all cluster needed as specified by the scheduling policy, found
      2 cluster(s)
    observedGeneration: 2
    reason: SchedulingPolicyFulfilled
    status: "True"
    type: ClusterResourcePlacementScheduled
  - lastTransitionTime: "2025-04-25T22:17:12Z"
    message: The rollout is being blocked by the rollout strategy in 1 cluster(s)
    observedGeneration: 2
    reason: RolloutNotStartedYet
    status: "False"
    type: ClusterResourcePlacementRolloutStarted
  observedResourceIndex: "3"
  placementStatuses:
  - clusterName: member1
    conditions:
    - lastTransitionTime: "2025-04-25T21:07:55Z"
      message: 'Successfully scheduled resources for placement in "member1" (affinity
        score: 0, topology spread score: 0): picked by scheduling policy'
      observedGeneration: 2
      reason: Scheduled
      status: "True"
      type: Scheduled
    - lastTransitionTime: "2025-04-25T22:17:12Z"
      message: Detected the new changes on the resources and started the rollout process
      observedGeneration: 2
      reason: RolloutStarted
      status: "True"
      type: RolloutStarted
    - lastTransitionTime: "2025-04-25T22:17:12Z"
      message: No override rules are configured for the selected resources
      observedGeneration: 2
      reason: NoOverrideSpecified
      status: "True"
      type: Overridden
    - lastTransitionTime: "2025-04-25T22:17:12Z"
      message: All of the works are synchronized to the latest
      observedGeneration: 2
      reason: AllWorkSynced
      status: "True"
      type: WorkSynchronized
    - lastTransitionTime: "2025-04-25T22:17:13Z"
      message: All corresponding work objects are applied
      observedGeneration: 2
      reason: AllWorkHaveBeenApplied
      status: "True"
      type: Applied
    - lastTransitionTime: "2025-04-25T22:17:13Z"
      message: Work object example-placement-work is not yet available
      observedGeneration: 2
      reason: NotAllWorkAreAvailable
      status: "False"
      type: Available
    failedPlacements:
    - condition:
        lastTransitionTime: "2025-04-25T22:17:13Z"
        message: Manifest is not yet available; Fleet will check again later
        reason: ManifestNotAvailableYet
        status: "False"
        type: Available
      kind: Service
      name: nginx
      namespace: test-namespace
      version: v1
  - clusterName: member2
    conditions:
    - lastTransitionTime: "2025-04-25T21:07:55Z"
      message: 'Successfully scheduled resources for placement in "member2" (affinity
        score: 0, topology spread score: 0): picked by scheduling policy'
      observedGeneration: 2
      reason: Scheduled
      status: "True"
      type: Scheduled
    - lastTransitionTime: "2025-04-25T22:17:12Z"
      message: The rollout is being blocked by the rollout strategy
      observedGeneration: 2
      reason: RolloutNotStartedYet
      status: "False"
      type: RolloutStarted
```
In above example, rollout does happen to member1, it's just waiting for the resource to become available. But in the CRP condition, it still shows "RolloutNotStarted". And in member2, we actually loses more information that it currently is still running on the earlier version.

#### Proposed API
The proposed solution is to modify the definition of `crp.status.ObservedResourceIndex` from the index of latest resourceSnapshot observed on hub cluster to the index of the latest resourceSnapshot observed across all targeted memberclusters. For example, when the currently latest index is 4 on hub cluster but 2, 2, 3 on member clusters respectively, current `crp.statusObservedResourceIndex` is 4, but will be 3 with the proposed changed. This does not break existing rollingUpdate strategy as all member clusters will be updated to 4 eventually and crp will reach the final terminate state. And with updateRun or other external rollout strategy, users can find it more accurate to reflect the current CRP status: CRP now shows the version that user rolls out, not the default-latest version.

There are two options to handle per-cluster status:
1. Each cluster still reports its status based on `crp.status.ObservedResourceIndex`. This aligns with current API.
2. Add an `ObservedResourceIndex` inside per-cluster placement status. And each cluster's status is based on this version. For example, when current version on each cluster is 2, 3, 3 respectively, the first cluster shows RolloutNotStarted or RolloutUnknown as it's not being updated to version 3 yet, while with this new option, it's RolloutStarted and observedResourceIndex is 2. This somehow breaks current API but it displays more accurate status.

With the proposed solution, we can easily support different rollout strategy, show cluster status more accurately, and can support rollout back in the future.

#### Revised Apr 28:

When rollout strategy type is set to `External`, rollout is managed by an external controller and crp is not aware of the actual rollout plan/target. In this case, setting `ObservedResourceIndex` and `RolloutStarted/Applied/Available` conditions in the crp whole status does not make sense and can cause confusion. In this case, we only demonstrate the per-cluster observedResourceIndex and conditions in `crp.statusResoucePlacementStatuses` array. CRP condition will only show `Scheduled` and `RolloutStarted Unknown`.

<b>
#### Revised Apr 29:

Per offline discussion, we have decided that:

* Per-cluster resourcePlacement status:
  For both `RollingUpdate` and `External` rollout strategies, `ObservedResourceIndex` and conditions including `RolloutStarted/Overridden/WorkSynchronized/Applied/Available` should reflect the current observed resource index of the member cluster and be consistent with the corresponding `clusterResourceBinding` status.
* CRP status:
  CRP status serves as an aggregate of the per-cluster resourcePlacement status.
  * For `RollingUpdate` strategy:
     `ObservedResourceIndex` is the default-latest resourceSnapshot index and the conditions are based on it.
  * For `External` strategy:
    `ObservedResourceIndex` is not empty only if all the member clusters observe the same resource index.
    When `ObservedResourceIndex` is empty, all conditions except `Scheduled` should be in `Unknown` state.
    When `ObservedResourceIndex` is not empty, conditions show the aggregation of the clusters' status:
    * If any one of the cluster is `Unknown`, aggregated state is `Unknown`.
    * Else if any one of the cluster is `False`, aggregated state is `False`.
    * If all the clusters are `True`, aggregated state is `True`.
 </b>

Fixes #

I have:

### DIFF
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -822,12 +822,10 @@ type ClusterResourcePlacementStatus struct {
 	// Each snapshot has a different resource index.
 	// One resource snapshot can contain multiple clusterResourceSnapshots CRs in order to store large amount of resources.
 	// To get clusterResourceSnapshot of a given resource index, use the following command:
-	// `kubectl get ClusterResourceSnapshot --selector=kubernetes-fleet.io/resource-index=$ObservedResourceIndex `
-	// ObservedResourceIndex is the resource index that the conditions in the ClusterResourcePlacementStatus observe.
-	// For example, a condition of `ClusterResourcePlacementWorkSynchronized` type
-	// is observing the synchronization status of the resource snapshot with the resource index $ObservedResourceIndex.
+	// `kubectl get ClusterResourceSnapshot --selector=kubernetes-fleet.io/resource-index=$ObservedResourceIndex`
 	// If the rollout strategy type is `RollingUpdate`, `ObservedResourceIndex` is the default-latest resource snapshot index.
-	// If the rollout strategy type is `External`, rollout and version control are managed by an external controller, and this field remains empty.
+	// If the rollout strategy type is `External`, rollout and version control are managed by an external controller,
+	// and this field is not empty only if all targeted clusters observe the same resource index in `PlacementStatuses`.
 	// +kubebuilder:validation:Optional
 	ObservedResourceIndex string `json:"observedResourceIndex,omitempty"`
 
@@ -846,6 +844,11 @@ type ClusterResourcePlacementStatus struct {
 	// +listMapKey=type
 
 	// Conditions is an array of current observed conditions for ClusterResourcePlacement.
+	// All conditions except `ClusterResourcePlacementScheduled` correspond to the resource snapshot at the index specified by `ObservedResourceIndex`.
+	// For example, a condition of `ClusterResourcePlacementWorkSynchronized` type
+	// is observing the synchronization status of the resource snapshot with index `ObservedResourceIndex`.
+	// If the rollout strategy type is `External`, and `ObservedResourceIndex` is unset due to clusters reporting different resource indices,
+	// conditions except `ClusterResourcePlacementScheduled` will be empty or set to Unknown.
 	// +kubebuilder:validation:Optional
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
@@ -958,13 +961,13 @@ type ResourcePlacementStatus struct {
 	DiffedPlacements []DiffedResourcePlacement `json:"diffedPlacements,omitempty"`
 
 	// ObservedResourceIndex is the index of the resource snapshot that is currently being rolled out to the given cluster.
-	// During rollout, depending on the rollout strategy, clusters may observe different resource snapshot indices.
-	// ObservedResourceIndex is the resource snapshot index observed by the conditions in the ResourcePlacementStatus.
 	// This field is only meaningful if the `ClusterName` is not empty.
 	// +kubebuilder:validation:Optional
 	ObservedResourceIndex string `json:"observedResourceIndex,omitempty"`
 
-	// Conditions is an array of current observed conditions for ResourcePlacementStatus.
+	// Conditions is an array of current observed conditions on the cluster.
+	// Each condition corresponds to the resource snapshot at the index specified by `ObservedResourceIndex`.
+	// For example, the condition of type `RolloutStarted` is observing the rollout status of the resource snapshot with index `ObservedResourceIndex`.
 	// +kubebuilder:validation:Optional
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -822,6 +822,9 @@ type ClusterResourcePlacementStatus struct {
 	// One resource snapshot can contain multiple clusterResourceSnapshots CRs in order to store large amount of resources.
 	// To get clusterResourceSnapshot of a given resource index, use the following command:
 	// `kubectl get ClusterResourceSnapshot --selector=kubernetes-fleet.io/resource-index=$ObservedResourceIndex `
+	// During rollout, clusters may run different versions of the resource snapshot concurrently.
+	// In this case, ObservedResourceIndex represents the index of the latest resource snapshot installed across all clusters.
+	// Note that this may differ from the index of the latest resource snapshot available in the hub cluster depending on the rollout strategy.
 	// ObservedResourceIndex is the resource index that the conditions in the ClusterResourcePlacementStatus observe.
 	// For example, a condition of `ClusterResourcePlacementWorkSynchronized` type
 	// is observing the synchronization status of the resource snapshot with the resource index $ObservedResourceIndex.
@@ -953,6 +956,13 @@ type ResourcePlacementStatus struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:MaxItems=100
 	DiffedPlacements []DiffedResourcePlacement `json:"diffedPlacements,omitempty"`
+
+	// ObservedResourceIndex is the index of the resource snapshot that is currently rolled out on the given cluster.
+	// During rollout, depending on the rollout strategy, clusters may observe different resource indices.
+	// ObservedResourceIndex is the resource snapshot index observed by the conditions in the ResourcePlacementStatus.
+	// This field is only meaningful if the `ClusterName` is not empty.
+	// +kubebuilder:validation:Optional
+	ObservedResourceIndex string `json:"observedResourceIndex,omitempty"`
 
 	// Conditions is an array of current observed conditions for ResourcePlacementStatus.
 	// +kubebuilder:validation:Optional
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -958,7 +958,7 @@ type ResourcePlacementStatus struct {
 	DiffedPlacements []DiffedResourcePlacement `json:"diffedPlacements,omitempty"`
 
 	// ObservedResourceIndex is the index of the resource snapshot that is currently being rolled out to the given cluster.
-	// During rollout, depending on the rollout strategy, clusters may observe different resource indices.
+	// During rollout, depending on the rollout strategy, clusters may observe different resource snapshot indices.
 	// ObservedResourceIndex is the resource snapshot index observed by the conditions in the ResourcePlacementStatus.
 	// This field is only meaningful if the `ClusterName` is not empty.
 	// +kubebuilder:validation:Optional
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -913,6 +913,11 @@ type ResourcePlacementStatus struct {
 	// +kubebuilder:validation:Optional
 	ClusterName string `json:"clusterName,omitempty"`
 
+	// ObservedResourceIndex is the index of the resource snapshot that is currently being rolled out to the given cluster.
+	// This field is only meaningful if the `ClusterName` is not empty.
+	// +kubebuilder:validation:Optional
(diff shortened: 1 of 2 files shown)


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -954,6 +957,13 @@ type ResourcePlacementStatus struct {
 	// +kubebuilder:validation:MaxItems=100
 	DiffedPlacements []DiffedResourcePlacement `json:"diffedPlacements,omitempty"`
 
+	// ObservedResourceIndex is the index of the resource snapshot that is currently rolled out on the given cluster.
```
ryanzhang-oss:
```suggestion
	// ObservedResourceIndex is the index of the resource snapshot that is currently being rolled out to the given cluster.
```

ryanzhang-oss (approved):

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -954,6 +957,13 @@ type ResourcePlacementStatus struct {
 	// +kubebuilder:validation:MaxItems=100
 	DiffedPlacements []DiffedResourcePlacement `json:"diffedPlacements,omitempty"`
 
+	// ObservedResourceIndex is the index of the resource snapshot that is currently being rolled out to the given cluster.
+	// During rollout, depending on the rollout strategy, clusters may observe different resource indices.
+	// ObservedResourceIndex is the resource snapshot index observed by the conditions in the ResourcePlacementStatus.
+	// This field is only meaningful if the `ClusterName` is not empty.
+	// +kubebuilder:validation:Optional
+	ObservedResourceIndex string `json:"observedResourceIndex,omitempty"`
```
jwtty, line 965: Another possible solution is that, we do not add this `ObservedResourceIndex` in each member cluster, while we only change the `ObservedResourceIndex` to the latest resourceSnapshot index version installed on the member clusters. This does not break current API.

circy9 (commented):
Assuming members are all running snapshot v1 now, if snapshot v2 is discovered and rolled out to member1 (still waiting for availability) and not to member2 yet, how does the per-CRP and per-member status look like?

jwtty:
> Assuming members are all running snapshot v1 now, if snapshot v2 is discovered and rolled out to member1 (still waiting for availability) and not to member2 yet, how does the per-CRP and per-member status look like?

It's the exact content as the example I show in the PR description. CRP whole status shows `RolloutStarted` false as our API definition by this type is `RolloutStarted set to True when rollout started on all clusters` (this is a bit weird, true). And Per-member status is: for member1, it has `RolloutStarted`, ..., `NotAvailableYet`; while member2, it shows `RolloutNotStartedYet`.

apis/placement/v1beta1/clusterresourceplacement_types.go
jwtty: For per-cluster placement status, we have two options:
1. Be compatible with existing behavior:
   With `RollingUpdate` strategy, this per-cluster `ObservedResourceIndex` is always set to the latest resourceSnapshot index, and if rollout has not reached this member cluster yet, the condition for this cluster is `RolloutNotStartedYet`.
   With `External` strategy, this per-cluster `ObservedResourceIndex` is set to whatever the cluster currently has. If the rollout has not reached this member cluster, the condition for this cluster can be `RolloutStarted` all the way to `Available` on the older version. 
   **Pros**: Compatible with current behavior. It's clear that cluster does not new version yet is in the `RolloutNotStarted` condition.
   **Cons**: Need to maintain two series of logic and two versions of API explanation.
2. Kind of breaking current behavior:
    Do not distinguish between the rollout strategy. This per-cluster `ObservedResourceIndex` is set to whatever the cluster currently has. If the rollout has not reached this member cluster, the condition for this cluster can be `RolloutStarted` all the way to `Available` on the older version. 
    **Pros**: Behavior aligned between rollout strategies. We posted more details about each cluster. What version the cluster is running on.
    **Cons**: Break current RollingUpdate status.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -954,6 +957,13 @@ type ResourcePlacementStatus struct {
 	// +kubebuilder:validation:MaxItems=100
 	DiffedPlacements []DiffedResourcePlacement `json:"diffedPlacements,omitempty"`
 
+	// ObservedResourceIndex is the index of the resource snapshot that is currently being rolled out to the given cluster.
+	// During rollout, depending on the rollout strategy, clusters may observe different resource indices.
```
ryanzhang-oss: what does this mean?

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -954,6 +957,13 @@ type ResourcePlacementStatus struct {
...
+	// ObservedResourceIndex is the resource snapshot index observed by the conditions in the ResourcePlacementStatus.
```
ryanzhang-oss: not clear what this means

apis/placement/v1beta1/clusterresourceplacement_types.go
jwtty: For example, with concurrent updateRuns, different clusters in different stages can have different resource snapshot indices, one with version 1 and one with version 2 for example.

apis/placement/v1beta1/clusterresourceplacement_types.go
jwtty: I was following the wording on line 826. It means the conditions of the member cluster are based on this `ObservedResourceIndex`, whether this version of data is rolled out, applied, and available.

jwtty: But the question is, for rollingUpdate, should this per-cluster `ObservedResourceIndex` be the same as the default-latest CRP resource snapshot index or just whatever exists currently on the member cluster.

jwtty: For updateRun, it will be whichever version that currently exists on the member cluster.

apis/placement/v1beta1/clusterresourceplacement_types.go
jwtty: Reached an agreement through offline discussion, see PR description.

apis/placement/v1beta1/clusterresourceplacement_types.go
jwtty: Updated.

apis/placement/v1beta1/clusterresourceplacement_types.go
jwtty: Updated, hope it's clearer now.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -954,7 +960,14 @@ type ResourcePlacementStatus struct {
 	// +kubebuilder:validation:MaxItems=100
 	DiffedPlacements []DiffedResourcePlacement `json:"diffedPlacements,omitempty"`
 
-	// Conditions is an array of current observed conditions for ResourcePlacementStatus.
+	// ObservedResourceIndex is the index of the resource snapshot that is currently being rolled out to the given cluster.
+	// This field is only meaningful if the `ClusterName` is not empty.
+	// +kubebuilder:validation:Optional
+	ObservedResourceIndex string `json:"observedResourceIndex,omitempty"`
```
circy9: This field should appear before the overrides snapshots as the overrides are on top of this given resource snapshots. The union of this field and the override snapshot fields are part of the workload spec for this cluster.

clusterName -> ObservedResourceIndex -> ApplicableResourceOverrides

circy9 (commented):
Based on offline discussions, we agree to do the following:
 
1. For all strategies, per-cluster observedresourceindex and RolloutStarted/Overridden/WorkSynchronized/Applied/Available should reflect the current observedresourceindex status of the member cluster (consistent with the corresponding binding status).
 
2. For external rollout strategy, per-CRP observedresourceindex and RolloutStarted/Overridden/WorkSynchronized/Applied/Available should be an aggregate of the per-cluster status:
- observedresourceindex is non-empty only if all the clusters have the same observedresourceindex.
- When observedresourceindex is empty, all the conditions will be unknonw.
- When observedresourceindex is set,
  - If any one of the cluster is unknown, the aggregated condition will be unknown.
  - (Else) if any one of the cluster is false, the aggregated condition is false.
  - If all of the clusters are true, the aggregated condition is true.
  
3. For rollingupdate strategy, per-CRP observedresourceindex is the lastest resourceIndex and the conditions are for the latest.

apis/placement/v1beta1/clusterresourceplacement_types.go
jwtty: Thanks, fixed.

ryanzhang-oss (approved):

PR #35, opened 2025-04-28
interface: add the new CR-based envelope API [Experimental]

michaelawyu:

This PR adds an experimental design of the proposed CR-based envelope API.

I have:

N/A

* Please see the comments for additional details.


apis/placement/v1alpha1/envelope_types.go
```diff
@@ -0,0 +1,129 @@
...
+// EnvelopeSpec helps wrap resources for placement.
+type EnvelopeSpec struct {
+	// A map of wrapped manifests.
+	//
+	// Each manifest is uniquely identified by a string key, typically a filename that represents
+	// the manifest.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinProperties=1
+	// +kubebuilder:validation:MaxProperties=50
+	Manifests map[string]Manifest `json:"manifests"`
```
michaelawyu, line 49: Note: a map is used here to make sure that each manifest is uniquely identified with a key, which can later be used in status reporting.

apis/placement/v1alpha1/envelope_types.go
```diff
@@ -0,0 +1,129 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ClusterResourceEnvelope wraps cluster-scoped resources for placement.
+type ClusterResourceEnvelope struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// The desired state of ClusterResourceEnvelope.
+	// +kubebuilder:validation:Required
+	Spec EnvelopeSpec `json:"spec"`
```
michaelawyu, line 37: Note: we keep the `spec` field here just as a common practice; if necessary we could expose the inner object directly to reduce nesting.

apis/placement/v1alpha1/envelope_types.go
```diff
@@ -0,0 +1,129 @@
...
+	// Each manifest is uniquely identified by a string key, typically a filename that represents
+	// the manifest.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinProperties=1
+	// +kubebuilder:validation:MaxProperties=50
+	Manifests map[string]Manifest `json:"manifests"`
+}
+
+// Manifest is a wrapped resource.
+type Manifest struct {
```
michaelawyu, line 53: We could put the YAML content directly as the map value; it is now wrapped in a struct just in case we might need to add some resource-specific options in the future.

michaelawyu: For example, in earlier discussions we have received requests regarding the possibility of doing API server validation on wrapped resources as an extra layer of defense -> if implemented, we might need to have a `bypassValidation` option here to allow wrapping of resources that are expected to fail API server validation (e.g., the corresponding API is only available on the member cluster side, or a mutating webhook is expected to modify the object on the member cluster side).

apis/placement/v1alpha1/envelope_types.go
```diff
@@ -0,0 +1,129 @@
...
+	// The resource data.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:EmbeddedResource
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Data runtime.RawExtension `json:"data"`
+}
+
+// ClusterResourceEnvelopeStatus is the observed status of a ClusterResourceEnvelope.
+type ClusterResourceEnvelopeStatus struct {
```
michaelawyu: The status fields are added just in case in the future we might want to implement additional checks for wrapped resources. In the initial stage this might not be necessary, and it should be safe for us to drop these fields at first.

ryanzhang-oss: let's not add status now as we can always add them later

michaelawyu: Sure, I will drop the status field 👌

PR #48, opened 2025-05-02
feat: add priority queue to reconciling work in apply_controller

sudheerv:

[isssue#46]

add priority queue to reconciling work in apply_controller

apply_controllers in the memberagent currently watches all the
Work resources created in the designated namespace on the Hub
cluster and reconciles them in order. This can result in serious
bottleneck in applying newer resources or newer updates to
existing resources over time especially when there are 1000s
of namespaces created a long time ago and never have any
updates to them, yet reconciled one by one before the
controller can get to the newer ones.

Fixes #

This PR integrates a priority queue for adding work objects to
the watch list to prioritize work objects that are newer, have
recent updates or have their conditions not Ready yet.

Note that the PR uses both `For` and `Watches` in the controller
setup. This is needed to avoid a runtime panic when using just `Watches`

This can also be done using just `For` without `Watches`, but that model 
would not use a `PriorityQueue` and just plays directly with the events. 
With that model, we will only watch the changed/new work objects and 
completely ignore old objects. With the `PriorityQueue`, we are actually 
adding every single Work object to the queue just in a prioritized order.

Signed-off-by: Sudheer Vinukonda <sudheerv@apache.org>

I have:

tested it in our production environment

Namespace Readiness (without patch)
```
test-sudheer-ns-10: NamespaceCreated: 2025-04-09T21:47:33Z, AllReady: 2025-04-09T22:13:15Z
test-sudheer-ns-11: NamespaceCreated: 2025-04-09T22:24:32Z, AllReady: 2025-04-09T23:03:57Z
test-sudheer-ns-12: NamespaceCreated: 2025-04-09T22:24:47Z, AllReady: 2025-04-09T23:03:57Z
test-sudheer-ns-13: NamespaceCreated: 2025-04-09T22:24:56Z, AllReady: 2025-04-09T23:03:57Z
```
Namespace Readiness (with patch)
```
test-sudheer-ns-37: NamespaceCreated: 2025-04-11T17:30:46Z, AllReady: 2025-04-11T17:31:16Z
test-sudheer-ns-38: NamespaceCreated: 2025-04-11T17:30:57Z, AllReady: 2025-04-11T17:31:27Z
test-sudheer-ns-39: NamespaceCreated: 2025-04-11T17:31:07Z, AllReady: 2025-04-11T17:31:37Z
test-sudheer-ns-40: NamespaceCreated: 2025-04-11T17:31:17Z, AllReady: 2025-04-11T17:31:48Z
```

### DIFF
--- a/pkg/controllers/clusterresourceplacement/resource_selector.go
+++ b/pkg/controllers/clusterresourceplacement/resource_selector.go
@@ -357,6 +357,8 @@ func (r *Reconciler) shouldSelectResource(gvr schema.GroupVersionResource) bool 
 
 // generateRawContent strips all the unnecessary fields to prepare the objects for dispatch.
 func generateRawContent(object *unstructured.Unstructured) ([]byte, error) {
+	// Make a deep copy of the object as we are modifying it.
+	object = object.DeepCopy()
 	// we keep the annotation/label/finalizer/owner references/delete grace period
 	object.SetResourceVersion("")
 	object.SetGeneration(0)


sudheerv:
@ryanzhang-oss  Can you take a look at the PR pls? Some checks are failing - is there a format for the title that I should follow?

pkg/controllers/work/apply_controller.go
```diff
@@ -101,6 +107,102 @@ const (
...
+func (h *PriorityQueueEventHandler) WorkPendingApply(ctx context.Context, obj client.Object) bool {
+	var work fleetv1beta1.Work
+	ns := obj.GetNamespace()
+	name := obj.GetName()
+	err := h.Client.Get(ctx, client.ObjectKey{
+		Namespace: ns,
+		Name:      name,
+	}, &work)
+	if err != nil {
+		// Log and return
```
ryanzhang-oss: why is "workPendingApply" == false when we can't get the work?

pkg/controllers/work/apply_controller.go
```diff
@@ -101,6 +107,102 @@ const (
...
+		klog.ErrorS(err, "Failed to get the work", "name", name, "ns", ns)
+		return false
+	}
+	availCond := meta.FindStatusCondition(work.Status.Conditions, fleetv1beta1.WorkConditionTypeAvailable)
+	appliedCond := meta.FindStatusCondition(work.Status.Conditions, fleetv1beta1.WorkConditionTypeApplied)
+
+	// check if the object has been recently modified
+	availCondLastUpdatedTime := availCond.LastTransitionTime.Time
+	appliedCondLastUpdatedTime := appliedCond.LastTransitionTime.Time
+	if time.Since(availCondLastUpdatedTime) < 1*time.Hour || time.Since(appliedCondLastUpdatedTime) < 1*time.Hour {
```
ryanzhang-oss:
```suggestion
	if time.Since(availCondLastUpdatedTime) < workAgeToReconcile  || time.Since(appliedCondLastUpdatedTime) < workAgeToReconcile  {
```

pkg/controllers/work/apply_controller.go
```diff
@@ -698,7 +812,8 @@ func (r *ApplyWorkReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		WithOptions(ctrloption.Options{
 			MaxConcurrentReconciles: r.concurrency,
 		}).
-		For(&fleetv1beta1.Work{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&fleetv1beta1.Work{}).
+		Watches(&fleetv1beta1.Work{}, eventHandler).
```
ryanzhang-oss: This handler does not seem to filter out status change? That will be a lot more noises.

ryanzhang-oss (commented):
Can you add a feature flag for that is turned off by default?

sudheerv:
> Can you add a feature flag (that is turned off by default) for this feature?

Added. Turns out there is duplication in the code on the HEAD branch (between work/apply_controller.go and workapplier/controller.go. I missed that in the original PR since v0.9.5 did not have this. I'm presuming there is a planned cleanup for work/apply_controller.go but in this PR, I added the code to both controllers. Plmk if this is not the correct understanding.

pkg/controllers/work/apply_controller.go
sudheerv: added filter

pkg/controllers/work/apply_controller.go
sudheerv: updated.

pkg/controllers/work/apply_controller.go
sudheerv: I was on two minds, since trying to add the object to queue may not really help in this case since the object GET failed but changed it to keep the semantics consistent.

pkg/controllers/work/apply_controller.go
```diff
@@ -42,6 +42,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apiextensionshelpers "k8s.io/apiextensions-apiserver/pkg/apihelpers"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
```
ryanzhang-oss: This controller is no longer in use, we will clean it up soon

pkg/controllers/workapplier/controller.go
```diff
@@ -71,17 +77,135 @@ const (
 	workFieldManagerName = "work-api-agent"
 )
 
+const (
+	workAgeToReconcile = 1 * time.Hour
```
ryanzhang-oss: can this also be a parameter with a default?

sudheerv: Ah, makes sense. Updated (only updated the workapplier/controller.go and not the work/apply_controller.go)

pkg/controllers/workapplier/controller.go
```diff
@@ -483,6 +614,26 @@ func (r *Reconciler) Leave(ctx context.Context) error {
...
+
+	// Create the event handler that uses the priority queue
+	eventHandler := &PriorityQueueEventHandler{
+		Queue:  queue, // Attach the priority queue to the event handler
+		Client: r.hubClient,
+	}
+
+	if r.watchWorkWithPriorityQueue {
+		workAgeToReconcile = time.Duration(r.watchWorkReconcileAgeMinutes) * time.Minute
+		return ctrl.NewControllerManagedBy(mgr).Named("work-controller").
```
ryanzhang-oss:
```suggestion
		return ctrl.NewControllerManagedBy(mgr).Named("work-applier-controller").
```

pkg/controllers/workapplier/controller.go
```diff
@@ -483,6 +614,26 @@ func (r *Reconciler) Leave(ctx context.Context) error {
...
+			WithOptions(ctrloption.Options{
+				MaxConcurrentReconciles: r.concurrentReconciles,
+			}).
+			For(&fleetv1beta1.Work{}).
+			Watches(&fleetv1beta1.Work{}, eventHandler).
```
ryanzhang-oss: are these not duplicate? Do you still want to watch for status change?

sudheerv: >> are these not duplicate? 

Without using `For(&fleetv1beta1.Work{}).`, `Watches(&fleetv1beta1.Work{}, eventHandler).` causes a runtime panic. 

Here's what chatgpt says

```
			For(&fleetv1beta1.Work{}).
			Watches(&fleetv1beta1.Work{}, eventHandler).

without using For, there is a runtime panic. Why are both needed

ChatGPT said:
In Kubernetes controller-runtime, using both:

go
Copy
Edit
For(&fleetv1beta1.Work{}).
Watches(&fleetv1beta1.Work{}, eventHandler).
might look redundant, but they serve different purposes, and omitting For(...) can lead to a runtime panic — exactly as you observed.

⚙️ Why For(...) Is Required
go
Copy
Edit
For(&fleetv1beta1.Work{})
This tells the controller what the primary resource is — the kind it Reconciles. It's required for the controller to set up the internal machinery that connects:

the GVK (Group-Version-Kind) of Work

to your Reconciler

and creates the initial informers and caches behind the scenes.

If you remove For(...), controller-runtime doesn’t know what kind of object your Reconciler is responsible for, so when events come in (even via Watches(...)), it doesn't know how to hydrate them correctly, leading to nil objects or panics during reconciliation.

🧭 What Watches(...) Does
go
Copy
Edit
Watches(&fleetv1beta1.Work{}, eventHandler)
This adds an additional watch on the same or different resource with a custom event handler, often to control:

When objects should be enqueued

Custom logic like filtering or prioritization

Watching secondary resources (e.g., Secrets, Pods related to a Work)

In your case, you're watching the same resource (Work) twice:

Once via For(...) (default watch that enqueues on every change)

Again via Watches(...) with a custom eventHandler (maybe for fine-grained control, e.g., filtering status-only changes or using a priority queue)

This is a valid and common pattern when you need to override default behavior.
```



re: status change, I'm ignoring the status changes via the Update function below.

```
func (h *PriorityQueueEventHandler) Update(ctx context.Context, evt event.TypedUpdateEvent[client.Object], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
	// Ignore updates where only the status changed
	oldObj := evt.ObjectOld.DeepCopyObject()
	newObj := evt.ObjectNew.DeepCopyObject()

	// Zero out the status
	if oldWork, ok := oldObj.(*fleetv1beta1.Work); ok {
		oldWork.Status = fleetv1beta1.WorkStatus{}
	}
	if newWork, ok := newObj.(*fleetv1beta1.Work); ok {
		newWork.Status = fleetv1beta1.WorkStatus{}
	}

	if !equality.Semantic.DeepEqual(oldObj, newObj) {
		// ignore status changes to prevent noise
		h.AddToPriorityQueue(ctx, evt.ObjectNew, true)
	}
	klog.V(4).InfoS("ignoring update event with only status change", "work", evt.ObjectNew.GetName())
}

```

PR #51, opened 2025-05-06
feat: add priority queue to reconciling work in apply_controller

sudheerv:
add priority queue to reconciling work in apply_controller

apply_controllers in the memberagent currently watches all the Work resources created in the designated namespace on the Hub cluster and reconciles them in order. This can result in serious bottleneck in applying newer resources or newer updates to existing resources over time especially when there are 1000s of namespaces created a long time ago and never have any updates to them, yet reconciled one by one before the controller can get to the newer ones.

Fixes #

This PR integrates a priority queue for adding work objects to the watch list to prioritize work objects that are newer, have recent updates or have their conditions not Ready yet.

Note that the PR uses both For and Watches in the controller setup. This is needed to avoid a runtime panic when using just Watches

This can also be done using just For without Watches, but that model would not use a PriorityQueue and just plays directly with the events. With that model, we will only watch the changed/new work objects and completely ignore old objects. With the PriorityQueue, we are actually adding every single Work object to the queue just in a prioritized order.

Signed-off-by: Sudheer Vinukonda sudheerv@apache.org

I have:

tested it in our production environment

Namespace Readiness (without patch)

```
test-sudheer-ns-10: NamespaceCreated: 2025-04-09T21:47:33Z, AllReady: 2025-04-09T22:13:15Z
test-sudheer-ns-11: NamespaceCreated: 2025-04-09T22:24:32Z, AllReady: 2025-04-09T23:03:57Z
test-sudheer-ns-12: NamespaceCreated: 2025-04-09T22:24:47Z, AllReady: 2025-04-09T23:03:57Z
test-sudheer-ns-13: NamespaceCreated: 2025-04-09T22:24:56Z, AllReady: 2025-04-09T23:03:57Z
```
Namespace Readiness (with patch)
```
test-sudheer-ns-37: NamespaceCreated: 2025-04-11T17:30:46Z, AllReady: 2025-04-11T17:31:16Z
test-sudheer-ns-38: NamespaceCreated: 2025-04-11T17:30:57Z, AllReady: 2025-04-11T17:31:27Z
test-sudheer-ns-39: NamespaceCreated: 2025-04-11T17:31:07Z, AllReady: 2025-04-11T17:31:37Z
test-sudheer-ns-40: NamespaceCreated: 2025-04-11T17:31:17Z, AllReady: 2025-04-11T17:31:48Z
```


pkg/controllers/workapplier/controller.go
```diff
@@ -71,17 +77,135 @@ const (
...
+		Namespace: ns,
+		Name:      name,
+	}, &work)
+	if err != nil {
+		// Log and return
+		klog.ErrorS(err, "Failed to get the work", "name", name, "ns", ns)
+		return true
+	}
+	availCond := meta.FindStatusCondition(work.Status.Conditions, fleetv1beta1.WorkConditionTypeAvailable)
+	appliedCond := meta.FindStatusCondition(work.Status.Conditions, fleetv1beta1.WorkConditionTypeApplied)
```
ryanzhang-oss, line 115: these can be nil

pkg/controllers/workapplier/controller.go
```diff
@@ -71,17 +77,136 @@ const (
...
+
+	// check if the object has been recently modified
+	availCondLastUpdatedTime := availCond.LastTransitionTime.Time
+	appliedCondLastUpdatedTime := appliedCond.LastTransitionTime.Time
+	if time.Since(availCondLastUpdatedTime) < workAgeToReconcile || time.Since(appliedCondLastUpdatedTime) < workAgeToReconcile {
+		return true
+	}
+
```
ryanzhang-oss, line 123: I think what we actually need is the LastModifiedTime, we will add that annotation.

ryanzhang-oss: do we still want to say that the work is "pending apply" even if it's fully applied?

pkg/controllers/workapplier/controller.go
```diff
@@ -71,17 +77,136 @@ const (
...
+func (h *PriorityQueueEventHandler) AddToPriorityQueue(ctx context.Context, obj client.Object, alwaysAdd bool) {
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Namespace: obj.GetNamespace(),
+			Name:      obj.GetName(),
+		},
+	}
+
+	objAge := time.Since(obj.GetCreationTimestamp().Time)
```
ryanzhang-oss, line 141: again, I think we need LastModifiedTime.

pkg/controllers/workapplier/controller.go
```diff
@@ -71,17 +77,136 @@ const (
...
+
+	var objPriority int
+	if alwaysAdd || objAge < workAgeToReconcile || h.WorkPendingApply(ctx, obj) {
```
ryanzhang-oss, line 144: "workAgeToReconcile" is already taken into consideration in the WorkPendingApply function, does it make sense to use it here again?

ryanzhang-oss (approved):

pkg/controllers/workapplier/controller.go
sudheerv: Ah yes, I missed this. Will fix it.

PR #52, opened 2025-05-06
feat: add new work applier metrics

michaelawyu:

This PR adds new work applier metrics.

An old metric, migrated from the old work applier, has been removed, as the annotation required to emit data points has not been set since we moved away from v1alpha1 API (i.e., the annotation is only set in the v1alpha1 CRP controller).

I have:

- [x] Unit tests

Please see the comments in the PR code change for further explanation on the choice of the metrics.


michaelawyu:
Note: as mentioned in the PR description, the original work applier metric has been removed as it hasn't been set for a long while.

At this moment it might no longer be possible for us to capture the whole workflow latency (i.e., from the moment CRP controller detects a resource change to the moment where the manifests are eventually applied), due to a variety of factors (e.g., rolling out). It's still possible to add a metric that tracks the latency between the moment where work generator prepares a work object to the moment where manifests are finally applied, though how useful such a metric can be remains a topic to further discuss.

Note that the controller runtime default metrics for the work applier are always available, so we do know the latency on work applier reconciliations.

pkg/controllers/workapplier/controller_test.go
```diff
@@ -188,6 +189,12 @@ func TestMain(m *testing.M) {
 	// Initialize the variables.
 	initializeVariables()
 
+	// Register the metrics.
+	ctrlmetrics.Registry.MustRegister(
+		metrics.FleetWorkProcessingRequestsTotal,
```
britaniar, line 194: Why are we registering it twice? Once in the beginning of the member agent and here.

pkg/controllers/workapplier/metrics_test.go
```diff
@@ -0,0 +1,539 @@
...
+            	fleet_manifest_processing_requests_total{apply_status="manifestapplyfailed",availability_status="skipped",diff_detection_status="not_found",diff_reporting_status="skipped",drift_detection_status="not_found"} 1
+            	fleet_manifest_processing_requests_total{apply_status="skipped",availability_status="skipped",diff_detection_status="found",diff_reporting_status="founddiff",drift_detection_status="not_found"} 1
+            	fleet_manifest_processing_requests_total{apply_status="skipped",availability_status="skipped",diff_detection_status="not_found",diff_reporting_status="failed",drift_detection_status="not_found"} 1
+            	fleet_manifest_processing_requests_total{apply_status="skipped",availability_status="skipped",diff_detection_status="not_found",diff_reporting_status="nodifffound",drift_detection_status="not_found"} 2
+            	fleet_manifest_processing_requests_total{apply_status="unknown",availability_status="unknown",diff_detection_status="not_found",diff_reporting_status="unknown",drift_detection_status="not_found"} 1
+			`,
+		},
+	}
+
+	for _, tc := range testCases {
```
britaniar, line 519: Hi Chen, would it be possible to test the metrics through integration test like updaterun does? We kind established how we wanted to test them and updaterun seems most similar to your case.
https://github.com/jwtty/kubefleet/blob/00ca14e5c8ed158488e3cdf16d3ccfef74b3f08c/pkg/controllers/updaterun/controller_integration_test.go#L256

michaelawyu: Hi Britania! Yeah, let me add both. There happens to be a few cases that are difficult to emulate with integration tests.

michaelawyu: Hi Britania! I fear that this pattern might not necessarily apply to the work applier metrics -> the problem is that the metrics we add at this moment for the work applier is a counter that counts individual processing requests (and their results), and in the integration tests we do not have a way to reliably tell how many requests have been processed as the controller might have requeues, failures, etc.

michaelawyu: For the staged updaterun ITs, the metric added (if I understand it correctly) is a gauge for timestamps, and the test comparing logic is essentially checking if a timestamp has been set:

```go
cmp.Comparer(func(a, b *prometheusclientmodel.Gauge) bool {
			return (a.GetValue() > 0) == (b.GetValue() > 0)
		}),
```

This makes sense for this specific metric I guess, and the labels (names, etc.) for the metric are generally guaranteed to be consistent in a IT spec. But for the work applier and its metrics situation are a bit different; the labels (applied or not, available or not, etc.) and their respective counters might be in state of a flux during a IT spec due to controllers retrying/requeueing/failing loops. With that being said, we could still add metric ITs, but for work applier the test would be merely checking if some data has been emitted (or we need to write some complex comparisons to ignore data points on an individual basis), which might not be perfect (or a bit overkilling) just in my personal opinion.

michaelawyu: But I totally understand your concern; if you would like to discuss this further, please let me know 🙏

pkg/controllers/workapplier/metrics.go
```diff
@@ -17,29 +17,194 @@ limitations under the License.
...
+	workAppliedCond := meta.FindStatusCondition(work.Status.Conditions, fleetv1beta1.WorkConditionTypeApplied)
+	switch {
+	case workAppliedCond == nil:
+		workApplyStatus = workApplyStatusSkipped
+	case workAppliedCond.ObservedGeneration != work.Generation:
+		// Normally this should never occur; as the method is called right after the status
+		// is refreshed.
+		workApplyStatus = workApplyStatusUnknown
+	case workAppliedCond.Status == metav1.ConditionTrue:
+		workApplyStatus = workApplyStatusApplied
```
ryanzhang-oss: can we use the reason of the condition?

pkg/controllers/workapplier/metrics.go
```diff
@@ -17,29 +17,194 @@ limitations under the License.
...
 
-	klog.V(2).InfoS("No last update timestamp found on the Work object", "work", klog.KObj(work))
+	metrics.FleetWorkProcessingRequestsTotal.WithLabelValues(
+		workApplyStatus,
+		workAvailabilityStatus,
+		workDiffReportedStatus,
+	).Inc()
+
+	// Increment the manifest processing request counter.
+	for idx := range work.Status.ManifestConditions {
```
ryanzhang-oss, line 131: does it make sense to add manifest condition if the work generation is not the latest? Is it possible for it to be not latest? If not, add an unexpected error?

cmd/memberagent/main.go
```diff
@@ -103,7 +103,12 @@ func init() {
 	utilruntime.Must(placementv1beta1.AddToScheme(scheme))
 	//+kubebuilder:scaffold:scheme
 
-	metrics.Registry.MustRegister(fleetmetrics.JoinResultMetrics, fleetmetrics.LeaveResultMetrics, fleetmetrics.WorkApplyTime)
+	metrics.Registry.MustRegister(
```
ryanzhang-oss, line 106: I think the latency of apply is still something we want to track (not in the PR).

pkg/controllers/workapplier/metrics.go
michaelawyu: Sure, I will make the edits.

pkg/controllers/workapplier/metrics.go
michaelawyu: Hi Ryan! This method is called only after the status refreshing methods so it's kind of guaranteed to have fresh status conditions; but I could add a sanity check just in case.

cmd/memberagent/main.go
michaelawyu: Yeah, that is true. It's a bit tricky to implement the whole-flow latency though, but doing a partial bookkeeping (e.g., from work generator to the work applier) would still be relatively easy I think

pkg/controllers/workapplier/metrics.go
```diff
@@ -36,105 +39,87 @@ import (
...
+		// At this point Fleet does not set the Applied condition to the Unknown status; this
+		// branch is added just for completeness reasons.
+		workApplyStatus = workOrManifestStatusUnknown
+	case len(workAppliedCond.Reason) == 0:
 		// Normally this should never occur; this branch is added just for completeness reasons.
-		workApplyStatus = workApplyStatusUnknown
+		_ = controller.NewUnexpectedBehaviorError(fmt.Errorf("failed to track work metrics: applied condition has no reason (work=%v)", klog.KObj(work)))
+		shouldSkip = true
+	default:
+		workApplyStatus = strings.ToLower(workAppliedCond.Reason)
```
britaniar: why do we make it all lowercase? is it not harder to read this way?

pkg/controllers/workapplier/metrics.go
```diff
@@ -36,105 +39,87 @@ import (
...
-
-	// The values for the manifest drift detection status label.
-	manifestDriftDetectionStatusFound    = "found"
-	manifestDriftDetectionStatusNotFound = "not_found"
-
-	// The values for the manifest diff detection status label.
-	manifestDiffDetectionStatusFound    = "found"
-	manifestDiffDetectionStatusNotFound = "not_found"
+	manifestDriftOrDiffDetectionStatusFound    = "found"
+	manifestDriftOrDiffDetectionStatusNotFound = "not_found"
```
britaniar: I think we should try to be consistent with the reason. Our reasons don't use "_"
```suggestion
	manifestDriftOrDiffDetectionStatusNotFound = "notfound"
```

pkg/controllers/workapplier/metrics.go
michaelawyu: Oh, that's because Prometheus actually recommends using snake case (`_`, all lower letters) as mentioned in the docs; I do not particularly have a strong preference on this though. Let me make the edits.

pkg/controllers/workapplier/metrics.go
michaelawyu: Same as the issue before; it's added here because Prometheus prefers snake case. But also as mentioned before I do not have a strong preference on this part. Let me make it camelCase then.

britaniar (approved):

PR #54, opened 2025-05-06
feat: enable profile in agents

britaniar:

Fixes #

I have: enable pprof in hub and member agent

image

image


cmd/hubagent/main.go
```diff
@@ -136,6 +139,17 @@ func main() {
...
 	}
 
+	if opts.EnablePprof {
+		klog.InfoS("Starting profiling", "port", opts.ProfilePort)
+		go func() {
+			server := &http.Server{
+				Addr:              fmt.Sprintf(":%d", opts.ProfilePort),
+				ReadHeaderTimeout: 5 * time.Second,
+			}
+			klog.ErrorS(server.ListenAndServe(), "unable to start profiling server")
```
Arvindthiru: Can we get the error, check it's not nil and then log the error ?

charts/member-agent/values.yaml
```diff
@@ -61,3 +61,5 @@ useCAAuth: false
 
 enableV1Alpha1APIs: true
 enableV1Beta1APIs: false
+
+enablePprof: false
```
Arvindthiru: nit: new line at the end of the file

charts/hub-agent/templates/deployment.yaml
```diff
@@ -33,6 +33,7 @@ spec:
             - --enable-cluster-inventory-apis={{ .Values.enableClusterInventoryAPI }}
             - --enable-staged-update-run-apis={{ .Values.enableStagedUpdateRunAPIs }}
             - --enable-eviction-apis={{ .Values.enableEvictionAPIs}}
+            - --enable-pprof={{ .Values.enablePprof }}
```
Arvindthiru, line 36: Do we need to add the containerPort for hub-agent's deployment file as well ?

cmd/hubagent/main.go
```diff
@@ -136,6 +139,17 @@ func main() {
 		exitWithErrorFunc()
 	}
 
+	if opts.EnablePprof {
+		klog.InfoS("Starting profiling", "port", opts.ProfilePort)
+		go func() {
+			server := &http.Server{
```
jwtty: We need to shutdown the server before process terminates.

cmd/hubagent/main.go
```diff
@@ -18,10 +18,13 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"math"
+	"net/http"
 	"os"
 	"strings"
 	"sync"
+	"time"
```
jwtty: Reading the pprof doc https://pkg.go.dev/net/http/pprof, they are saying:
```
To use pprof, link this package into your program:
import _ "net/http/pprof"
```

charts/hub-agent/values.yaml
```diff
@@ -38,6 +38,7 @@ enableV1Beta1APIs: true
 enableClusterInventoryAPI: true
 enableStagedUpdateRunAPIs: true
 enableEvictionAPIs: true
+enablePprof: false
```
ryanzhang-oss: can be default to true

charts/member-agent/templates/deployment.yaml
```diff
@@ -75,6 +76,9 @@ spec:
             - containerPort: 8091
               name: memberhealthz
               protocol: TCP
+            - containerPort: 6065
+              name: pprof
+              protocol: TCP
```
ryanzhang-oss: why is this not in the hub agent?

cmd/hubagent/main.go
britaniar: image

michaelawyu:
A note: this PR is currently failing as the member agents are stuck in a crash loop -> the PR itself is fine actually; the failure comes from the fact that at this moment our compatibility tests are configured to prefer the new (latest) charts, and installing the old image with the new charts defined in this PR would trigger an error (the old image does not recognize the new flag).

michaelawyu:
> A note: ...

I will send a PR to fine-tune the behavior of the compatibility tests; I think it should be fine to disregard the failures for now, or we could wait a bit -> after the fix PR is merged things should turn green

charts/member-agent/templates/deployment.yaml
```diff
@@ -75,6 +77,9 @@ spec:
             - containerPort: 8091
               name: memberhealthz
               protocol: TCP
+            - containerPort: 6065
```
ryanzhang-oss: why is this hard coded while the parameter port is from the value?

charts/member-agent/templates/deployment.yaml
```diff
@@ -34,6 +34,8 @@ spec:
             - -add_dir_header
             - --enable-v1alpha1-apis={{ .Values.enableV1Alpha1APIs }}
             - --enable-v1beta1-apis={{ .Values.enableV1Beta1APIs }}
+            - --enable-pprof={{ .Values.enablePprof | default true }}
+            - --pprof-port={{ .Values.pprofPort | default 6065 }}
```
ryanzhang-oss: the code default is 6065 too, is another default necessary?

cmd/hubagent/options/options.go
```diff
@@ -158,6 +164,8 @@ func (o *Options) AddFlags(flags *flag.FlagSet) {
 	flags.DurationVar(&o.ForceDeleteWaitTime.Duration, "force-delete-wait-time", 15*time.Minute, "The duration the hub agent waits before force deleting a member cluster.")
 	flags.BoolVar(&o.EnableStagedUpdateRunAPIs, "enable-staged-update-run-apis", true, "If set, the agents will watch for the ClusterStagedUpdateRun APIs.")
 	flags.BoolVar(&o.EnableEvictionAPIs, "enable-eviction-apis", true, "If set, the agents will watch for the Eviction and PlacementDisruptionBudget APIs.")
+	flags.BoolVar(&o.EnablePprof, "enable-pprof", true, "If set, the pprof profiling is enabled.")
```
ryanzhang-oss: how much overhead does this introduce?

cmd/memberagent/main.go
```diff
@@ -124,6 +126,20 @@ func main() {
 		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 	}
 
+	if *enablePprof {
```
jwtty: A quick question: Can we use the controller-runtime built-in pprof directly: https://book.kubebuilder.io/reference/pprof-tutorial

britaniar: It seems to work! Thanks for finding that. It's similar implementation as metrics.

cmd/hubagent/options/options.go
britaniar: It does introduce overhead with large workloads. I think we should have it as false in our production like they recommend.
https://book.kubebuilder.io/reference/pprof-tutorial
image

ryanzhang-oss: then we should make the default to false

charts/hub-agent/values.yaml
ryanzhang-oss: now that the code default it false

charts/member-agent/values.yaml
```diff
@@ -61,3 +61,7 @@ useCAAuth: false
...
```
ryanzhang-oss: we can turn it on for upstream helm chart

ryanzhang-oss (approved):

jwtty (approved):
lgtm

PR #58, opened 2025-05-08
fix: address a SSA related issue which might cause all apply attempts (except the first one) to fail due to conflicts 

michaelawyu:

This PR fixes an issue in Fleet where apply attempts might fail to due field manager conflicts when SSA (server-side apply) is used; the user will see their first apply op successfully, but any field change later cannot be applied.

Note that this bug exists in both the new and the old work applier; only the new work applier receives the fix.

I have:

- [x] Unit tests
- [x] E2E tests 

A few additional E2E test cases will be sent as separate PRs.


michaelawyu:
Wrong branch; will re-send the PR.

PR #62, opened 2025-05-13
feat: add CEL validation to prevent waitTime on Approval StageTask

Arvindthiru:

Fixes #

I have:


config/crd/bases/placement.kubernetes-fleet.io_clusterstagedupdateruns.yaml
```diff
@@ -1814,6 +1814,13 @@ spec:
                             type: object
                           maxItems: 2
                           type: array
+                          x-kubernetes-validations:
+                          - message: AfterStageTaskType is Approval, waitTime is not
```
jwtty, line 1818: Just curious: why do we have this line break between "not" and "allowed" here but not the strategies.

pkg/controllers/updaterun/controller.go
```diff
@@ -79,6 +79,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	}
 	runObjRef := klog.KObj(&updateRun)
 
+	// Remove waitTime from the updateRun status for AfterStageTask for type Approval.
```
ryanzhang-oss, line 82: IIRC, this is just to fix the existing approval CRs?

ryanzhang-oss: why put it here instead of in the init?

pkg/controllers/updaterun/initialization.go
```diff
@@ -518,3 +524,14 @@ func (r *Reconciler) recordInitializationFailed(ctx context.Context, updateRun *
...
+
+func (r *Reconciler) removeWaitTimeFromAfterStageTasks(ctx context.Context, updateStrategy *placementv1beta1.ClusterStagedUpdateStrategy) error {
+	for i := range updateStrategy.Spec.Stages {
+		for j := range updateStrategy.Spec.Stages[i].AfterStageTasks {
+			if updateStrategy.Spec.Stages[i].AfterStageTasks[j].Type == placementv1beta1.AfterStageTaskTypeApproval {
+				updateStrategy.Spec.Stages[i].AfterStageTasks[j].WaitTime = nil
+			}
+		}
+	}
+	return r.Update(ctx, updateStrategy)
```
jwtty: Let's use `r.Client.Update` to be consistent, also please wrap the error with `controller.NewUpdateIgnoreConflictError()`

pkg/controllers/updaterun/initialization.go
```diff
@@ -250,6 +250,12 @@ func (r *Reconciler) generateStagesByStrategy(
 		// other err can be retried.
 		return controller.NewAPIServerError(true, err)
 	}
+	// Remove waitTime from the after stage tasks of type Approval.
+	if err := r.removeWaitTimeFromAfterStageTasks(ctx, &updateStrategy); err != nil {
+		klog.ErrorS(err, "Failed to remove wait time from after stage tasks", "clusterStagedUpdateStrategy", updateStrategy.Name, "clusterStagedUpdateRun", updateRunRef)
```
jwtty:
```suggestion
		klog.ErrorS(err, "Failed to remove wait time from after stage tasks", "clusterStagedUpdateStrategy", klog.KObj(updateStrategy), "clusterStagedUpdateRun", updateRunRef)
```

pkg/controllers/updaterun/initialization.go
```diff
@@ -250,6 +250,12 @@ func (r *Reconciler) generateStagesByStrategy(
...
+		return fmt.Errorf("%w: %s", errInitializedFailed, err.Error())
```
jwtty: This error is from apiserver and can be retried, so it should not be wrapped with `errInitializedFailed`. See an example here: https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/controllers/updaterun/initialization.go#L315C4-L316C50

pkg/controllers/updaterun/controller.go
```diff
@@ -341,3 +347,18 @@ func emitUpdateRunStatusMetric(updateRun *placementv1beta1.ClusterStagedUpdateRu
...
+	if updateRun.Status.StagedUpdateStrategySnapshot != nil {
+		for i := range updateRun.Status.StagedUpdateStrategySnapshot.Stages {
+			for j := range updateRun.Status.StagedUpdateStrategySnapshot.Stages[i].AfterStageTasks {
+				if updateRun.Status.StagedUpdateStrategySnapshot.Stages[i].AfterStageTasks[j].Type == placementv1beta1.AfterStageTaskTypeApproval {
+					updateRun.Status.StagedUpdateStrategySnapshot.Stages[i].AfterStageTasks[j].WaitTime = nil
+				}
+			}
+		}
+		return r.Client.Status().Update(ctx, updateRun)
```
jwtty: ditto, wrap with `controller.NewAPIServerError()`

pkg/controllers/updaterun/controller.go
jwtty: Fix existing updateRun CRs. We have to do it here because when an updateRun is executing or being deleted, the reconciler won't enter initialization any more. In deletion, we update to remove the finalizer; during execution/validation, we update the updateRun status. All these updates will fail if the `waitTime` is not set to nil.

ryanzhang-oss: discussed with Arvind, looks like it's enough to just correct the in-memory status and piggyback that when the status is updated at the end of the reconcile.

pkg/controllers/updaterun/initialization.go
```diff
@@ -250,6 +250,12 @@ func (r *Reconciler) generateStagesByStrategy(
 		// other err can be retried.
 		return controller.NewAPIServerError(true, err)
 	}
+	// Remove waitTime from the after stage tasks of type Approval.
+	if err := r.removeWaitTimeFromAfterStageTasks(ctx, &updateStrategy); err != nil {
```
ryanzhang-oss: maybe consider using a mutation webhook to achieve this as this does not guarantee to correct the all the strategy (orphaned one).

ryanzhang-oss: or we can keep this in until we know there are no more malformated strategy

config/crd/bases/placement.kubernetes-fleet.io_clusterstagedupdateruns.yaml
Arvindthiru: auto-generated code so I'm not sure

pkg/controllers/updaterun/initialization.go
Arvindthiru: Even with a mutation webhook we won't be able to fix orphaned CSUS?, since it's never being updated and it's only used as an information source

pkg/controllers/updaterun/initialization.go
```diff
@@ -404,7 +407,7 @@ func validateAfterStageTask(tasks []placementv1beta1.AfterStageTask) error {
 	}
 	for i, task := range tasks {
 		if task.Type == placementv1beta1.AfterStageTaskTypeTimedWait {
-			if task.WaitTime.Duration <= 0 {
+			if task.WaitTime != nil && task.WaitTime.Duration <= 0 {
```
jwtty: Should we use "if task.WaitTime == nil || task.WaitTime.Duration <= 0"? We require users to provide a value in this case. There could be chance that the code is released but the CRD with CEL check has not been updated to latest.

pkg/controllers/updaterun/initialization.go
```diff
@@ -404,8 +407,11 @@ func validateAfterStageTask(tasks []placementv1beta1.AfterStageTask) error {
 	}
 	for i, task := range tasks {
 		if task.Type == placementv1beta1.AfterStageTaskTypeTimedWait {
-			if task.WaitTime.Duration <= 0 {
-				return fmt.Errorf("task %d has wait duration <= 0", i)
+			if task.WaitTime == nil {
+				return fmt.Errorf("task %d of type TimedWait has wait duration set to nil", i)
+			}
+			if task.WaitTime != nil && task.WaitTime.Duration <= 0 {
```
jwtty: nit: no need task.WaitTime != nil here anymore

pkg/controllers/updaterun/initialization.go
```diff
@@ -404,8 +407,11 @@ func validateAfterStageTask(tasks []placementv1beta1.AfterStageTask) error {
 	}
 	for i, task := range tasks {
 		if task.Type == placementv1beta1.AfterStageTaskTypeTimedWait {
+			if task.WaitTime == nil {
+				return fmt.Errorf("task %d of type TimedWait has wait duration set to nil", i)
+			}
 			if task.WaitTime.Duration <= 0 {
-				return fmt.Errorf("task %d has wait duration <= 0", i)
+				return fmt.Errorf("task %d of type TimedWait  has wait duration <= 0", i)
```
jwtty:
```suggestion
				return fmt.Errorf("task %d of type TimedWait has wait duration <= 0", i)
```

jwtty (dismissed):

PR #63, opened 2025-05-13
fix: fix updateRun error msg when a binding does not have expected state

jwtty:
Currently when there're 2 concurrent updateRuns for the same CRP and the same one preempts the first, the first would fail with a very long and hard-to-understand error msg:
```
- lastTransitionTime: "2025-05-13T03:45:41Z"
    message: 'cannot continue the ClusterStagedUpdateRun: unexpected behavior which
      cannot be handled by the controller: the updating cluster `member1` in the stage
      staging does not match the cluster status: &{ClusterName:member1 ResourceOverrideSnapshots:[]
      ClusterResourceOverrideSnapshots:[] Conditions:[{Type:Started Status:True ObservedGeneration:1
      LastTransitionTime:2025-05-13 03:45:26 +0000 UTC Reason:ClusterUpdatingStarted
      Message:Cluster update started}]}, binding: {State:Bound ResourceSnapshotName:example-placement-1-snapshot
      ResourceOverrideSnapshots:[] ClusterResourceOverrideSnapshots:[] SchedulingPolicySnapshotName:example-placement-0
      TargetCluster:member1 ClusterDecision:{ClusterName:member1 Selected:true ClusterScore:0xc002737d70
      Reason:Successfully scheduled resources for placement in "member1" (affinity
      score: 0, topology spread score: 0): picked by scheduling policy} ApplyStrategy:<nil>},
      condition: &Condition{Type:RolloutStarted,Status:True,ObservedGeneration:3,LastTransitionTime:2025-05-13
      03:45:38 +0000 UTC,Reason:RolloutStarted,Message:Detected the new changes on
      the resources and started the rollout process, resourceSnapshotIndex: 1, clusterStagedUpdateRun:
      example-run-1,}'
    observedGeneration: 1
    reason: UpdateRunFailed
    status: "False"
    type: Succeeded
```

This PR makes the error msg easier to understand:
```
- lastTransitionTime: "2025-05-15T22:03:26Z"
        message: 'failed to process the request due to a client error: the clusterResourceBinding
          of the updating cluster `member1` in the stage `staging` is not up-to-date
          with the desired status, please check the status of binding `example-placement-member1-043b2fcc`
          and see if there is a concurrent updateRun referencing the same clusterResourcePlacement
          and updating the same binding'
        observedGeneration: 1
        reason: ClusterUpdatingFailed
        status: "False"
        type: Succeeded
```

Fixes #

I have:

### DIFF
--- a/pkg/controllers/updaterun/execution.go
+++ b/pkg/controllers/updaterun/execution.go
@@ -176,7 +176,8 @@ func (r *Reconciler) executeUpdatingStage(
 			// This issue mostly happens when there are concurrent updateRuns referencing the same clusterResourcePlacement but releasing different versions.
 			// After the 1st updateRun updates the binding, and before the controller re-checks the binding status, the 2nd updateRun updates the same binding, and thus the 1st updateRun is preempted and observes the binding not matching the desired state.
 			preemptedErr := controller.NewUserError(fmt.Errorf("the clusterResourceBinding of the updating cluster `%s` in the stage `%s` is not up-to-date with the desired status, "+
-				"please check the status of binding `%s` and see if there is a concurrent updateRun referencing the same clusterResourcePlacement and updating the same cluster", clusterStatus.ClusterName, updatingStageStatus.StageName, klog.KObj(binding)))
+				"please check the status of binding `%s` and see if there is a concurrent updateRun referencing the same clusterResourcePlacement and updating the same cluster",
+				clusterStatus.ClusterName, updatingStageStatus.StageName, klog.KObj(binding)))
 			klog.ErrorS(preemptedErr, "The binding has been changed during updating",
 				"binding spec insync", inSync, "binding state", binding.Spec.State,
 				"binding RolloutStarted", rolloutStarted, "binding", klog.KObj(binding), "clusterStagedUpdateRun", updateRunRef)
--- a/pkg/controllers/updaterun/execution.go
+++ b/pkg/controllers/updaterun/execution.go
@@ -170,12 +170,16 @@ func (r *Reconciler) executeUpdatingStage(
 		}
 
 		// Now the cluster has to be updating, the binding should point to the right resource snapshot and the binding should be bound.
-		if insync, rolloutStarted := isBindingSyncedWithClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus), condition.IsConditionStatusTrue(meta.FindStatusCondition(binding.Status.Conditions, string(placementv1beta1.ResourceBindingRolloutStarted)), binding.Generation); !insync || binding.Spec.State != placementv1beta1.BindingStateBound || !rolloutStarted {
+		insync := isBindingSyncedWithClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus)
+		rolloutStarted := condition.IsConditionStatusTrue(meta.FindStatusCondition(binding.Status.Conditions, string(placementv1beta1.ResourceBindingRolloutStarted)), binding.Generation)
+		if !insync || !rolloutStarted || binding.Spec.State != placementv1beta1.BindingStateBound {
 			// This issue mostly happens when there are concurrent updateRuns referencing the same clusterResourcePlacement but releasing different versions.
 			// After the 1st updateRun updates the binding, and before the controller re-checks the binding status, the 2nd updateRun updates the same binding, and thus the 1st updateRun is preempted and observes the binding not matching the desired state.
 			preemptedErr := controller.NewUserError(fmt.Errorf("the clusterResourceBinding of the updating cluster `%s` in the stage `%s` is not up-to-date with the desired status, "+
-				"please check the binding status and see if there is a concurrent updateRun referencing the same clusterResourcePlacement and updating the same binding", clusterStatus.ClusterName, updatingStageStatus.StageName))
-			klog.ErrorS(preemptedErr, "The binding has been changed during updating", "binding spec insync", insync, "binding state", binding.Spec.State, "binding RolloutStarted", rolloutStarted, "binding", klog.KObj(binding), "clusterStagedUpdateRun", updateRunRef)
+				"please check the status of binding `%s` and see if there is a concurrent updateRun referencing the same clusterResourcePlacement and updating the same binding", clusterStatus.ClusterName, updatingStageStatus.StageName, klog.KObj(binding)))
+			klog.ErrorS(preemptedErr, "The binding has been changed during updating",
+				"binding spec insync", insync, "binding state", binding.Spec.State,
+				"binding RolloutStarted", rolloutStarted, "binding", klog.KObj(binding), "clusterStagedUpdateRun", updateRunRef)
 			markClusterUpdatingFailed(clusterStatus, updateRun.Generation, preemptedErr.Error())
 			return 0, fmt.Errorf("%w: %s", errStagedUpdatedAborted, preemptedErr.Error())
 		}
--- a/pkg/controllers/updaterun/execution.go
+++ b/pkg/controllers/updaterun/execution.go
@@ -122,7 +122,7 @@ func (r *Reconciler) executeUpdatingStage(
 		binding := toBeUpdatedBindingsMap[clusterStatus.ClusterName]
 		if !condition.IsConditionStatusTrue(clusterStartedCond, updateRun.Generation) {
 			// The cluster has not started updating yet.
-			if diff := getBindingDiffFromClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus); diff != "" {
+			if !isBindingSyncedWithClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus) {
 				klog.V(2).InfoS("Found the first cluster that needs to be updated", "cluster", clusterStatus.ClusterName, "stage", updatingStageStatus.StageName, "clusterStagedUpdateRun", updateRunRef)
 				// The binding is not up-to-date with the cluster status.
 				binding.Spec.State = placementv1beta1.BindingStateBound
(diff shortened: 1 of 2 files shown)


pkg/controllers/updaterun/execution.go
```diff
@@ -170,11 +170,13 @@ func (r *Reconciler) executeUpdatingStage(
...
+		if diff := getBindingDiffFromClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus); diff != "" || binding.Spec.State != placementv1beta1.BindingStateBound ||
 			!condition.IsConditionStatusTrue(meta.FindStatusCondition(binding.Status.Conditions, string(placementv1beta1.ResourceBindingRolloutStarted)), binding.Generation) {
-			unexpectedErr := controller.NewUnexpectedBehaviorError(fmt.Errorf("the updating cluster `%s` in the stage %s does not match the cluster status: %+v, binding: %+v, condition: %+v",
-				clusterStatus.ClusterName, updatingStageStatus.StageName, clusterStatus, binding.Spec, binding.GetCondition(string(placementv1beta1.ResourceBindingRolloutStarted))))
-			klog.ErrorS(unexpectedErr, "The binding has been changed during updating, please check if there's concurrent clusterStagedUpdateRun", "clusterStagedUpdateRun", updateRunRef)
+			unexpectedErr := controller.NewUnexpectedBehaviorError(fmt.Errorf("the clusterResourceBinding of the updating cluster `%s` in the stage `%s` does not have expected status: "+
```
zhiying-lin: we cannot log this as unexpected, otherwise, it will trigger the alert. Should it be user error?

pkg/controllers/updaterun/execution.go
```diff
@@ -170,11 +170,13 @@ func (r *Reconciler) executeUpdatingStage(
...
+				"binding spec diff: %s; binding state (want Bound): %s; binding RolloutStarted (want true): %t, please check if there is concurrent clusterStagedUpdateRun",
```
zhiying-lin: >please check if there is concurrent clusterStagedUpdateRun

do you mean the concurrent updateRuns for the same CRP?
Should we fail the second clusterStagedUpdateRun when we init it?

pkg/controllers/updaterun/execution.go
```diff
@@ -419,25 +421,30 @@ func (r *Reconciler) updateApprovalRequestAccepted(ctx context.Context, appReq *
 	return nil
 }
 
-// isBindingSyncedWithClusterStatus checks if the binding is up-to-date with the cluster status.
-func isBindingSyncedWithClusterStatus(resourceSnapshotName string, updateRun *placementv1beta1.ClusterStagedUpdateRun, binding *placementv1beta1.ClusterResourceBinding, cluster *placementv1beta1.ClusterUpdatingStatus) bool {
+// getBindingDiffFromClusterStatus checks if the binding is up-to-date with the cluster status.
+// It returns an empty string if the binding is up-to-date, otherwise a string about detailed difference message.
+func getBindingDiffFromClusterStatus(resourceSnapshotName string, updateRun *placementv1beta1.ClusterStagedUpdateRun, binding *placementv1beta1.ClusterResourceBinding, cluster *placementv1beta1.ClusterUpdatingStatus) string {
```
zhiying-lin:
```suggestion
func reportStatusDiff(resourceSnapshotName string, updateRun *placementv1beta1.ClusterStagedUpdateRun, binding *placementv1beta1.ClusterResourceBinding, cluster *placementv1beta1.ClusterUpdatingStatus) string {
```
How about reportStatusDiff? when i looked at the getBindingDiffFromClusterStatus, i thought you would like to return a binding.

pkg/controllers/updaterun/execution.go
zhiying-lin: do we have such e2e tests? if not, can we add it in a separate PR?

pkg/controllers/updaterun/execution.go
jwtty: hmmm I thought about this. But it may not necessarily a user error either. It can be a code bug. Should I just not wrap it?

pkg/controllers/updaterun/execution.go
jwtty: Yes, concurrent updateRuns for the same CRP, I'll update the log. Actually we allow multiple updateRuns just like multiple release pipelines. And in this case, the 2nd preempts the 1st one and the 1st one fails.

We are going to add some validations for concurrent updateRuns. I'll add e2e tests with that PR.

pkg/controllers/updaterun/execution.go
```diff
@@ -122,7 +122,7 @@ func (r *Reconciler) executeUpdatingStage(
 		binding := toBeUpdatedBindingsMap[clusterStatus.ClusterName]
 		if !condition.IsConditionStatusTrue(clusterStartedCond, updateRun.Generation) {
 			// The cluster has not started updating yet.
-			if !isBindingSyncedWithClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus) {
+			if diff := getBindingDiffFromClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus); diff != "" {
```
ryanzhang-oss: where is the diff used?  It seems that we called the getBindingDiffFromClusterStatus function again later?

pkg/controllers/updaterun/execution_test.go
```diff
@@ -26,24 +26,24 @@ import (
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
 )
 
-func TestIsBindingSyncedWithClusterStatus(t *testing.T) {
+func TestGetBindingDiffFromClusterStatus(t *testing.T) {
```
ryanzhang-oss: is there a UT for exec function?

pkg/controllers/updaterun/execution.go
```diff
@@ -170,13 +170,14 @@ func (r *Reconciler) executeUpdatingStage(
...
-			unexpectedErr := controller.NewUnexpectedBehaviorError(fmt.Errorf("the updating cluster `%s` in the stage %s does not match the cluster status: %+v, binding: %+v, condition: %+v",
-				clusterStatus.ClusterName, updatingStageStatus.StageName, clusterStatus, binding.Spec, binding.GetCondition(string(placementv1beta1.ResourceBindingRolloutStarted))))
-			klog.ErrorS(unexpectedErr, "The binding has been changed during updating, please check if there's concurrent clusterStagedUpdateRun", "clusterStagedUpdateRun", updateRunRef)
-			markClusterUpdatingFailed(clusterStatus, updateRun.Generation, unexpectedErr.Error())
-			return 0, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
+		if insync, rolloutStarted := isBindingSyncedWithClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus), condition.IsConditionStatusTrue(meta.FindStatusCondition(binding.Status.Conditions, string(placementv1beta1.ResourceBindingRolloutStarted)), binding.Generation); !insync || binding.Spec.State != placementv1beta1.BindingStateBound || !rolloutStarted {
```
zhiying-lin:
```suggestion
                isSync := isBindingSyncedWithClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus)
		rolloutStarted := condition.IsConditionStatusTrue(meta.FindStatusCondition(binding.Status.Conditions, string(placementv1beta1.ResourceBindingRolloutStarted)), binding.Generation)
		if !isSync || binding.Spec.State != placementv1beta1.BindingStateBound || !rolloutStarted {
```
The line is so long and creating the variable will help the readability.

pkg/controllers/updaterun/execution.go
```diff
@@ -170,13 +170,14 @@ func (r *Reconciler) executeUpdatingStage(
...
+			// This issue mostly happens when there are concurrent updateRuns referencing the same clusterResourcePlacement but releasing different versions.
+			// After the 1st updateRun updates the binding, and before the controller re-checks the binding status, the 2nd updateRun updates the same binding, and thus the 1st updateRun is preempted and observes the binding not matching the desired state.
+			preemptedErr := controller.NewUserError(fmt.Errorf("the clusterResourceBinding of the updating cluster `%s` in the stage `%s` is not up-to-date with the desired status, "+
+				"please check the binding status and see if there is a concurrent updateRun referencing the same clusterResourcePlacement and updating the same binding", clusterStatus.ClusterName, updatingStageStatus.StageName))
+			klog.ErrorS(preemptedErr, "The binding has been changed during updating", "binding spec insync", insync, "binding state", binding.Spec.State, "binding RolloutStarted", rolloutStarted, "binding", klog.KObj(binding), "clusterStagedUpdateRun", updateRunRef)
```
zhiying-lin:
```suggestion
			klog.ErrorS(preemptedErr, "The binding has been changed during updating", 
			                 "binding spec iSsync", insync, "binding state", binding.Spec.State, "binding RolloutStarted", 
                                          rolloutStarted, "binding", klog.KObj(binding), "clusterStagedUpdateRun", updateRunRef)
```
https://go.dev/wiki/CodeReviewComments#line-length

pkg/controllers/updaterun/execution.go
jwtty: fixed, thanks

pkg/controllers/updaterun/execution.go
jwtty: Per offline discussion, we report it as a user error instead.

pkg/controllers/updaterun/execution.go
jwtty: Per offline discussion, we do not need to expose too many internal details to the users. And this function does not need change any more.

pkg/controllers/updaterun/execution.go
jwtty: Fixed, thanks!

pkg/controllers/updaterun/execution_test.go
jwtty: We only have IT for the exec function. I'll add more for concurrent updateRuns in a separate PR.

pkg/controllers/updaterun/execution.go
jwtty: Function restored.

zhiying-lin (approved):

PR #64, opened 2025-05-14
test: set the upgrade compatibility test pipeline to use commit-specific setup

michaelawyu:

This PR sets the  upgrade compatibility test pipeline to use commit-specific setup (i.e., use old helm chart with old image).

I have:

N/A

N/A


test/upgrade/setup.sh
```diff
@@ -52,18 +35,8 @@ TAG=$IMAGE_TAG make docker-build-hub-agent
...
-    git checkout $PREVIOUS_COMMIT
-    echo "Checked out source code at $PREVIOUS_COMMIT."
-fi
-
-echo "Switch back to the test/upgrade directory to avoid consistency issues."
-cd test/upgrade
-
 # Create the kind clusters.
-echo "Creating the kind clusters..."
+echo "Creatinå the kind clusters..."
```
ryanzhang-oss:
```suggestion
echo "Creating the kind clusters..."
```

.github/workflows/upgrade.yml
```diff
@@ -56,22 +56,84 @@ jobs:
...
         env:
           KUBECONFIG: '/home/runner/.kube/config'
           HUB_SERVER_URL: 'https://172.19.0.2:6443'
+      
+      - name: Travel to the current state
+        # Note: Fleet always uses the version compatibility test suite from the 
+        # baseline commit, i.e., the commit that triggers the workflow.
+        run: |
+          echo "Returning to the current state..."
+          git checkout $PREVIOUS_COMMIT
```
ryanzhang-oss, line 98: why is this not current commit?

.github/workflows/upgrade.yml
```diff
@@ -56,22 +56,84 @@ jobs:
       - name: Set up Ginkgo CLI
         run: |
           go install github.com/onsi/ginkgo/v2/ginkgo@v2.19.1
+      
+      - name: Travel back in time to the before upgrade version
+        run: |
+          GIT_TAG="${{ github.event.inputs.beforeTagOrCommit }}"
+          PREVIOUS_BRANCH=$(git branch --show-current)
```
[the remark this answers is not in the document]
ryanzhang-oss: Second this, there are a lot of duplicated code here

test/upgrade/setup.sh
michaelawyu: Fixed.

test/upgrade/setup.sh
```diff
@@ -154,3 +127,5 @@ do
         --set enableV1Alpha1APIs=false \
         --set enableV1Beta1APIs=true
 done
+
+echo "Setup fo the before upgrade environment has been completed."
```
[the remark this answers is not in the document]
michaelawyu: Done.

.github/workflows/upgrade.yml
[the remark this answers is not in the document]
michaelawyu: Hi Ryan! This part is a bit tricky unfortunately. The commands manipulate the current state of the repo; as a script file it would mean that once traveling back the script would be gone (I could add a step that copies out the script as a untracked file, but that would be a bit weird I guess -> but if it is OK with you I could make the changes); for the separate GitHub Action proposal, it would require a separate public repo dedicated to the Action as required by GitHub, which is also tricky.

.github/workflows/upgrade.yml
michaelawyu: Hi Ryan! The upgrade pipeline often runs on dangling commits (commits that have not been checked as the PR hasn't been merged) so we couldn't just jump to the HEAD.

ryanzhang-oss (approved):

PR #66, opened 2025-05-16
feat: watch delete event for ClusterApprovalRequest & reconcile CSUR 

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -275,7 +275,7 @@ func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
 			},
 			//
 			DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
-				klog.V(2).InfoS("Handling a clusterApprovalRequest update event", "clusterApprovalRequest", klog.KObj(e.Object))
+				klog.V(2).InfoS("Handling a clusterApprovalRequest delete event", "clusterApprovalRequest", klog.KObj(e.Object))
 				handleClusterApprovalRequestDelete(e.Object, q)
 			},
 		}).Complete(r)
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -268,17 +268,22 @@ func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
 		Named("clusterresource-stagedupdaterun-controller").
 		For(&placementv1beta1.ClusterStagedUpdateRun{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Watches(&placementv1beta1.ClusterApprovalRequest{}, &handler.Funcs{
-			// We only care about when an approval request is approved.
+			// We watch for ClusterApprovalRequest to be approved.
 			UpdateFunc: func(ctx context.Context, e event.UpdateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 				klog.V(2).InfoS("Handling a clusterApprovalRequest update event", "clusterApprovalRequest", klog.KObj(e.ObjectNew))
-				handleClusterApprovalRequest(e.ObjectOld, e.ObjectNew, q)
+				handleClusterApprovalRequestUpdate(e.ObjectOld, e.ObjectNew, q)
+			},
+			//
+			DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+				klog.V(2).InfoS("Handling a clusterApprovalRequest update event", "clusterApprovalRequest", klog.KObj(e.Object))
+				handleClusterApprovalRequestDelete(e.Object, q)
 			},
 		}).Complete(r)
 }
 
-// handleClusterApprovalRequest finds the ClusterStagedUpdateRun creating the ClusterApprovalRequest,
+// handleClusterApprovalRequestUpdate finds the ClusterStagedUpdateRun creating the ClusterApprovalRequest,
 // and enqueues it to the ClusterStagedUpdateRun controller queue only when the approved condition is changed.
-func handleClusterApprovalRequest(oldObj, newObj client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+func handleClusterApprovalRequestUpdate(oldObj, newObj client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 	oldAppReq, ok := oldObj.(*placementv1beta1.ClusterApprovalRequest)
 	if !ok {
 		klog.V(2).ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("cannot cast runtime object to ClusterApprovalRequest")),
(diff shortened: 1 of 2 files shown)


pkg/controllers/updaterun/controller.go
```diff
@@ -268,17 +268,22 @@ func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
...
 		For(&placementv1beta1.ClusterStagedUpdateRun{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Watches(&placementv1beta1.ClusterApprovalRequest{}, &handler.Funcs{
-			// We only care about when an approval request is approved.
+			// We watch for ClusterApprovalRequest to be approved.
 			UpdateFunc: func(ctx context.Context, e event.UpdateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
 				klog.V(2).InfoS("Handling a clusterApprovalRequest update event", "clusterApprovalRequest", klog.KObj(e.ObjectNew))
-				handleClusterApprovalRequest(e.ObjectOld, e.ObjectNew, q)
+				handleClusterApprovalRequestUpdate(e.ObjectOld, e.ObjectNew, q)
+			},
+			// We watch for ClusterApprovalRequest deletion events to recreate it ASAP.
```
ryanzhang-oss, line 276: why do u we care if the approval is already approved? do we not care if it's already approved or not?

pkg/controllers/updaterun/controller.go
```diff
@@ -312,6 +317,34 @@ func handleClusterApprovalRequest(oldObj, newObj client.Object, q workqueue.Type
...
+func handleClusterApprovalRequestDelete(obj client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	appReq, ok := obj.(*placementv1beta1.ClusterApprovalRequest)
+	if !ok {
+		klog.V(2).ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("cannot cast runtime object to ClusterApprovalRequest")),
+			"Invalid object type", "object", klog.KObj(obj))
+		return
+	}
+	approved := condition.IsConditionStatusTrue(meta.FindStatusCondition(appReq.Status.Conditions, string(placementv1beta1.ApprovalRequestConditionApproved)), appReq.Generation)
+	approvalAccepted := condition.IsConditionStatusTrue(meta.FindStatusCondition(appReq.Status.Conditions, string(placementv1beta1.ApprovalRequestConditionApprovalAccepted)), appReq.Generation)
+	if approved && approvalAccepted {
```
jwtty: We only need `approvalAccepted` here: `approved` condition is ignored when `approvalAccepted` condition is set.

pkg/controllers/updaterun/controller.go
```diff
@@ -312,6 +317,34 @@ func handleClusterApprovalRequest(oldObj, newObj client.Object, q workqueue.Type
...
+		klog.V(2).InfoS("The approval request has been approved and accepted, ignore queueing for delete event", "clusterApprovalRequest", klog.KObj(appReq))
+		return
+	}
+
+	updateRun := appReq.Spec.TargetUpdateRun
```
jwtty, line 336: In the event of updateRun deletion, the controller deletes all associated `clusterApprovalRequests`. Will these deletions trigger updateRun reconcile again? If it's yes, we can ignore this here by checking if updateRun has deletionTimestamp set.

Arvindthiru: Couple of concerns,

- We don't have the client object here to access the updateRun to check deletionTimestamp
- https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/controllers/updaterun/controller.go#L82-L93 this should handle another update run reconcile if it's deleting  right ?

pkg/controllers/updaterun/controller.go
```diff
@@ -312,6 +317,34 @@ func handleClusterApprovalRequest(oldObj, newObj client.Object, q workqueue.Type
...
+func handleClusterApprovalRequestDelete(obj client.Object, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	appReq, ok := obj.(*placementv1beta1.ClusterApprovalRequest)
+	if !ok {
+		klog.V(2).ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("cannot cast runtime object to ClusterApprovalRequest")),
+			"Invalid object type", "object", klog.KObj(obj))
+		return
+	}
+	isApproved := meta.FindStatusCondition(appReq.Status.Conditions, string(placementv1beta1.ApprovalRequestConditionApproved))
+	approvalAccepted := condition.IsConditionStatusTrue(meta.FindStatusCondition(appReq.Status.Conditions, string(placementv1beta1.ApprovalRequestConditionApprovalAccepted)), appReq.Generation)
+	if isApproved != nil && approvalAccepted {
```
jwtty, line 331: Why do we need to check if "isApproved != nil"?

pkg/controllers/updaterun/controller_test.go
```diff
@@ -312,3 +312,145 @@ func TestHandleClusterApprovalRequest(t *testing.T) {
...
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			queue := &controllertest.Queue{TypedInterface: workqueue.NewTypedRateLimitingQueue[reconcile.Request](
+				workqueue.DefaultTypedItemBasedRateLimiter[reconcile.Request]())}
+			handleClusterApprovalRequestDelete(tt.obj, queue)
+			if got := queue.Len() != 0; got != tt.shouldEnqueue {
+				t.Fatalf("handleClusterApprovalRequestDelete() shouldEnqueue test `%s` got %t, want %t", name, got, tt.shouldEnqueue)
```
jwtty: Do we need to put the name again in the error log? As t.Run() has the name input. Not sure how the error log looks like.

pkg/controllers/updaterun/execution_integration_test.go
```diff
@@ -972,6 +970,349 @@ var _ = Describe("UpdateRun execution tests", func() {
...
+			// Mark the deletion stage progressing condition as false with succeeded reason and add succeeded condition.
+			wantStatus.DeletionStageStatus.Conditions = append(wantStatus.DeletionStageStatus.Conditions, generateFalseProgressingCondition(updateRun, placementv1beta1.StageUpdatingConditionProgressing, true))
+			wantStatus.DeletionStageStatus.Conditions = append(wantStatus.DeletionStageStatus.Conditions, generateTrueCondition(updateRun, placementv1beta1.StageUpdatingConditionSucceeded))
+			// Mark updateRun progressing condition as false with succeeded reason and add succeeded condition.
+			wantStatus.Conditions[1] = generateFalseProgressingCondition(updateRun, placementv1beta1.StagedUpdateRunConditionProgressing, true)
+			wantStatus.Conditions = append(wantStatus.Conditions, generateTrueCondition(updateRun, placementv1beta1.StagedUpdateRunConditionSucceeded))
+			// Need to have a longer wait time for the test to pass, because of the long wait time specified in the update strategy.
+			timeout = time.Second * 90
```
jwtty, line 1263: What if defining a longer timeout const instead of modifying exsiting one?

pkg/controllers/updaterun/execution_integration_test.go
```diff
@@ -972,6 +970,349 @@ var _ = Describe("UpdateRun execution tests", func() {
...
+					return fmt.Errorf("approvalRequest has different spec (-want +got):\n%s", diff)
+				}
+				if diff := cmp.Diff(wantApprovalRequest.Labels, approvalRequest.Labels); diff != "" {
+					return fmt.Errorf("approvalRequest has different labels (-want +got):\n%s", diff)
+				}
+				return nil
+			}, timeout, interval).Should(Succeed(), "failed to validate the approvalRequest")
+
+			By("Approving the approvalRequest")
+			approveClusterApprovalRequest(ctx, wantApprovalRequest.Name)
```
jwtty, line 1229: Before approving the request, try first delete it and verify it's recreated instantly?

pkg/controllers/updaterun/execution_integration_test.go
```diff
@@ -972,6 +970,349 @@ var _ = Describe("UpdateRun execution tests", func() {
 	})
 })
 
+var _ = Describe("UpdateRun execution tests - delete ClusterApprovalRequest, don't recreate", func() {
```
jwtty, line 939: Thank you for writing this test! We should definitely figure out a way to make this test more extensible. Right now it's too difficult to modify.

pkg/controllers/updaterun/controller.go
Arvindthiru: It's a strict check to ensure ApprovalRequest is approved with either true/false(as setting it to false is ignored once approvalAccepted condition is added) and then approvalAccepted condition is set to true

pkg/controllers/updaterun/execution_integration_test.go
Arvindthiru: The best way is to modify https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/controllers/updaterun/validation_integration_test.go#L420-L442 to take timeout as an argument and change it across all files (adds a lot of bloat to the PR review at the moment)

jwtty: Okay, I'll make this configurable.

PR #72, opened 2025-05-21
feat: support multi-version reporting in CRP for updateRun

jwtty:

Fixes #

I have:


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -975,10 +976,125 @@ func (r *Reconciler) setPlacementStatus(
...
 
+func (r *Reconciler) determineRolloutStateForCRPWithExternalRolloutStrategy(
+	ctx context.Context,
+	crp *fleetv1beta1.ClusterResourcePlacement,
+	selected []*fleetv1beta1.ClusterDecision,
+	allRPS []fleetv1beta1.ResourcePlacementStatus,
+	selectedResourceIDs []fleetv1beta1.ResourceIdentifier,
+) (bool, error) {
+	if len(selected) == 0 {
+		// This should not happen as we already checked in setPlacementStatus.
```
ryanzhang-oss, line 1001: then this should be an unexpected error?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -975,10 +976,125 @@ func (r *Reconciler) setPlacementStatus(
...
+		crp.Status.ObservedResourceIndex = ""
+		crp.Status.SelectedResources = []fleetv1beta1.ResourceIdentifier{}
+		crp.SetConditions(metav1.Condition{
+			Type:               string(fleetv1beta1.ClusterResourcePlacementRolloutStartedConditionType),
+			Status:             metav1.ConditionUnknown,
+			Reason:             condition.RolloutControlledByExternalControllerReason,
+			Message:            "Rollout is controlled by an external controller and different resource snapshot versions are observed across clusters",
+			ObservedGeneration: crp.Generation,
+		})
+		// As CRP status will refresh even if the spec has not changed, Fleet will reset any unused conditions
```
ryanzhang-oss:
```suggestion
		// As CRP status will refresh even if the spec has not changed, we reset any unused conditions
```

pkg/utils/controller/controller.go
```diff
@@ -362,6 +363,96 @@ func FetchAllClusterResourceSnapshots(ctx context.Context, k8Client client.Clien
 	return resourceSnapshots, nil
 }
 
+// CollectResourceIdentifiersFromClusterResourceSnapshot collects the resource identifiers selected by a series of clusterResourceSnapshot.
+// Given the index of the clusterResourceSnapshot, it collects resources from all of the master snapshots as well as
+func CollectResourceIdentifiersFromClusterResourceSnapshot(
+	ctx context.Context,
+	k8Client client.Client,
+	crp *fleetv1beta1.ClusterResourcePlacement,
```
ryanzhang-oss: nit, we actually only need the CRP name

pkg/utils/controller/controller.go
```diff
@@ -362,6 +363,96 @@ func FetchAllClusterResourceSnapshots(ctx context.Context, k8Client client.Clien
...
+		fleetv1beta1.ResourceIndexLabel: resourceSnapshotIndex,
+	}
+	resourceSnapshotList := &fleetv1beta1.ClusterResourceSnapshotList{}
+	if err := k8Client.List(ctx, resourceSnapshotList, labelMatcher); err != nil {
+		klog.ErrorS(err, "Failed to list the clusterResourceSnapshots associated with the clusterResourcePlacement",
+			"clusterResourcePlacement", crp.Name, "resourceSnapshotIndex", resourceSnapshotIndex)
+		return nil, NewAPIServerError(true, err)
+	}
+
+	if len(resourceSnapshotList.Items) == 0 {
```
ryanzhang-oss, line 385: this might not be an error as we never forbid this scenario (and might need to support this as a pure scheduling API)

pkg/utils/controller/controller.go
```diff
@@ -362,6 +363,96 @@ func FetchAllClusterResourceSnapshots(ctx context.Context, k8Client client.Clien
...
+	var masterResourceSnapshot *fleetv1beta1.ClusterResourceSnapshot
+	for i, resourceSnapshot := range resourceSnapshotList.Items {
+		// only master has this annotation
+		if len(resourceSnapshot.Annotations[fleetv1beta1.ResourceGroupHashAnnotation]) != 0 {
+			masterResourceSnapshot = &resourceSnapshotList.Items[i]
+			break
+		}
+	}
+	if masterResourceSnapshot == nil {
+		err := NewUserError(fmt.Errorf("no master clusterResourceSnapshot found for clusterResourcePlacement `%s`", crp.Name))
```
ryanzhang-oss: this not userError, we created the master snapshot

pkg/utils/controller/controller.go
```diff
@@ -362,6 +363,96 @@ func FetchAllClusterResourceSnapshots(ctx context.Context, k8Client client.Clien
...
+				return NewUnexpectedBehaviorError(err)
+			}
+			identifier := fleetv1beta1.ResourceIdentifier{
+				Group:     uResource.GetObjectKind().GroupVersionKind().Group,
+				Version:   uResource.GetObjectKind().GroupVersionKind().Version,
+				Kind:      uResource.GetObjectKind().GroupVersionKind().Kind,
+				Name:      uResource.GetName(),
+				Namespace: uResource.GetNamespace(),
+			}
+			if !selectedResourcesMap[identifier] {
```
ryanzhang-oss: when are we going to have dup resources in the same resource index?

pkg/utils/overrider/overrider_test.go
```diff
@@ -65,7 +65,7 @@ func TestFetchAllMatchingOverridesForResourceSnapshot(t *testing.T) {
 				Kind:    "Service",
 			}: true,
 			{
-				Group:   "",
+				Group:   "apps",
```
ryanzhang-oss, line 68: AFAIK, this wrong group name didn't break the test, right? The fix wasn't really related to this PR?

pkg/controllers/rollout/controller.go
```diff
@@ -158,8 +161,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	}
 	klog.V(2).InfoS("Found the latest resourceSnapshot for the clusterResourcePlacement", "clusterResourcePlacement", crpName, "latestResourceSnapshot", klog.KObj(latestResourceSnapshot))
 
-	// fill out all the default values for CRP just in case the mutation webhook is not enabled.
```
ryanzhang-oss, line 161: curious why it was set so late?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -975,10 +976,125 @@ func (r *Reconciler) setPlacementStatus(
...
+		// For external rollout strategy, if clusters observe different resource snapshot versions,
+		// we set RolloutStarted to Unknown without any other conditions since we do not know exactly which version is rolling out.
+		// We also need to reset ObservedResourceIndex and selectedResources.
+		rolloutStartedUnknown, err := r.determineRolloutStateForCRPWithExternalRolloutStrategy(ctx, crp, selected, allRPS, selectedResourceIDs)
+		if err != nil {
+			return true, err
+		}
+		if rolloutStartedUnknown {
+			return true, nil
+		}
```
ryanzhang-oss:
```suggestion
		if err != nil ||  rolloutStartedUnknown {
			return true, err
		}
```

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -975,10 +976,125 @@ func (r *Reconciler) setPlacementStatus(
 		return false, nil
 	}
 
+	if crp.Spec.Strategy.Type == fleetv1beta1.ExternalRolloutStrategyType {
```
ryanzhang-oss, line 979: so we still set crp.Status.ObservedResourceIndex = latestIndex even when the rolloutStragy is external?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -324,6 +366,15 @@ func (r *Reconciler) setResourcePlacementStatusPerCluster(
...
 
+	// For External rollout strategy, the per-cluster status is set to whatever exists on the binding.
+	if crp.Spec.Strategy.Type == fleetv1beta1.ExternalRolloutStrategyType {
+		status.ObservedResourceIndex = resourceSnapshotIndexOnBinding
+		setResourcePlacementStatusBasedOnBinding(crp, binding, status, expectedCondTypes, res)
+		return res
+	}
+
+	// TODO (wantjian): we only change the per-cluster status for External rollout strategy for now, so set the ObservedResourceIndex as the latest.
+	status.ObservedResourceIndex = latestResourceSnapshot.GetLabels()[fleetv1beta1.ResourceIndexLabel]
```
ryanzhang-oss, line 377: what does this TODO mean? Are we going to set per-cluster status for rollingUpdate too?

pkg/controllers/clusterresourceplacement/controller.go
jwtty: No, for external strategy, crp.Status.ObservedResourceIndex is empty when clusters observe different resource snapshot indices or the observed index when clusters observe the same.

pkg/controllers/clusterresourceplacement/controller.go
jwtty: good point, fixed!

pkg/controllers/clusterresourceplacement/placement_status.go
jwtty: Yes, per our original discussion, we show the version existing on the binding for both External and rollingUpdate. This PR does not change the rollingUpdate one as without portal update to display the `ObservedResourceIndex`, it will look very weird to users: the status is available but it's actually available on a not-shown older version.

pkg/utils/overrider/overrider_test.go
jwtty: I fixed the apiversion in `DeploymentResourceContentForTest` and these tests all use that so they will fail.

pkg/utils/controller/controller.go
jwtty: Actually no, we checked if there's dup in `gatherSelectedResource` when collecting resources and generating resourceSnapshots, no need to check again. Removed.

pkg/utils/controller/controller.go
jwtty: Thanks, fixed!

pkg/utils/controller/controller.go
jwtty: I see. Updated to returning an empty list instead with logging.

pkg/utils/controller/controller.go
jwtty: Fixed! Thanks.

pkg/controllers/rollout/controller.go
jwtty: I only found one issue when I was doing manual test in my own environment without webhook enabled. I removed the strategy type and it was not filled and remained empty and thus the controller did not reconcile it.

pkg/controllers/clusterresourceplacement/controller.go
jwtty: Fixed.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -975,10 +976,124 @@ func (r *Reconciler) setPlacementStatus(
...
+			return false, err
+		}
+		crp.Status.SelectedResources = selectedResources
+	}
+
+	for i := range len(selected) {
+		rolloutStartedCond := meta.FindStatusCondition(allRPS[i].Conditions, string(fleetv1beta1.ResourceRolloutStartedConditionType))
+		if !condition.IsConditionStatusTrue(rolloutStartedCond, crp.Generation) &&
+			!condition.IsConditionStatusFalse(rolloutStartedCond, crp.Generation) {
+			klog.V(2).InfoS("Placement has External rollout strategy and some cluster is in RolloutStarted Unknown state, set RolloutStarted condition to True",
```
ryanzhang-oss: this comment says "set RolloutStarted condition to True" but the code below actually sets the condition to be unknown?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -975,10 +976,124 @@ func (r *Reconciler) setPlacementStatus(
...
+				Status:             metav1.ConditionUnknown,
+				Reason:             condition.RolloutControlledByExternalControllerReason,
+				Message:            fmt.Sprintf("Rollout is controlled by an external controller and cluster %s is in RolloutStarted Unknown state", allRPS[i].ClusterName),
+				ObservedGeneration: crp.Generation,
+			})
+			// As CRP status will refresh even if the spec has not changed, we reset any unused conditions
+			// to avoid confusion.
+			for i := condition.RolloutStartedCondition + 1; i < condition.TotalCondition; i++ {
+				meta.RemoveStatusCondition(&crp.Status.Conditions, string(i.ClusterResourcePlacementConditionType()))
+			}
```
ryanzhang-oss, line 1090: why remove the rest of conditions in the crp.status when we are looking at a RPS[i] condition?

pkg/controllers/clusterresourceplacement/controller.go
jwtty: Good catch! The condition should be set to "Unknown" if some of the cluster is in rolloutStarted unknown state, comment is wrong.

pkg/controllers/clusterresourceplacement/controller.go
jwtty: No, this is saying if there's a member cluster whose RolloutStarted condition is Unknown, we set the CRP RolloutStarted condition to Unknown too and the other conditions like Overriden, Applied etc are removed.

test/e2e/enveloped_object_placement_test.go
```diff
@@ -278,7 +278,8 @@ var _ = Describe("placing wrapped resources using a CRP", func() {
 			PlacementStatuses := make([]placementv1beta1.ResourcePlacementStatus, 0)
 			for _, memberClusterName := range allMemberClusterNames {
 				unavailableResourcePlacementStatus := placementv1beta1.ResourcePlacementStatus{
-					ClusterName: memberClusterName,
+					ClusterName:           memberClusterName,
+					ObservedResourceIndex: "0",
```
ryanzhang-oss, line 282: why adding this? was the test broken?

test/e2e/actuals_test.go
```diff
@@ -813,6 +850,69 @@ func crpStatusWithWorkSynchronizedUpdatedFailedActual(
 	}
 }
 
+func crpStatusWithExternalStrategyActual(
+	wantSelectedResourceIdentifiers []placementv1beta1.ResourceIdentifier,
+	wantObservedResourceIndex string,
+	wantAvailable bool,
```
ryanzhang-oss, line 856: nit: this is used more like wantCRPRolloutCompleted

test/e2e/actuals_test.go
```diff
@@ -813,6 +850,69 @@ func crpStatusWithWorkSynchronizedUpdatedFailedActual(
...
+	wantSelectedClusters []string,
+	wantObservedResourceIndexPerCluster []string,
+	wantAvailablePerCluster []bool,
```
ryanzhang-oss, line 859: nit: this is used more like wantRolloutCompletedPerCluster

test/e2e/enveloped_object_placement_test.go
jwtty: Yes, now per-cluster resourcePlacementStatus reports `ObservedResourceIndex`, no matter whether it's rollingUpdate or updateRun.

test/e2e/actuals_test.go
jwtty: sure, will fix in next PR.

test/e2e/actuals_test.go
jwtty: sure, will fix in next PR.

PR #80, opened 2025-05-27
feat: add webhook configuration for clusterResourcePlacementDisruptionBudget

britaniar:

Fixes #

I have:


pkg/controllers/clusterresourceplacementeviction/controller_intergration_test.go
```diff
@@ -530,6 +530,259 @@ var _ = Describe("Test ClusterResourcePlacementEviction Controller", func() {
 			checkEvictionCompleteMetric(customRegistry, "false", "true")
 		})
 	})
+
+	It("Eviction Blocked - PickAll CRP, PDB with MaxUnavailable set as Percentage, eviction denied due to misconfigured PDB", func() {
```
Arvindthiru: All the three ITs have the same logic except the crpdb can we move the logic to a function and pass the crpdb to reuse code?

pkg/webhook/webhook.go
```diff
@@ -31,6 +31,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/kubefleet-dev/kubefleet/pkg/webhook/clusterresourceplacementdisruptionbudget"
```
Arvindthiru: nit: import order

pkg/webhook/webhook.go
```diff
@@ -380,6 +382,22 @@ func (w *Config) buildFleetValidatingWebhooks() []admv1.ValidatingWebhook {
...
+		{
+			Name:                    "fleet.clusterresourceplacementdisruptionbudget.validating",
+			ClientConfig:            w.createClientConfig(clusterresourceplacementdisruptionbudget.ValidationPath),
+			FailurePolicy:           &failFailurePolicy,
+			SideEffects:             &sideEffortsNone,
+			AdmissionReviewVersions: admissionReviewVersions,
+			Rules: []admv1.RuleWithOperations{
+				{
+					Operations: []admv1.OperationType{
+						admv1.Create,
```
Arvindthiru: We should also handle update case for CRPDB ?

test/e2e/utils_test.go
```diff
@@ -799,6 +799,20 @@ func cleanupCRP(name string) {
 	Eventually(removedActual, workloadEventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove CRP %s", name)
 }
 
+// cleanupCRPDB deletes the CRPDB and waits until the resource is not found.
+func cleanupCRPDB(name string) {
```
Arvindthiru: We already have this function in place, https://github.com/kubefleet-dev/kubefleet/blob/main/test/e2e/utils_test.go#L964-L973

test/e2e/webhook_test.go
```diff
@@ -1366,3 +1367,216 @@ var _ = Describe("webhook tests for ClusterResourcePlacementEviction CREATE oper
...
+				IntVal: 1,
+			}
+			crpdb.Spec.MinAvailable = nil // Clear MinAvailable to test MaxUnavailable
+			By(fmt.Sprintf("expecting denial of UPDATE disruption budget %s", crpName))
+			err := hubClient.Update(ctx, &crpdb)
+			if k8sErrors.IsConflict(err) {
+				return err
+			}
+			var statusErr *k8sErrors.StatusError
+			Expect(errors.As(err, &statusErr)).To(BeTrue(), fmt.Sprintf("Update CRPDB call produced error %s. Error type wanted is %s.", reflect.TypeOf(err), reflect.TypeOf(&k8sErrors.StatusError{})))
```
Arvindthiru: https://onsi.github.io/ginkgo/#testing-external-systems, need to use g.Expect or else Eventually won't retry,

From link attached,
"You may be wondering why we need to pass in a dedicated Gomega instance to the polled function. That's because the default global-level assertions are implicitly tied to Ginkgo's Fail handler. The first failed assertion in an Eventually would cause the spec to fail with no possibility to retry. By passing in a fresh Gomega instance, Eventually can monitor for failures itself and control the final failure/success state of the assertion it is governing."

Arvindthiru (approved):

PR #81, opened 2025-05-29
fix: fix updateRun stuck with reportDiff mode

jwtty:
1. Fix updateRun does not work with CRP with reportDiff mode.
2. Update metrics emit test to use `ctrlmetrics.Registry` instead.
3. Refactor updateRun IT to make it easier for testing.

Fixes #76 

I have:


pkg/utils/binding/binding.go
```diff
@@ -25,7 +25,7 @@ import (
 
 // HasBindingFailed checks if ClusterResourceBinding has failed based on its conditions.
 func HasBindingFailed(binding *placementv1beta1.ClusterResourceBinding) bool {
```
ryanzhang-oss: this function is used by rollout and CRP besides updateRun. The reportDiff condition are handled separately there. I haven't thought through if this change is safe for those cases. If yes, we should also modify those code. This is the core of the state machine transition logic so we need to be very careful.

pkg/controllers/updaterun/execution_integration_test.go
```diff
@@ -566,263 +527,469 @@ var _ = Describe("UpdateRun execution tests", func() {
...
+			validateUpdateRunMetricsEmitted(generateProgressingMetric(updateRun), generateStuckMetric(updateRun), generateFailedMetric(updateRun))
 		})
+	})
+})
 
-		It("Should mark the 3rd cluster in the 1st stage as succeeded after marking the binding available", func() {
-			By("Validating the 3rd clusterResourceBinding is updated to Bound")
-			binding := resourceBindings[numTargetClusters-5] // cluster-5
-			validateBindingState(ctx, binding, resourceSnapshot.Name, updateRun, 0)
+var _ = Describe("UpdateRun execution tests - fewer clusters with single stage", func() {
```
ryanzhang-oss: what does "fewer clusters with single stage" mean?

pkg/utils/binding/binding.go
jwtty: I decide not to change it and check if binding fails in reportDiff in updateRun.

pkg/controllers/updaterun/execution_integration_test.go
jwtty: I created a smaller test case with only 3 clusters and single stage making it easier to add tests, e.g. validate different types/combinations of after-stage tasks, reportDiff, and approvalRequest related tests.

pkg/controllers/clusterresourceplacementeviction/controller_test.go
```diff
@@ -1513,7 +1507,7 @@ func TestReconcileForIncompleteEvictionMetric(t *testing.T) {
 		t.Errorf("reconcile should have failed")
 	}
 
-	metricFamilies, err := customRegistry.Gather()
+	metricFamilies, err := ctrlmetrics.Registry.Gather()
```
ryanzhang-oss, line 1510: I assume that this is not related to the reportDiff change? Where was FleetEvictionStatus registered?

ryanzhang-oss (approved):

pkg/controllers/clusterresourceplacementeviction/controller_test.go
jwtty: When I write the UT, I realized this metrics UT needs fix. I fixed for both updateRun and eviction.

PR #83, opened 2025-06-02
feat: better handling of cost property calculation corner cases

michaelawyu:

This PR enables the Azure property provider to raise warnings/errors as conditions when handling cost calculation corner cases.

I have:

- [x] Unit tests
- [x] Integration tests


michaelawyu:
**Important**: there are a few additional PRs that will be sent separately to enable E2E and upgrade tests for functionalities in this PR.

Also -> this PR involves a condition type change; will send a separate PR to set up the cleanup process (with upgrade tests for verification).

pkg/propertyprovider/azure/provider_integration_test.go
```diff
@@ -172,29 +172,88 @@ var (
...
-					Properties: map[clusterv1beta1.PropertyName]clusterv1beta1.PropertyValue{
-						propertyprovider.NodeCountProperty: {
-							Value: fmt.Sprintf("%d", len(nodes)),
+				perCPUCoreCost := "0.0"
+				perGBMemoryCost := "0.0"
+				costCollectionWarnings := []string{}
+				var costCollectionErr error
+
+				switch {
+				case len(nodes) == 0:
```
britaniar, line 209: So we do nothing in this case?

britaniar: Do we want a log at least?

pkg/propertyprovider/azure/provider_integration_test.go
```diff
@@ -172,29 +172,88 @@ var (
...
-						return fmt.Errorf("on-demand price not found for SKU %s", node.Labels[trackers.AKSClusterNodeSKULabelName])
+						missingSKUSet[sku] = true
 					}
 					totalCost += cost
 				}
-				perCPUCost := fmt.Sprintf(CostPrecisionTemplate, totalCost/totalCPUCores)
-				perGBMemoryCost := fmt.Sprintf(CostPrecisionTemplate, totalCost/totalMemoryGBs)
+				missingSKUs := []string{}
+				for sku := range missingSKUSet {
+					missingSKUs = append(missingSKUs, sku)
```
britaniar, line 199: Seems like the integration test is failing around here. May need to sort so that the list is consistent.

pkg/propertyprovider/azure/trackers/nodes.go
```diff
@@ -129,10 +133,13 @@ func (nt *NodeTracker) calculateCosts() {
 
 	// Sum up the total costs.
 	totalHourlyRate := 0.0
+	missingSKUs := make([]string, 0)
 	for sku, ns := range nt.nodeSetBySKU {
 		hourlyRate, found := nt.pricingProvider.OnDemandPrice(sku)
```
ryanzhang-oss, line 162: the new lib will return a fake value instead of reporting notFound

pkg/propertyprovider/azure/provider_integration_test.go
michaelawyu: Yeah, it is considered normal (no nodes, no node related costs); we do have logs in the provider node controller to confirm on the no nodes situation, but adding a log here should help too -> will make the edit

pkg/propertyprovider/azure/provider_integration_test.go
michaelawyu: Yeah, my bad. I forgot that the tests now are comparing the output string verbatim -> will add the sort.

pkg/propertyprovider/azure/trackers/nodes.go
michaelawyu: Hi Ryan!

Their code,

```
// OnDemandPrice returns the last known on-demand price for a given instance type, returning false if there is no
// known on-demand pricing for the instance type.
func (p *Provider) OnDemandPrice(instanceType string) (float64, bool) {
	p.mu.RLock()
	defer p.mu.RUnlock()
	price, ok := p.onDemandPrices[instanceType]
	if !ok {
		// if we don't have a price, check if it's a known SKU with missing price
		if price, ok = skusWithMissingPrice[instanceType]; ok {
			return price, true
		}
		return 0.0, false
	}
	return price, true
}
```

only covers the known missing entries, so we still need to count for unknown cases our side.

michaelawyu: Though like you mentioned, Karpenter returning 999 would make it look quite weird on our side. We haven't upgraded to that version as I understand, but...

pkg/propertyprovider/azure/provider_integration_test.go
michaelawyu: A log entry has been added to the `calculateCosts()` function. (The current code selection is part of the integration test).

pkg/propertyprovider/azure/trackers/nodes.go
michaelawyu: The code has been updated to reflect the changes in the latest Karpenter/Az provider version.

pkg/propertyprovider/azure/provider.go
```diff
@@ -203,17 +202,20 @@ func (p *PropertyProvider) Collect(_ context.Context) propertyprovider.PropertyC
...
-		// track the previously returned condition. A timestamp will be added in the upper layer.
+	perCPUCost, perGBMemoryCost, warnings, err := p.nodeTracker.Costs()
+	switch {
+	case err != nil:
+		// An error occurred when calculating costs; do no set the cost properties and
+		// track the error.
 		conds = append(conds, metav1.Condition{
-			Type:    PropertyCollectionSucceededConditionType,
+			Type:    CostPropertiesCollectionSucceededCondType,
 			Status:  metav1.ConditionFalse,
```
ryanzhang-oss, line 212: what happens when this false?

pkg/propertyprovider/azure/provider.go
```diff
@@ -222,6 +224,29 @@ func (p *PropertyProvider) Collect(_ context.Context) propertyprovider.PropertyC
 			Value:           fmt.Sprintf(CostPrecisionTemplate, perGBMemoryCost),
 			ObservationTime: metav1.Now(),
 		}
+		conds = append(conds, metav1.Condition{
+			Type:    CostPropertiesCollectionSucceededCondType,
+			Status:  metav1.ConditionTrue,
+			Reason:  CostPropertiesCollectionDegradedReason,
```
ryanzhang-oss, line 230: since the property selector doesn't read this reason so it will take this value as IS? How does the user know?

pkg/propertyprovider/azure/provider.go
michaelawyu: Hi Ryan! When this is set to False, users will see the condition with the error details and no cost data will be exposed for the cluster (the properties themselves will not be added) -> if users have cost property based filters (e.g., pick clusters with cpu prices <1), such clusters will be ignored.

pkg/propertyprovider/azure/provider.go
michaelawyu: Hi Ryan! Yeah, this is consistent with the behavior before, where we would take the data returned from Karpenter as they are (even if they are stale), and if the cluster has some SKUs that cannot be recognized, we consider them to be free of cost.

michaelawyu: I totally understand the concern here of course; later we might be able to offer a way to filter out clusters with degraded properties... we do not support boolean typed properties yet, 0/1-based properties can do, but they might not look very good.

michaelawyu:
Note: sync'd offline with ryanzhangoss; the behavior has been slightly tweaked: considering that the system works best when handling homogeneous clusters (clusters using nodes of exactly one SKU), the property provider will now report partially missing SKUs as an error (i.e., no costs will be reported; the scheduler will ignore such clusters when processing cost-related queries) rather than returning skewed cost data.

ryanzhang-oss (approved):

pkg/propertyprovider/azure/provider.go
```diff
@@ -55,14 +55,13 @@ const (
...
 	// The condition related values in use by the Azure property provider.
-
-	// PropertyCollectionSucceededConditionType is a condition type that indicates whether a
-	// property collection attempt has succeeded.
-	PropertyCollectionSucceededConditionType         = "AKSClusterPropertyCollectionSucceeded"
-	PropertyCollectionSucceededReason                = "AllPropertiesCollectedSuccessfully"
-	PropertyCollectionFailedCostErrorReason          = "FailedToCollectCosts"
-	PropertyCollectionSucceededMessage               = "All properties have been collected successfully"
-	PropertyCollectionFailedCostErrorMessageTemplate = "An error has occurred when collecting cost properties: %v"
+	CostPropertiesCollectionSucceededCondType   = "AKSClusterCostPropertiesCollectionSucceeded"
```
ryanzhang-oss, line 58: I didn't notice the condition name is changed...

PR #84, opened 2025-06-03
feat: add unify interface between cluster scoped and namespace scoped resources

ryanzhang-oss:

add unify interface between cluster scoped and namespace scoped resources to be the foundation of the refactor

Fixes #

I have:


apis/interface.go
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 package apis
 
 import (
+	"github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
```
zhiying-lin, line 21: another group?

apis/placement/v1beta1/binding_types.go
```diff
@@ -263,6 +263,30 @@ func (b *ClusterResourceBinding) GetCondition(conditionType string) *metav1.Cond
 	return meta.FindStatusCondition(b.Status.Conditions, conditionType)
 }
 
+// GetBindingSpec returns the binding spec.
+func (b *ClusterResourceBinding) GetBindingSpec() *ResourceBindingSpec {
+	return &b.Spec
+}
+
+// SetBindingSpec sets the binding spec.
```
zhiying-lin, line 271: Should we keep it simple, just set the spec, otherwise, the func name should be SetBindingSpecIfNil?

btw, should we use the deep copy in case there are the multiple callers updating the same spec?

apis/interface.go
```diff
@@ -36,3 +37,91 @@ type ConditionedObj interface {
 	client.Object
 	Conditioned
 }
+
+// A PlacementSpec contains placementSpec
```
zhiying-lin, line 41:
```suggestion
// PlacementSpec offers the functionality to work with the placementSpec.
```

apis/interface.go
```diff
@@ -36,3 +37,91 @@ type ConditionedObj interface {
...
+}
+
+// A PlacementStatus contains placementStatus
+// +kubebuilder:object:generate=false
+type PlacementStatus interface {
+	GetPlacementStatus() *v1beta1.PlacementStatus
+	SetPlacementStatus(*v1beta1.PlacementStatus)
+}
+
+// A PlacementObj is for kubernetes resource placement object.
```
zhiying-lin, line 55:
```suggestion
// PlacementObj offers the functionality to work with kubernetes resource placement object.
```

PR #85, opened 2025-06-04
fix: address a git time travel issue in the upgrade tests

michaelawyu:

This PR fixes a git time travel issue in the upgrade compatibility tests.

I have:

N/A

N/A

### DIFF
--- a/.github/workflows/upgrade.yml
+++ b/.github/workflows/upgrade.yml
@@ -8,10 +8,7 @@ on:
   workflow_dispatch:
     inputs:
       beforeTagOrCommit:
-        description: 'The baseline tag or commit to build Fleet agents from; if not specified, the latest tag will be used'     
-        required: false
-      afterTagOrCommit:
-        description: 'The new tag or commit to build Fleet agents from; if not specified, the last commit in the codebase (PR or branch) will be used'     
+        description: 'The tag or commit to build the before upgrade version of the Fleet agents from; if not specified, the latest tag will be used'     
         required: false
   pull_request:
     branches:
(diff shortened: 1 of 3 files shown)


test/upgrade/setup.sh
```diff
@@ -63,7 +63,7 @@ done
 
 # Install the hub agent to the hub cluster.
 kind export kubeconfig --name $HUB_CLUSTER
-helm install hub-agent ../../charts/hub-agent/ \
+helm install hub-agent charts/hub-agent/ \
```
ryanzhang-oss, line 66: how did this work if the dir is not correct?

michaelawyu: Hi Ryan! It was working because the setup part used to travel back in time so it runs using the old version, which involves steps that jump between the root dir and the `/test/upgrade` dir.

I realized last night it's a bit difficult to reason about, so this PR updates the behavior to become more consistent in the sense that we still leverage the old source code but all the upgrade test related scripts are from the latest commit.

.github/workflows/upgrade.yml
```diff
@@ -85,55 +81,32 @@ jobs:
           echo "Checked out source code at $GIT_TAG."
 
       - name: Prepare the fleet using the before upgrade version
-        run: cd test/upgrade && chmod +x setup.sh && ./setup.sh 3 && cd -
+        # Set up the Fleet using images built the older source code but with the current setup script.
+        run: |
+          git checkout $PREVIOUS_COMMIT -- test/upgrade
```
Arvindthiru: Why are we checking out to the $PREVIOUS_COMMIT, it holds the latest branch's commit,

From previous step,
PREVIOUS_BRANCH=$(git branch --show-current)
PREVIOUS_COMMIT=$(git rev-parse HEAD)

nit: the var name is confusing

michaelawyu: Oh, `git checkout COMMIT -- dir` will check out the current version of the code for the specific dir (without touching other parts of the directory -> this is needed so that we could always use the scripts from now instead of from before). Let me tweak the var name and add a comment to be more specific.

Arvindthiru (approved):
LGTM

PR #94, opened 2025-06-06
fix: update unexpected error and API Server log error types

britaniar:

Fixes #

I have:
- updated the error wrap functions to ensure APIServer errors are not being categorized as unexpected errors

### DIFF
--- a/pkg/utils/controller/controller.go
+++ b/pkg/utils/controller/controller.go
@@ -87,6 +87,8 @@ func NewExpectedBehaviorError(err error) error {
 // NewAPIServerError returns error types when accessing data from cache or API server.
 func NewAPIServerError(fromCache bool, err error) error {
 	if err != nil {
+		// The func may return other unexpected runtime errors other than API server errors.
+		// https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/client.go#L334-L339
 		if fromCache && isUnexpectedCacheError(err) {
 			return NewUnexpectedBehaviorError(err)
 		}
@@ -98,8 +100,9 @@ func NewAPIServerError(fromCache bool, err error) error {
 
 func isUnexpectedCacheError(err error) bool {
 	// may need to add more error code based on the production
-	// Cache will return notFound for GET.
-	return !apierrors.IsNotFound(err)
+	// When the cache is missed, it will query API server and return API server errors.
+	var statusErr *apierrors.StatusError
+	return !errors.Is(err, context.Canceled) && !errors.As(err, &statusErr) && !errors.Is(err, context.DeadlineExceeded)
 }
 
 // NewUserError returns ErrUserError type error when err is not nil.
--- a/pkg/utils/controller/controller_test.go
+++ b/pkg/utils/controller/controller_test.go
@@ -111,10 +111,10 @@ func TestNewAPIServerError(t *testing.T) {
 			wantErr:   ErrAPIServerError,
 		},
 		{
-			name:      "reading from cache: unexpectedBehaviorError",
+			name:      "reading from cache: apiServerError",
 			fromCache: true,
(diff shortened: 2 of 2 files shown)


pkg/utils/controller/controller.go
```diff
@@ -69,6 +69,9 @@ var (
 // NewUnexpectedBehaviorError returns ErrUnexpectedBehavior type error when err is not nil.
 func NewUnexpectedBehaviorError(err error) error {
 	if err != nil {
+		if !isUnexpectedCacheError(err) {
+			return NewAPIServerError(false, err)
```
jwtty: Not sure I'm following: `NewAPIServerError` calls `isUnexpectedCacheError` when reading from cache. Why do we need such check? And why would we call `NewUnexpectedBehaviorError` when encountering an apiServer error?

pkg/utils/controller/controller.go
```diff
@@ -99,7 +102,8 @@ func NewAPIServerError(fromCache bool, err error) error {
 func isUnexpectedCacheError(err error) bool {
 	// may need to add more error code based on the production
 	// Cache will return notFound for GET.
-	return !apierrors.IsNotFound(err)
+	var statusErr *apierrors.StatusError
+	return !errors.Is(err, context.Canceled) && !errors.As(err, &statusErr) && !errors.Is(err, context.DeadlineExceeded)
```
jwtty, line 105: All `apierrors.StatusError` are NOT categorized to `isUnexpectedCacheError`?

pkg/utils/controller/controller.go
britaniar: We have recently hit errors where an APIServer error has been wrapped by an unexpected error. I decided to add a check here for when we only call NewUnexpectedBehaviorError to ensure we don't emit unexpectedErrors for APIServer errors.

zhiying-lin: Can we fix the place where we wrap the api server as unexpected error? we should not wrap any api server error as the unexpected one.

pkg/utils/controller/controller.go
```diff
@@ -87,9 +90,6 @@ func NewExpectedBehaviorError(err error) error {
 // NewAPIServerError returns error types when accessing data from cache or API server.
 func NewAPIServerError(fromCache bool, err error) error {
 	if err != nil {
```
zhiying-lin, line 89: hi Britania, now i recall why i defined the unexpectedCache error when fromCache is true,

https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/client.go#L334-L339 the func may return other unexpected runtime errors other than api server error.

Could you keep this and a comment to record the reason (easily forget why we specially handle this)? and update the isUnexpectedCacheError as you did.

pkg/utils/controller/controller.go
zhiying-lin: my bad, sorry for the confusion. Could you please remove this?

pkg/utils/controller/controller.go
zhiying-lin: Please add a comment to record the history.

when the cache is missed, it will query the API server and any api server errors are expected.

pkg/utils/controller/controller_test.go
```diff
@@ -128,6 +128,18 @@ func TestNewAPIServerError(t *testing.T) {
 			err:       apierrors.NewConflict(schema.GroupResource{}, "conflict", nil),
 			wantErr:   ErrAPIServerError,
 		},
+		{
```
zhiying-lin, line 131: add these two cases when the fromCache is true

add a case when fromCache is true and input err: runtime.NewMissingKindErr("unstructured object has no kind"), which is possible when reading from the cache

zhiying-lin (approved):

PR #95, opened 2025-06-06
fix: fix updateRun handling unscheduled bindings

jwtty:
UpdateRun should not report an unscheduled binding with latest policy snapshot as unexpected: this can happen when a member cluster leaves the cluster or a CRP with pickN policy and scaling down. In these cases, the unscheduled bindings should be cleaned up.

Fixes #

I have:

### DIFF
--- a/test/e2e/updaterun_test.go
+++ b/test/e2e/updaterun_test.go
@@ -26,6 +26,7 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	placementv1alpha1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1alpha1"
@@ -120,7 +121,7 @@ var _ = Describe("test CRP rollout with staged update run", func() {
 			createStagedUpdateRunSucceed(updateRunNames[0], crpName, resourceSnapshotIndex1st, strategyName)
 		})
 
-		It("Should rollout resources to member-cluster-2 only and completes stage canary", func() {
+		It("Should rollout resources to member-cluster-2 only and complete stage canary", func() {
 			checkIfPlacedWorkResourcesOnMemberClustersInUpdateRun([]*framework.Cluster{allMemberClusters[1]})
 			checkIfRemovedWorkResourcesFromMemberClustersConsistently([]*framework.Cluster{allMemberClusters[0], allMemberClusters[2]})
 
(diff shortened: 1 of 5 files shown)


test/e2e/updaterun_test.go
```diff
@@ -444,6 +445,193 @@ var _ = Describe("test CRP rollout with staged update run", func() {
...
 
+	Context("Test cluster scale out and shrink using pickN policy with staged update run", Ordered, func() {
+		var strategy *placementv1beta1.ClusterStagedUpdateStrategy
+		updateRunNames := []string{}
+
+		BeforeAll(func() {
+			// Create a test namespace and a configMap inside it on the hub cluster.
+			createWorkResources()
+
+			// Create the CRP with external rollout strategy and pick fixed policy.
```
ryanzhang-oss:
```suggestion
			// Create the CRP with external rollout strategy and pick N = 1
```

test/e2e/updaterun_test.go
```diff
@@ -444,6 +445,193 @@ var _ = Describe("test CRP rollout with staged update run", func() {
...
+			checkIfPlacedWorkResourcesOnMemberClustersInUpdateRun(allMemberClusters)
+		})
+
+		It("Should update crp status as completed", func() {
+			crpStatusUpdatedActual := crpStatusWithExternalStrategyActual(workResourceIdentifiers(), resourceSnapshotIndex1st, true, allMemberClusterNames,
+				[]string{resourceSnapshotIndex1st, resourceSnapshotIndex1st, resourceSnapshotIndex1st}, []bool{true, true, true}, nil, nil)
+			Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP %s status as expected", crpName)
+		})
+
+		It("Update the crp to only keep 2 clusters (member-cluster-2 and member-cluster-3)", func() {
```
ryanzhang-oss, line 588: I wonder why it is always 2,3 instead of 1,3 or 1,2?

test/e2e/updaterun_test.go
```diff
@@ -444,6 +445,193 @@ var _ = Describe("test CRP rollout with staged update run", func() {
...
+			Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP %s status as expected", crpName)
+		})
+
+		It("Should create a staged update run successfully", func() {
+			createStagedUpdateRunSucceed(updateRunNames[1], crpName, resourceSnapshotIndex1st, strategyName)
+		})
+
+		It("Should still have resources on member-cluster-2 and member-cluster-3 only and completes stage canary", func() {
+			checkIfPlacedWorkResourcesOnMemberClustersInUpdateRun(allMemberClusters[1:])
+			// TODO: need a way to check the status of staged update run that have member-cluster-2 and member-cluster-3 updated.
```
ryanzhang-oss: what do u mean by "updated"?

ryanzhang-oss (dismissed):

test/e2e/updaterun_test.go
jwtty: The 3 clusters should have the same score and are thus ordered by name. I believe this was designed to keep selection stable under same score scenario. So member 3 is always selected first, then member 2 and lastly member 1.

test/e2e/updaterun_test.go
jwtty: At this test point, updateRun is still running half way: only canary stage is completed. This TODO is to check updateRun status in this case. We currently only validate updateRun status after it completes in e2e. We validates partial-completed state in ITs.

test/e2e/updaterun_test.go
jwtty: good catch, thanks!

PR #97, opened 2025-06-09
fix: add some delays when creating resourceSnapshot

zhiying-lin:

add some delays when creating resourceSnapshot to avoid too frequent update.

Fixes #

I have:

- [X] Run `make reviewable` to ensure this PR is ready for review.

tested in UT


pkg/controllers/clusterinventory/clusterprofile/controller_test.go
```diff
@@ -159,13 +159,13 @@ func TestSyncClusterProfileCondition(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			reconciler.syncClusterProfileCondition(tt.memberCluster, tt.clusterProfile)
 			condition := meta.FindStatusCondition(tt.clusterProfile.Status.Conditions, clusterinventory.ClusterConditionControlPlaneHealthy)
-			if condition == nil {
+			if condition == nil { //nolint: staticcheck // false positive SA5011: possible nil pointer dereference
```
jwtty, line 162: Does this suddenly show up?

charts/hub-agent/README.md
```diff
@@ -19,24 +19,25 @@ _See [helm install](https://helm.sh/docs/helm/helm_install/) for command documen
...
+| ConcurrentResourceChangeSyncs     | The number of resourceChange reconcilers that are allowed to run concurrently.                                                                               | `20`                                             |
+| logFileMaxSize                    | Max size of log file before rotation                                                                                                                         | `1000000`                                        |
+| MaxFleetSizeSupported             | The max number of member clusters this fleet supports.                                                                                                       | `100`                                            |
+| resourceSnapshotCreationInterval  | The interval (in seconds) at which resource snapshots are created.                                                                                           | `1m`                                            |
```
jwtty: The example is not in seconds?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -490,14 +493,22 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
...
-	if latestResourceSnapshot != nil &&
-		latestResourceSnapshotHash != resourceHash &&
-		latestResourceSnapshot.Labels[fleetv1beta1.IsLatestSnapshotLabel] == strconv.FormatBool(true) {
+	if latestResourceSnapshot != nil && latestResourceSnapshotHash != resourceHash && latestResourceSnapshot.Labels[fleetv1beta1.IsLatestSnapshotLabel] == strconv.FormatBool(true) {
+		// When the latest resource snapshot without the isLastest label, it means it fails to create the new
+		// resource snapshot in the last reconcile and we don't need to check and delay the request.
+		if since := time.Since(latestResourceSnapshot.CreationTimestamp.Time); since < r.ResourceSnapshotCreationInterval {
+			// If the latest resource snapshot is created less than a minute ago, requeue the request to avoid
+			// too frequent update.
+			klog.V(2).InfoS("The latest resource snapshot is created less than a minute ago, skipping the update", "clusterResourcePlacement", crpKObj,
```
jwtty: "a minute" is hard coded.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -490,14 +493,22 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
...
+				"clusterResourceSnapshot", klog.KObj(latestResourceSnapshot), "creationTime", latestResourceSnapshot.CreationTimestamp, "configuredResourceSnapshotCreationInterval", r.ResourceSnapshotCreationInterval, "afterDuration", r.ResourceSnapshotCreationInterval-since)
+			return ctrl.Result{Requeue: true, RequeueAfter: r.ResourceSnapshotCreationInterval - since}, nil, nil
```
jwtty, line 504: A conceptual question: If there are 10 resource changes during one minute, I suppose this requeue would be called 10 times, is the CRP going to be reconciled 10 times after one minute, or just once as these requeues are combined?

pkg/controllers/clusterresourceplacement/placement_controllerv1alpha1.go
```diff
@@ -69,6 +69,10 @@ type Reconciler struct {
 	Recorder record.EventRecorder
 
 	Scheme *runtime.Scheme
+
+	// ResourceSnapshotCreationInterval is the interval to create a new resourcesnapshot
+	// to avoid too frequent updates.
+	ResourceSnapshotCreationInterval time.Duration
```
jwtty, line 75: Why do we need this change?

test/e2e/setup.sh
```diff
@@ -123,7 +123,8 @@ helm install hub-agent ../../charts/hub-agent/ \
     --set webhookClientConnectionType=service \
     --set forceDeleteWaitTime="1m0s" \
     --set clusterUnhealthyThreshold="3m0s" \
-    --set logFileMaxSize=1000000
+    --set logFileMaxSize=1000000 \
+    --set resourceSnapshotCreationInterval="0s"
```
jwtty: Will e2e fail if it's set to 1min? I wonder if it would be better to set it as 1min as this would be the default to users.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: in the queue, there will be only 1 key as they're using the same key and will be combined. After 1 min, the created snapshot will include these 10 resource changes.

pkg/controllers/clusterinventory/clusterprofile/controller_test.go
zhiying-lin: yes, it failed the lint check

charts/hub-agent/README.md
zhiying-lin: fixed

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: fixed

pkg/controllers/clusterresourceplacement/placement_controllerv1alpha1.go
zhiying-lin: The structure is shared by both v1alpha1 and v1beta1 CRP. I know the file name is so confusing. :(

test/e2e/setup.sh
zhiying-lin: yes, as we discussed in the meeting, it fails a lot of cases, and other side effect is our e2e will take longer. That's why Ryan suggested to make it configurable and set as 0 in our e2e tests.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: reference link, https://github.com/kubernetes/client-go/blob/4e82e684120e9c368762fafcae29bce7e4a63a62/util/workqueue/queue.go#L241

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -490,14 +493,22 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
 	// mark the last resource snapshot as inactive if it is different from what we have now or 3) when some
 	// sub-indexed cluster resource snapshots belonging to the same group have not been created, the master
 	// cluster resource snapshot should exist and be latest.
-	if latestResourceSnapshot != nil &&
-		latestResourceSnapshotHash != resourceHash &&
-		latestResourceSnapshot.Labels[fleetv1beta1.IsLatestSnapshotLabel] == strconv.FormatBool(true) {
+	if latestResourceSnapshot != nil && latestResourceSnapshotHash != resourceHash && latestResourceSnapshot.Labels[fleetv1beta1.IsLatestSnapshotLabel] == strconv.FormatBool(true) {
```
jwtty, line 496: Can we also add one integration test? Since our e2e test never tests the requeue behavior, I think it would be good to have: create a CRP -> verify one resource snapshot generated -> add/update some resources -> verify only one new resource snapshot generated and the difference between the two creationTimestamp is larger than the interval.

zhiying-lin: added e2e tests instead. 

To test the resourceSnapshot behavior in IT, we need to setup few other controllers, eg. resourceChange reconciler.

jwtty (approved):

charts/hub-agent/README.md
```diff
@@ -19,24 +19,25 @@ _See [helm install](https://helm.sh/docs/helm/helm_install/) for command documen
...
+| ConcurrentResourceChangeSyncs     | The number of resourceChange reconcilers that are allowed to run concurrently.                                                                               | `20`                                             |
+| logFileMaxSize                    | Max size of log file before rotation                                                                                                                         | `1000000`                                        |
+| MaxFleetSizeSupported             | The max number of member clusters this fleet supports.                                                                                                       | `100`                                            |
+| resourceSnapshotCreationInterval  | The interval at which resource snapshots are created.                                                                                           | `1m`                                            |
```
ryanzhang-oss, line 43: nit: the dividers don't seem to align

ryanzhang-oss (approved):

PR #99, opened 2025-06-10
test: increase time in clusterresourcebinding watcher IT

zhiying-lin:

Test ClusterResourceBinding Watcher - update status [It] Should enqueue the clusterResourcePlacement name for reconciling, when clusterResourceBinding status changes - RolloutStarted [Serial]
/home/runner/work/kubefleet/kubefleet/pkg/controllers/clusterresourcebindingwatcher/watcher_integration_test.go:123

https://github.com/kubefleet-dev/kubefleet/actions/runs/15528629716/job/43712797542

The key cannot be found within 10s.
Increase the time to reduce the flakiness

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterresourcebindingwatcher/watcher_integration_test.go
+++ b/pkg/controllers/clusterresourcebindingwatcher/watcher_integration_test.go
@@ -37,7 +37,7 @@ const (
 	testReason1                      = "testReason1"
 	testReason2                      = "testReason2"
 
-	eventuallyTimeout    = time.Second * 10
+	eventuallyTimeout    = time.Second * 20
 	consistentlyDuration = time.Second * 10
 	interval             = time.Millisecond * 250
 )
@@ -653,7 +653,14 @@ func validateWhenUpdateClusterResourceBindingStatusWithCondition(conditionType f
 
 	By("Checking placement controller queue")
 	eventuallyCheckPlacementControllerQueue(crb.GetLabels()[fleetv1beta1.CRPTrackingLabel])
-	fakePlacementController.ResetQueue()
+	// Reset the queue to avoid the multiple events triggered.
+	Consistently(func() error {
+		if fakePlacementController.Key() == testCRPName {
+			By("By finding the key and resetting the placement queue")
+			fakePlacementController.ResetQueue()
+		}
+		return nil
+	}, consistentlyDuration, interval).Should(Succeed(), "placementController queue should be stable empty after resetting")
 }
 
 func eventuallyCheckPlacementControllerQueue(key string) {


pkg/controllers/clusterresourcebindingwatcher/watcher_integration_test.go
```diff
@@ -653,7 +653,14 @@ func validateWhenUpdateClusterResourceBindingStatusWithCondition(conditionType f
 
 	By("Checking placement controller queue")
 	eventuallyCheckPlacementControllerQueue(crb.GetLabels()[fleetv1beta1.CRPTrackingLabel])
-	fakePlacementController.ResetQueue()
+	// Reset the queue to avoid the multiple events triggered.
+	Consistently(func() error {
```
jwtty: Will this function ever return an error?

pkg/controllers/clusterresourcebindingwatcher/watcher_integration_test.go
```diff
@@ -37,7 +37,7 @@ const (
 	testReason1                      = "testReason1"
 	testReason2                      = "testReason2"
 
-	eventuallyTimeout    = time.Second * 10
+	eventuallyTimeout    = time.Second * 20
```
jwtty, line 43: The failed test seems to be just updating the binding status and expecting the crp to be requeue'ed. Why can it delay for more than 10s?

pkg/controllers/clusterresourcebindingwatcher/watcher_integration_test.go
zhiying-lin: it won't, in another crp watcher IT, we find a flakiness and suspect it was caused by multiple events triggered for CREATE.

The consistently statement is to make sure the queue is stable enough. It should not be related to this flakiness.

pkg/controllers/clusterresourcebindingwatcher/watcher_integration_test.go
zhiying-lin: i also don't know why and cannot reproduce in my local. Not sure if it's related to the informer cache mechanism. Just want to see if increasing the timeout helps or not.

zhiying-lin: Asked copilot to analyze this issue and it sounds reasonable,

```
1. The Problem: The test is calling
  eventuallyCheckPlacementControllerQueue
  with a 10-second timeout expecting the
  fake controller to have received a
  specific key, but there's a race
  condition between when the status update
   happens and when the controller
  processes the event.
  2. Race Condition Details:
    - Line 741: Status update is performed
   on the ResourceBinding
    - Line 744: The test immediately calls
   eventuallyCheckPlacementControllerQueue

    - The controller's reconcile function
  (which calls Enqueue) might not have
  been triggered yet by the time the test
  checks
  3. Why it's flaky: Controller-runtime's
  event processing is asynchronous. The
  status update triggers an event, but
  there's no guarantee when the controller
   will process it.
```
increasing the timeout should help

jwtty: Looking at the other test failure: https://github.com/kubefleet-dev/kubefleet/actions/runs/16924523797/job/47957596584?pr=182 In both cases, it was the first test in the suite that was executed right after the controller manager began the run and started the workers that failed. So probably the controller did not get the event in time. And for the 2 failed cases, one failed for ClusterResourceBinding enqueue and the other failed for ResourceBinding enqueue so it should not be an issue with our newly added ResourceBinding. Agree to increase the test timeout and monitor.

jwtty (approved):

pkg/controllers/clusterresourcebindingwatcher/watcher_integration_test.go
ryanzhang-oss: I still don't get it. The controller-runtime relies on the informer, how can the informer not detect the change on a pretty much empty cluster for over 10 seconds?  we can try but I am still not sure why.

ryanzhang-oss (approved):

jwtty (approved):

PR #113, opened 2025-06-23
feat: allow backoff with a rate limiter in the work applier

michaelawyu:

This PR sets up the work applier to use a rate limiter to enable backoffs when processing the same work object (of the same generation and the same processing result) over and over again.

I have:

- [x] Unit tests


pkg/controllers/workapplier/controller.go
```diff
@@ -210,47 +208,57 @@ type Reconciler struct {
...
-		acRequestAfter = minRequestAfterDuration
-	}
-
-	dcRequestAfter := driftCheckRequestAfter
-	if dcRequestAfter < minRequestAfterDuration {
-		klog.V(2).InfoS("Drift check requeue after duration is too short; set to the longer default", "driftCheckRequestAfter", dcRequestAfter)
-		dcRequestAfter = minRequestAfterDuration
+	if requeueRateLimiter == nil {
+		klog.V(2).InfoS("No requeue rate limiter provided; using default rate limiter")
+		requeueRateLimiter = NewRequeueFastSlowWithExponentialBackoffRateLimiter(
```
ryanzhang-oss: why not create a defaultRateLimiter in the backoff.go instead of spelling it out here?

pkg/controllers/workapplier/backoff.go
```diff
@@ -0,0 +1,222 @@
...
+// NewRequeueFastSlowWithExponentialBackoffRateLimiter creates a RequeueFastSlowWithExponentialBackoffRateLimiter.
+func NewRequeueFastSlowWithExponentialBackoffRateLimiter(
+	maxFastAttempts int,
+	fastDelaySeconds float64,
+	exponentialBase float64,
+	maxDelaySeconds float64,
+) *RequeueFastSlowWithExponentialBackoffRateLimiter {
+	if maxFastAttempts < minMaxFastAttempts {
+		maxFastAttempts = minMaxFastAttempts
+		klog.V(2).InfoS("maxFastAttempts is below the minimum value; set it to the minimum value instead", minMaxFastAttempts)
```
ryanzhang-oss: info or error?

cmd/memberagent/main.go
```diff
@@ -393,8 +391,8 @@ func Start(ctx context.Context, hubCfg, memberConfig *rest.Config, hubOpts, memb
 			5,
 			// Use the default worker count (4) for parallelized manifest processing.
 			parallelizer.DefaultNumOfWorkers,
-			time.Second*time.Duration(*availabilityCheckInterval),
-			time.Second*time.Duration(*driftDetectionInterval),
+			// Use the default requeue rate limiter.
+			nil,
```
ryanzhang-oss: I don't see this being set anywhere, why keep this parameter?

pkg/controllers/workapplier/backoff.go
```diff
@@ -0,0 +1,222 @@
...
+		// The Work object has never been tracked before. No action is needed.
+	case lastTrackedGen != work.Generation:
+		// A new generation of the Work object has been observed.
+		// Reset the requeue counter.
+		r.requeueCounter[namespacedName] = 0
+	case hashErr != nil:
+		// Cannot compute the hash of the processing result. Normally this should not occur.
+		// Reset the requeue counter just in case.
+		r.requeueCounter[namespacedName] = 0
+	case lastTrackedProcessingResHash != curProcessingResHash:
```
ryanzhang-oss: why do we care about the results other than the last one? IIRC, the hash is computed on the result history

pkg/controllers/workapplier/backoff.go
```diff
@@ -0,0 +1,222 @@
...
+		Namespace: work.Namespace,
+		Name:      work.Name,
+	}
+
+	// Check if the Work object has been tracked before; if so, verify if its has a generation change
+	// or the processing result has changed.
+	lastTrackedGen, isTracked := r.lastTrackedGeneration[namespacedName]
+	lastTrackedProcessingResHash := r.lastTrackedProcessingResultHash[namespacedName]
+	curProcessingResHash, hashErr := computeProcessingResultHash(work, bundles)
+	switch {
```
ryanzhang-oss: is this this whole switch just

if isTracked && (lastTrackedGen != work.Generation || hashErr != nil || lastTrackedProcessingResHash != curProcessingResHash) {
		r.requeueCounter[namespacedName] = 0
}

pkg/controllers/workapplier/backoff.go
```diff
@@ -0,0 +1,222 @@
...
+		// Reset the requeue counter.
+		r.requeueCounter[namespacedName] = 0
+	default:
+		// The Work object has been tracked before and its generation and processing result have not changed.
+		// No action is needed.
+	}
+
+	// Update the last tracked generation and processing result for the Work object.
+	r.lastTrackedGeneration[namespacedName] = work.Generation
+	r.lastTrackedProcessingResultHash[namespacedName] = curProcessingResHash
```
ryanzhang-oss, line 211: what if hashErr != nil?

pkg/controllers/workapplier/backoff.go
```diff
@@ -0,0 +1,222 @@
...
+		Namespace: work.Namespace,
+		Name:      work.Name,
+	}
+	delete(r.requeueCounter, namespacedName)
+	delete(r.lastTrackedGeneration, namespacedName)
+	delete(r.lastTrackedProcessingResultHash, namespacedName)
+}
+
+// NumRequeues returns the number of times a Work object has been requeued.
+func (r *RequeueFastSlowWithExponentialBackoffRateLimiter) NumRequeues(work *fleetv1beta1.Work) int {
```
ryanzhang-oss: where is this used?

pkg/controllers/workapplier/controller.go
michaelawyu: Will make the edit.

pkg/controllers/workapplier/backoff.go
michaelawyu: Will make the edit.

cmd/memberagent/main.go
michaelawyu: Hi Ryan! I was thinking that later we might need to expose some of the parameters as command line args; but for now I could drop this argument if you'd like.

pkg/controllers/workapplier/backoff.go
michaelawyu: Hi Ryan! There are cases where, say, some of the manifests fails to apply first, and then in the current iteration becomes applied; in these cases we must compare the current result and the last result, and reset the requeue counter so that there won't be backoffs that delays availability checks.

pkg/controllers/workapplier/backoff.go
michaelawyu: Yeah, I am just using switch to keep the branches a bit clearer.

pkg/controllers/workapplier/backoff.go
michaelawyu: Hi Ryan! When `hashErr` is not nil the requeue counter is reset and the last tracked hash becomes an empty string; in the next iteration if the hash err disappears, there is guaranteed to be a hash mismatch so that the requeue counter restarts; and if the hash err persists, the counter will be reset again as well just to be on the safer side.

michaelawyu: But if you prefer it to be more explicit I could make the edit.

pkg/controllers/workapplier/backoff.go
michaelawyu: Ah, this one is a leftover as originally I was making the rate limiter in conformance with the workqueue rate limiter interface (which has `NumRequeues` as a required method); it is not necessary any more, let me drop this.

cmd/memberagent/main.go
michaelawyu: Dropped the argument for now.

pkg/controllers/workapplier/backoff.go
michaelawyu: Removed.

michaelawyu:
For simplicity reasons, integration/E2E tests will be submitted separately.

pkg/controllers/workapplier/backoff.go
ryanzhang-oss: I am bit confused, instead of just one if/else, the switch makes it look like there are many different conditions. I didn't realize that until I read it carefully.  I don't get why spreading the same action into 3 different switch cases would make it clearer. Normally each swtich case should have different actions. I think a fallthrough might make more sense

pkg/controllers/workapplier/backoff.go
ryanzhang-oss: I don't get it. In this case, if the lastTrackedHash is computed with "last failed to apply" and the currentHash is computed on " the last succees", why is that not enough?

pkg/controllers/workapplier/controller.go
```diff
@@ -196,6 +194,27 @@ func (h *PriorityQueueEventHandler) Generic(ctx context.Context, evt event.Typed
 	h.AddToPriorityQueue(ctx, evt.Object, false)
 }
 
+var defaultRequeueRateLimiter *RequeueFastSlowWithExponentialBackoffRateLimiter = NewRequeueFastSlowWithExponentialBackoffRateLimiter(
+	// Allow fast requeues for 20 attempts (5 seconds each, 100 seconds of fast requeues).
+	20,
```
ryanzhang-oss: why 20?

pkg/controllers/workapplier/controller.go
```diff
@@ -196,6 +194,27 @@ func (h *PriorityQueueEventHandler) Generic(ctx context.Context, evt event.Typed
...
+	// applier uses static requeue intervals, specifically 5 seconds (if the work object is unavailable),
+	// and 15 seconds (if the work object is available). There are a number of test cases that
+	// implicitly assume this behavior (e.g., a test case might expect that the availability check completes
+	// w/in 10 seconds), which is why the rate limiter uses the 5 seconds fast requeue delay by default.
+	// If you need to change this value and see that many test cases begin to fail, update the test
+	// cases accordingly.
+	5,
+	// Then switch to exponential backoff with a base of 1.2 with a cap of 5 minutes.
+	1.2,
+	300,
```
ryanzhang-oss: this is probably still a bit too aggressive

cmd/memberagent/main.go
```diff
@@ -87,8 +87,6 @@ var (
 	propertyProvider             = flag.String("property-provider", "none", "The property provider to use for the agent.")
 	region                       = flag.String("region", "", "The region where the member cluster resides.")
 	cloudConfigFile              = flag.String("cloud-config", "/etc/kubernetes/provider/config.json", "The path to the cloud cloudconfig file.")
-	availabilityCheckInterval    = flag.Int("availability-check-interval", 5, "The interval in seconds between attempts to check for resource availability when resources are not yet available.")
-	driftDetectionInterval       = flag.Int("drift-detection-interval", 15, "The interval in seconds between attempts to detect configuration drifts in the cluster.")
 	watchWorkWithPriorityQueue   = flag.Bool("enable-watch-work-with-priority-queue", false, "If set, the apply_work controller will watch/reconcile work objects that are created new or have recent updates")
```
ryanzhang-oss, line 90: given the arbitary nature of the backoff numbers, I think we need an easier way to configure it when we do perf testing.

pkg/controllers/workapplier/controller.go
```diff
@@ -210,47 +229,35 @@ type Reconciler struct {
...
 	parallelizer                 *parallelizer.Parallerlizer
 
-	availabilityCheckRequeueAfter time.Duration
-	driftCheckRequeueAfter        time.Duration
+	requeueRateLimiter *RequeueFastSlowWithExponentialBackoffRateLimiter
 }
 
+// NewReconciler returns a new Work object reconciler for the work applier.
+//
+// TO-DO (chenyu1): evaluate if KubeFleet needs to expose the requeue rate limiter
```
ryanzhang-oss, line 236: yes, next PR.

pkg/controllers/workapplier/controller.go
```diff
@@ -475,14 +482,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-	// If the Work object is not yet available, reconcile again.
-	if !isWorkObjectAvailable(work) {
-		klog.V(2).InfoS("Work object is not yet in an available state; requeue to monitor its availability", "work", workRef)
-		return ctrl.Result{RequeueAfter: r.availabilityCheckRequeueAfter}, nil
-	}
-	// Otherwise, reconcile again for drift detection purposes.
-	klog.V(2).InfoS("Work object is available; requeue to check for drifts", "work", workRef)
-	return ctrl.Result{RequeueAfter: r.driftCheckRequeueAfter}, nil
+	// Requeue the Work object with a delay based on the requeue rate limiter.
+	requeueDelay := r.requeueRateLimiter.When(work, bundles)
```
ryanzhang-oss, line 492: actually, I think the type of backoff should be different for avaiable vs not available work. For available work, we should immediately back off aggressively but for not avaialble yet ones, we should check more often. btw, what about untrackable resources?

pkg/controllers/workapplier/controller.go
michaelawyu: N/A -> per earlier discussion, the params have been exposed in this PR.

pkg/controllers/workapplier/controller.go
michaelawyu: per earlier discussion, has implemented the mechanism to skip to fast backoff when the work object has become available (or has diff reported).

cmd/memberagent/main.go
michaelawyu: per earlier discussion, the params have been exposed as CLI arguments.

pkg/controllers/workapplier/controller.go
michaelawyu: per earlier discussion, the default max. delay has been set to 900 secs (15 minutes). The fast backoff factor is now 1.5 instead of 1.2.

pkg/controllers/workapplier/controller.go
michaelawyu: Discussed offline. The logic has changed -> by default the rate limiter will now do 2 requeues with fixed delay (5 secs * 2), then start the backoff stages.

pkg/controllers/workapplier/backoff.go
michaelawyu: Discussed offline -> the style has been updated.

pkg/controllers/workapplier/backoff.go
michaelawyu: Discussed offline -> it is tracking the last result, not the full history.

pkg/controllers/internalmembercluster/v1beta1/member_suite_test.go
```diff
@@ -402,7 +402,7 @@ var _ = BeforeSuite(func() {
 
 	// This controller is created for testing purposes only; no reconciliation loop is actually
 	// run.
-	workApplier2 = workapplier.NewReconciler(hubClient, member2ReservedNSName, nil, nil, nil, nil, 0, 1, time.Second*5, time.Second*5, time.Minute, true, 60)
+	workApplier2 = workapplier.NewReconciler(hubClient, member2ReservedNSName, nil, nil, nil, nil, 0, 1, time.Minute, true, 60, nil)
```
ryanzhang-oss: Why setting the 	watchWorkWithPriorityQueue as true?

pkg/controllers/workapplier/backoff.go
```diff
@@ -0,0 +1,301 @@
...
+const (
+	processingResultStrTpl = "%s,%s,%s"
+)
+
+// RequeueMultiStageWithExponentialBackoffRateLimiter is a rate limiter that allows requeues of various
+// delays in multiple stages; it helps to control the rate of requeues for Work objects while keeping
+// the system as responsive as possible.
+//
+// It will:
+//   - first, requeue with a fixed delay for a limited number of attempts;
```
ryanzhang-oss, line 71: I wonder why we need this part? can't we start with 5 seconds and then directly go to slowe exp back off? I am not sure what the second 5 seconds buy us. It will slow down the availability time detection a bit comparing to the initial 2 second gap

ryanzhang-oss: maybe set the default to 1?

pkg/controllers/internalmembercluster/v1beta1/member_suite_test.go
michaelawyu: Ah, it has been this way in the current master branch; I have patched it out.

pkg/controllers/workapplier/backoff.go
michaelawyu: Default has been set to 1 👌

PR #116, opened 2025-06-26
fix: update crp status if there is a delay in creating the snapshot

zhiying-lin:

keep updating the crp status if crp cannot create the resourceSnapshot immediately.

Fixes #

I have:


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -203,10 +203,32 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 		return ctrl.Result{}, err
 	}
 
-	res, latestResourceSnapshot, err := r.getOrCreateClusterResourceSnapshot(ctx, crp, envelopeObjCount,
+	createResourceSnapshotRes, latestResourceSnapshot, err := r.getOrCreateClusterResourceSnapshot(ctx, crp, envelopeObjCount,
 		&fleetv1beta1.ResourceSnapshotSpec{SelectedResources: selectedResources}, int(revisionLimit))
-	if err != nil || res.Requeue {
-		return res, err
+	if err != nil {
+		// We don't requeue the request here immediately so that placement can keep tracking the rollout status.
```
ryanzhang-oss: what does this comment mean? return err will requeue

pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -1487,7 +1487,8 @@ func TestGetOrCreateClusterResourceSnapshot(t *testing.T) {
 					Spec: fleetv1beta1.ResourceSnapshotSpec{SelectedResources: []fleetv1beta1.ResourceContent{}},
 				},
 			},
-			wantRequeue: true,
+			wantRequeue:             true,
+			wantLatestSnapshotIndex: 0,
```
ryanzhang-oss, line 1493: which test case is for the delayed resourceSnapshot?

test/e2e/placement_with_custom_config_test.go
```diff
@@ -141,3 +141,123 @@ var _ = Describe("validating CRP when using customized resourceSnapshotCreationI
 		Eventually(finalizerRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove controller finalizers from CRP %s", crpName)
 	})
 })
+
+var _ = Describe("validating that CRP status can be updated after updating the resources when using customized resourceSnapshotCreationInterval", Label("custom"), Ordered, func() {
```
ryanzhang-oss: where is the RESOURCE_SNAPSHOT_CREATION_INTERVAL var set in e2e? I don't seem to find it.

charts/hub-agent/README.md
```diff
@@ -19,25 +19,25 @@ _See [helm install](https://helm.sh/docs/helm/helm_install/) for command documen
...
+| `MaxConcurrentClusterPlacement`    | Max concurrent ClusterResourcePlacement operations                                                                  | `100`                                            |
+| `ConcurrentResourceChangeSyncs`    | Max concurrent resourceChange reconcilers                                                                           | `20`                                             |
+| `logFileMaxSize`                  | Max log file size before rotation                                                                                   | `1000000`                                        |
+| `MaxFleetSizeSupported`            | Max number of member clusters supported                                                                             | `100`                                            |
+| `resourceSnapshotCreationInterval` | Interval for resource snapshot creation                                                                             | `1m`                                             |
```
ryanzhang-oss: the default 1m means cx would wait 30 sec, plus other delays in the pipeline, on average to see their change get propagated to the first cluster, seems a bit too long. How about 30 sec?

pkg/controllers/clusterresourceplacement/controller_test.go
zhiying-lin: There are two test cases:
line 1344, resource has changed too fast and there is an active snapshot with multiple revisionLimit
line 2059, resource has changed too fast, selected resources cross clusterResourceSnapshot limit, revision limit is 1

test/e2e/placement_with_custom_config_test.go
zhiying-lin: https://github.com/kubefleet-dev/kubefleet/blob/main/.github/workflows/ci.yml#L98 & https://github.com/kubefleet-dev/kubefleet/blob/main/.github/workflows/ci.yml#L146

charts/hub-agent/README.md
zhiying-lin: There will be few other factors which impact the resource propagation time, eg, the number of resources and their payload. But i'm ok to start with 30s.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: fixed the comment

charts/hub-agent/README.md
zhiying-lin: In the test, keep using 1m to reduce the test flakiness.

charts/hub-agent/values.yaml
```diff
@@ -18,7 +18,7 @@ enableGuardRail: true
 webhookClientConnectionType: service
 forceDeleteWaitTime: 15m0s
 clusterUnhealthyThreshold: 3m0s
-resourceSnapshotCreationInterval: 1m0s
+resourceSnapshotCreationInterval: 1m
```
zhiying-lin: @ryanzhang-oss changed back to 1m. 30s to detect new changes and 30s to bundle all the changes together into one snapshot.

To keep it simple, i used one argument. If we want to provide more flexibility, we can split it into two arguments: one is to control when to detect the new changes, the other one is to control the snapshot creation delay.

apis/placement/v1/resourcesnapshot_types.go
```diff
@@ -51,6 +51,9 @@ import (
 //
 // Each snapshot (excluding the first snapshot) MUST have the following annotations:
 //   - `SubindexOfResourceSnapshotAnnotation` to store the subindex of resource snapshot in the group.
+//
+// Snapshot may have the following annotations to indicate the time of next resourceSnapshot candidate detected by the controller:
```
zhiying-lin, line 57: decided to put the annotation on the snapshot itself instead of CRP.

The benefit is
1. no race condition, compared to CRP. CRP update may fail since user and crp controller can update the resource at the same time, while resourceSnapshot only has one owner.
2. later we may support override snapshot creation delay. It's also better to put this info on the override snapshot, other than CRP.

pkg/utils/annotations/annotations.go
```diff
@@ -98,3 +99,17 @@ func ExtractNumberOfEnvelopeObjFromResourceSnapshot(snapshot *fleetv1beta1.Clust
 	}
 	return envelopeObjCount, nil
 }
+
+// ExtractNextResourceSnapshotCandidateDetectionTimeFromResourceSnapshot extracts the next resource snapshot candidate detection time from the annotations of a clusterResourceSnapshot.
+// If the annotation does not exist, it returns 0 duration.
+func ExtractNextResourceSnapshotCandidateDetectionTimeFromResourceSnapshot(snapshot *fleetv1beta1.ClusterResourceSnapshot) (time.Time, error) {
```
ryanzhang-oss: please use the new interface ResourceSnapshotObj instead (and this function can even just take client.object)

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -547,6 +588,52 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
...
+	// And then the next 30s will be used to capture all the resource changes into one snapshot.
+	snapshotKObj := klog.KObj(latestResourceSnapshot)
+	now := time.Now()
+	half := r.ResourceSnapshotCreationInterval / 2
+	if since := now.Sub(latestResourceSnapshot.CreationTimestamp.Time); since < half {
+		// If the latest resource snapshot is created less than configured the resourceSnapshotCreationInterval,
+		//  requeue the request to avoid too frequent update.
+		klog.V(2).InfoS("The latest resource snapshot is just created, skipping the update",
+			"clusterResourceSnapshot", snapshotKObj, "creationTime", latestResourceSnapshot.CreationTimestamp, "configuredResourceSnapshotCreationInterval", r.ResourceSnapshotCreationInterval, "afterDuration", half-since)
+		return ctrl.Result{Requeue: true, RequeueAfter: half - since}, nil
```
ryanzhang-oss: this says we will retry at time "half" but we already know we won't create another snapshot since we will wait for another half, why retry it there? We already know that we will add an annotation NextResourceSnapshotCandidateDetectionTimeFromResourceSnapshot = time "half" at that time, right? 

Why not just requeue at time "half" + half

charts/hub-agent/values.yaml
ryanzhang-oss: yeah, two arguments would make a lot more sense, these are two separate settings conceptually as you described.

ryanzhang-oss: the algorithm could also be simplified to just the below high-level logic

```
if newChangeDetected {

    if NextResourceSnapshotCandidateDetectionTime == 0 {
          NextResourceSnapshotCandidateDetectionTime= now
    }

     nextSnapshotCreationTime = max ( lastSnapshot.creationTime + snapshot creation interval,   NextResourceSnapshotCandidateDetectionTime + resource change collection time) 

     if now < nextSnapshotCreationTime  {
          return  requeue to nextSnapshotCreationTime
     }
     createNewSnapshot()
}

```

apis/placement/v1/resourcesnapshot_types.go
ryanzhang-oss: what about policySnapshot?

charts/hub-agent/values.yaml
zhiying-lin: it's different. 

For example, we detect the first change at 61s, and snapshot creation interval is 60s,
we should not create the snapshot immediately, and need to wait until 61 + 30 to create the snapshot since we want to bundle the next 30s changes into one snapshot. 

That's why i requeue twice.

charts/hub-agent/values.yaml
```diff
@@ -18,7 +18,8 @@ enableGuardRail: true
 webhookClientConnectionType: service
 forceDeleteWaitTime: 15m0s
 clusterUnhealthyThreshold: 3m0s
-resourceSnapshotCreationInterval: 1m0s
+resourceSnapshotCreationInterval: 30s
+resourceChangesCollectionDuration: 30s
```
ryanzhang-oss: I feel that these two don't need to be the same
```suggestion
resourceSnapshotCreationInterval: 1m0s
resourceChangesCollectionDuration: 15s
```

cmd/hubagent/options/options.go
```diff
@@ -104,8 +104,11 @@ type Options struct {
 	PprofPort int
 	// DenyModifyMemberClusterLabels indicates if the member cluster labels cannot be modified by groups (excluding system:masters)
 	DenyModifyMemberClusterLabels bool
-	// ResourceSnapshotCreationInterval is the interval at which resource snapshots are created.
+	// ResourceSnapshotCreationInterval is the interval at which resource snapshots could be created.
```
ryanzhang-oss:
```suggestion
	// ResourceSnapshotCreationInterval is the minimum interval at which resource snapshots could be created.
```

cmd/hubagent/options/options.go
```diff
@@ -104,8 +104,11 @@ type Options struct {
...
+	// Whether the resource snapshot is created or not depends on the both ResourceSnapshotCreationInterval and resourceChangesCollectionDuration.
 	ResourceSnapshotCreationInterval time.Duration
```
ryanzhang-oss:
```suggestion
	ResourceSnapshotCreationMinimumInterval time. Duration
```

ryanzhang-oss: interval sounds like we are going to create a resourcesnapshot for every interval

pkg/utils/controller/controller.go
```diff
@@ -406,9 +406,22 @@ func CollectResourceIdentifiersFromClusterResourceSnapshot(
...
+	ctx context.Context,
+	k8Client client.Client,
+	crpName string,
+	masterResourceSnapshot *fleetv1beta1.ClusterResourceSnapshot,
+	resourceSnapshotIndex string,
+) ([]fleetv1beta1.ResourceIdentifier, error) {
 	allResourceSnapshots, err := FetchAllClusterResourceSnapshots(ctx, k8Client, crpName, masterResourceSnapshot)
 	if err != nil {
-		klog.ErrorS(err, "Failed to fetch all the clusterResourceSnapshots", "resourceSnapshotIndex", resourceSnapshotIndex, "clusterResourcePlacement", crpName)
+		klog.ErrorS(err, "Failed to fetch all the clusterResourceSnapshots", "clusterResourcePlacement", crpName)
```
ryanzhang-oss: why removing the resourceSnapshotIndex?

charts/hub-agent/values.yaml
zhiying-lin: changed to 30s & 15s. Tested our helm install hub agent(< 3s) and member agent (< 2s). 15s should be good enough.

ryanzhang-oss (approved):

PR #117, opened 2025-06-26
fix: address an issue where agent might panic if CSA cannot be used and apply strategy is not fully configured

michaelawyu:

This PR fixes an issue where the work applier might panic when

a) the user uses client side apply to apply an oversized object (>262144 bytes)
b) the user does not have the apply strategy fully configured

I have:

- [] Unit test -> N/A
- [x] Integration tests
- [x] E2E tests 

N/A


test/e2e/placement_apply_strategy_test.go
```diff
@@ -739,6 +741,195 @@ var _ = Describe("SSA", Ordered, func() {
 			ensureCRPAndRelatedResourcesDeleted(crpName, allMemberClusters)
 		})
 	})
+
+	Context("fall back to server-side apply when client-side apply cannot be used", func() {
```
ryanzhang-oss, line 727: I thought we had a test for this before

pkg/utils/defaulter/clusterresourceplacement.go
```diff
@@ -118,4 +112,11 @@ func SetDefaultsApplyStrategy(obj *fleetv1beta1.ApplyStrategy) {
 	if obj.WhenToTakeOver == "" {
 		obj.WhenToTakeOver = fleetv1beta1.WhenToTakeOverTypeAlways
 	}
+	// Always set the ServerSideApplyConfig when it is unset even if
+	// server-side apply itself might not be used, just for completeness reasons.
+	if obj.ServerSideApplyConfig == nil {
+		obj.ServerSideApplyConfig = &fleetv1beta1.ServerSideApplyConfig{
+			ForceConflicts: false,
+		}
+	}
```
ryanzhang-oss: Will user see this? I am not sure if user should see their applyStrategy's serversideApply Config suddenly show up

pkg/controllers/workapplier/controller_integration_test.go
```diff
@@ -5379,4 +5381,227 @@ var _ = Describe("switch apply strategies", func() {
 			// deletion; consequently this test suite would not attempt so verify its deletion.
 		})
 	})
+
+	FContext("falling back from CSA to SSA", Ordered, func() {
```
ryanzhang-oss:
```suggestion
	Context("falling back from CSA to SSA", Ordered, func() {
```

test/e2e/placement_apply_strategy_test.go
michaelawyu: Sync'd offline.

We do have a test case for this, unfortunately due to a complication on how that test case is implemented, the over-sized object is not actually applied (only created).

pkg/utils/defaulter/clusterresourceplacement.go
michaelawyu: Sync'd offline -> the user will not see this on the CRP API; the work object, however, will feature apply strategies that have been processed by the defaulter.

pkg/controllers/workapplier/controller_integration_test.go
michaelawyu: Edited.

PR #123, opened 2025-07-01
Update Bug repot templates

ryanzhang-oss:
add a note to the template to make it clear that this repo is separate from the Azure Kubernetes Fleet Manager.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,6 +7,10 @@ assignees: ''
 
 ---
 
+<!--
+Please open bugs in https://github.com/Azure/fleet if the request is specifically related to Azure Kubernetes Fleet Manager 
+-->
+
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,6 +7,10 @@ assignees: ''
 
 ---
 
+<!--
+Please open bugs in https://github.com/Azure/fleet if the request is specifically related to Azure Kubernetes Fleet Manager 
+-->
+
```
Arvindthiru, line 13: This change doesn't show up on preview, is this expected ?

image

britaniar: Yeah, I assume we want this present on the page actual page?

michaelawyu: Ah, I am guessing it's being recognized as an HTML comment instead?

PR #126, opened 2025-07-07
feat: convert some override webhook logic to CEL & add tests

britaniar:

Fixes #

I have:


apis/placement/v1alpha1/override_types.go
```diff
@@ -198,6 +200,9 @@ type JSONPatchOverride struct {
 	// Path defines the target location.
 	// Note: override will fail if the resource path does not exist.
 	// +required
+	// +kubebuilder:validation:XValidation:rule="!(self == '/kind' || self == '/apiVersion')",message="Path cannot target typeMeta fields (kind, apiVersion)"
+	// +kubebuilder:validation:XValidation:rule="!(self.startsWith('/metadata') && !(self.startsWith('/metadata/annotations') || self.startsWith('/metadata/labels')))",message="Path can only target annotations and labels under metadata"
+	// +kubebuilder:validation:XValidation:rule="!self.startsWith('/status')",message="Path cannot target status fields"
```
britaniar, line 205: Not sure if we want to keep these validations since we cannot cover all the scenarios due to the budget.

britaniar: The budget is currently calculated using the worst case scenario. It is [recommended to add a max length](https://github.com/kubernetes/enhancements/blob/bdb679de388fb1947e5514fe90d997bad797b4ab/keps/sig-api-machinery/2876-crd-validation-expression-language/README.md?plain=1#L951) to help with calculation costs, but I don't think we want to add a max length to this.

zhiying-lin: if we cannot remove the webhook validation rules using the CEL for this object or CEL rules are too complicated (reach the budget), just keep the validations in the webhook.

PR #127, opened 2025-07-08
feat: refactor work generator to use interface objects

ryanzhang-oss:

refactor work generator to use interface objects

Fixes #

I have:


pkg/utils/controller/binding_resolver.go
```diff
@@ -20,12 +20,45 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	// Check if the key contains a namespace separator
+	if namespace != "" {
+		// This is a namespaced ResourceBinding
+		rb := &placementv1beta1.ResourceBinding{}
+		key := types.NamespacedName{
+			Namespace: namespace,
+			Name:      name,
```
zhiying-lin, line 43: nit, it can be created as a common variable.

pkg/controllers/workgenerator/controller.go
```diff
@@ -86,89 +86,92 @@ type Reconciler struct {
...
+	// Get the binding using the utility function
+	placementKey := controller.GetObjectKeyFromRequest(req)
+	resourceBinding, err := controller.FetchBindingFromKey(ctx, r.Client, placementKey)
+	if err != nil {
 		if apierrors.IsNotFound(err) {
 			return controllerruntime.Result{}, nil
 		}
-		klog.ErrorS(err, "Failed to get the resource binding", "resourceBinding", bindingRef)
+		klog.ErrorS(err, "Failed to get the resource binding", "binding", bindingRef)
 		return controllerruntime.Result{}, controller.NewAPIServerError(true, err)
```
zhiying-lin, line 105: it is no longer controller.NewAPIServerError and it could be unexpected error.

pkg/controllers/workgenerator/controller.go
```diff
@@ -269,52 +272,55 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
 	err := r.Client.Status().Update(ctx, resourceBinding)
 	if err != nil {
-		klog.ErrorS(err, "Failed to update the resourceBinding status, will retry", "resourceBinding", bindingRef, "resourceBindingStatus", resourceBinding.Status)
+		klog.ErrorS(err, "Failed to update the binding status, will retry", "binding", bindingRef, "bindingStatus", resourceBinding.GetBindingStatus())
 		errAfterRetries := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-			var latestBinding fleetv1beta1.ClusterResourceBinding
-			if err := r.Client.Get(ctx, client.ObjectKeyFromObject(resourceBinding), &latestBinding); err != nil {
+			// Get the latest binding object using the utility function
+			placementKey := controller.GetObjectKeyFromObj(resourceBinding)
+			latestBinding, err := controller.FetchBindingFromKey(ctx, r.Client, placementKey)
```
zhiying-lin, line 284: ditto

pkg/controllers/workgenerator/controller.go
```diff
@@ -86,89 +86,92 @@ type Reconciler struct {
...
 	// add latency log
 	defer func() {
-		klog.V(2).InfoS("ClusterResourceBinding reconciliation loop ends", "resourceBinding", bindingRef, "latency", time.Since(startTime).Milliseconds())
+		klog.V(2).InfoS("Binding reconciliation loop ends", "binding", bindingRef, "latency", time.Since(startTime).Milliseconds())
 	}()
-	var resourceBinding fleetv1beta1.ClusterResourceBinding
-	if err := r.Client.Get(ctx, req.NamespacedName, &resourceBinding); err != nil {
+
+	// Get the binding using the utility function
+	placementKey := controller.GetObjectKeyFromRequest(req)
```
zhiying-lin, line 98: how about creating a func to getting the namespacedName from the obj directly?

instead of encoding/decoding the objectKey.

pkg/utils/controller/controller.go
```diff
@@ -322,8 +322,8 @@ var (
 	errResourceNotFullyCreated = errors.New("not all resource snapshot in the same index group are created")
 )
 
-// FetchAllClusterResourceSnapshots fetches the group of clusterResourceSnapshots using master clusterResourceSnapshot.
-func FetchAllClusterResourceSnapshots(ctx context.Context, k8Client client.Reader, placementKey string, masterResourceSnapshot fleetv1beta1.ResourceSnapshotObj) (map[string]fleetv1beta1.ResourceSnapshotObj, error) {
+// FetchAllResourceSnapshots fetches the group of clusterResourceSnapshots or resourceSnapshots using the latest master resourceSnapshot.
+func FetchAllResourceSnapshots(ctx context.Context, k8Client client.Reader, placementKey string, masterResourceSnapshot fleetv1beta1.ResourceSnapshotObj) (map[string]fleetv1beta1.ResourceSnapshotObj, error) {
```
zhiying-lin, line 326:
```suggestion
func FetchAllResourceSnapshots(ctx context.Context, k8Client client.Reader, placementKey types.NamespacedName, masterResourceSnapshot fleetv1beta1.ResourceSnapshotObj) (map[string]fleetv1beta1.ResourceSnapshotObj, error) {
```

zhiying-lin (commented):
finished reviewing all the code, not all the tests since some of the code can be simplified.

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: we don't have an object yet here. I don't get what you mean.

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: FetchBindingFromKey func just returns the client err, what has changed?

pkg/utils/controller/controller.go
ryanzhang-oss: there are a few cases that the caller are not coming from the request directly. for example, the name comes from a label or field (placeRef, CRPTracker, etc) and the namespace comes from an object. We need to add a converstion function too. There is no single common original format

zhiying-lin (approved):
Discussed offline, some improvements could be addressed in a separate PR, approve to unblock the progress.

PR #128, opened 2025-07-08
chore: Delete pkg/controllers/work Directory thoroughly

NLaureano:

The intention is to gracefully remove the directory "~/pkg/controller/work" which is no longer needed, catching any lingering mentions of the dir.

I have:

- [X] `Make test` has been run
- [ ] `Make test` runs without failures locally

CURRENT LOCAL TEST FAILURES:


test/upgrade/before/setup_test.go
```diff
@@ -136,8 +134,8 @@ var (
 	// Note that the aforementioned change is hub side exclusive and is for informational purposes only.
 	availableDueToUntrackableResCondAcyclicTransformer = cmpopts.AcyclicTransformer("AvailableDueToUntrackableResCond", func(cond metav1.Condition) metav1.Condition {
 		transformedCond := cond.DeepCopy()
-		if cond.Type == string(placementv1beta1.ResourcesAvailableConditionType) && cond.Reason == work.WorkNotTrackableReason {
-			transformedCond.Reason = condition.WorkNotAvailabilityTrackableReason
+		if cond.Type == string(placementv1beta1.ResourcesAvailableConditionType) && cond.Reason == "NotTrackable" {
```
britaniar: Use the constants instead of the string
```suggestion
		if cond.Type == string(placementv1beta1.ResourcesAvailableConditionType) && cond.Reason == WorkNotAllManifestsTrackableReason {
```

test/upgrade/before/setup_test.go
```diff
@@ -136,8 +134,8 @@ var (
...
+			transformedCond.Reason = "NotTrackable"
```
britaniar: This should not be changed. `condition.WorkNotAvailabilityTrackableReason` is different from "NotTrackable".
https://github.com/kubefleet-dev/kubefleet/blob/50907ffaa545ab2b6e33721c7d17c272ad336586/pkg/utils/condition/condition.go#L132

britaniar (approved):

PR #146, opened 2025-07-23
interface: Add v1 Override APIs, CRDs

Arvindthiru:

Fixes #

I have:


apis/placement/v1/commons.go
```diff
@@ -16,6 +16,13 @@ limitations under the License.
 
 package v1
 
+const (
+	// fleetPrefix is the prefix used for official fleet labels/annotations.
+	// Unprefixed labels/annotations are reserved for end-users
+	// See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions
+	fleetPrefix = "kubernetes-fleet.io/"
```
ryanzhang-oss, line 23: why duplicate this?

Arvindthiru: If we need this as a private var inside package v1 I don't see other options. If we change it to a public var we need a new directory under apis/common/commons.go which would host all resuable vars across our different apis v1alpha1, v1beta1 & v1

ryanzhang-oss: I think we can do that later

PR #149, opened 2025-07-25
feat: refactor clusterresourceplacementwatcher to support rp

zhiying-lin:

refactor clusterresourceplacementwatcher to support rp

Fixes #

I have:

- [X] Run `make reviewable` to ensure this PR is ready for review.

### DIFF
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -198,7 +198,7 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		klog.Info("Setting up clusterResourcePlacement watcher")
 		if err := (&clusterresourceplacementwatcher.Reconciler{
 			PlacementController: clusterResourcePlacementControllerV1Beta1,
-		}).SetupWithManager(mgr); err != nil {
+		}).SetupWithManagerForClusterResourcePlacement(mgr); err != nil {
 			klog.ErrorS(err, "Unable to set up the clusterResourcePlacement watcher")
 			return err
 		}
--- a/pkg/controllers/clusterresourceplacementwatcher/watcher_integration_test.go
+++ b/pkg/controllers/clusterresourceplacementwatcher/watcher_integration_test.go
@@ -26,10 +26,17 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	fleetv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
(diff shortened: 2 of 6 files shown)


cmd/hubagent/workload/setup.go
```diff
@@ -198,7 +198,7 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		klog.Info("Setting up clusterResourcePlacement watcher")
 		if err := (&clusterresourceplacementwatcher.Reconciler{
 			PlacementController: clusterResourcePlacementControllerV1Beta1,
-		}).SetupWithManager(mgr); err != nil {
+		}).SetupWithManagerForClusterResourcePlacement(mgr); err != nil {
```
ryanzhang-oss, line 201: this seems to indicate that we will have two reconcilers? One for CRP and one for RP?

zhiying-lin: Are you asking if we'll have two reconcilers for watcher or placement controllers?

Prefer to use two controllers for watcher, so that the progress won't be blocked by the CRD.

whether to use single controller for placement is an independent question.
There are two options:
1. one CRP watcher, one RP watcher and both enqueue the key to the (C)RP reconciler.
2. CRP and RP have separate reconciler and watcher enqueues the keys separately.

For now, these two options have not much diff, unless they want to share the same view/memory/etc and they have to be a single reconciler.

pkg/controllers/clusterresourceplacementwatcher/watcher_integration_test.go
```diff
@@ -151,3 +174,52 @@ var _ = Describe("Test ClusterResourcePlacement Watcher", Serial, func() {
...
+		// Reset the queue to avoid the multiple create events triggered.
+		Consistently(func() error {
+			if fakePlacementController.Key() == key {
+				By("By finding the key and resetting the placement queue")
+				fakePlacementController.ResetQueue()
+			}
+			return nil
+		}, consistentlyDuration, interval).Should(Succeed(), "placementController queue should be stable empty after resetting")
+	})
+
```
jwtty, line 208: Should we copy all existing UTs for RP?

zhiying-lin: as the setupWithManager func is pretty simple, the functionality should be tested by controller runtime package and we already have enough tests in the CRP. so i only copied partial IT to make sure RP watcher works.

jwtty (approved):

PR #157, opened 2025-07-30
feat: refactor crp watcher for the scheduler

zhiying-lin:

refactor crp scheduler watcher to support RP

Fixes #

I have:

Duplicate the full IT for the RP as there are some dependencies between the tests in order to keep the order and reduce the maintenance cost

### DIFF
--- a/pkg/scheduler/watchers/clusterresourceplacement/watcher.go
+++ b/pkg/scheduler/watchers/clusterresourceplacement/watcher.go
@@ -14,8 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package clusterresourceplacement features a controller that enqueues CRPs for the
-// scheduler to process where the CRP is marked for deletion.
+// Package clusterresourceplacement features a controller that enqueues placement objects for the
+// scheduler to process where the placement object is marked for deletion.
 package clusterresourceplacement
 
 import (
@@ -35,7 +35,7 @@ import (
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/controller"
 )
 
-// Reconciler reconciles the deletion of a CRP.
+// Reconciler reconciles the deletion of a placement object.
 type Reconciler struct {
 	// Client is the client the controller uses to access the hub cluster.
 	client.Client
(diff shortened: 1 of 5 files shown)


pkg/scheduler/watchers/clusterresourceplacement/watcher.go
```diff
@@ -101,9 +101,19 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
+// SetupWithManagerForClusterResourcePlacement sets up the controller with the manager.
+func (r *Reconciler) SetupWithManagerForClusterResourcePlacement(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).Named("clusterresourceplacement-scheduler-watcher").
 		For(&fleetv1beta1.ClusterResourcePlacement{}).
-		WithEventFilter(customPredicate).
+		WithEventFilter(buildCustomPredicate()).
+		Complete(r)
+}
+
```
britaniar, line 113: Add a comment
```suggestion
// SetupWithManagerForResourcePlacement sets up the controller with the manager.

File: pkg/scheduler/watchers/clusterresourceplacement/watcher.go [hunk 2]
```diff
@@ -35,45 +35,45 @@ import (
...
+	if placement.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(placement, fleetv1beta1.SchedulerCleanupFinalizer) {
+		// The placement has been deleted and still has the scheduler finalizer;
 		// enqueue it for the scheduler to process.
-		r.SchedulerWorkQueue.AddRateLimited(queue.PlacementKey(crp.Name))
+		r.SchedulerWorkQueue.AddRateLimited(placementKey)
 	}
 
 	// No action is needed for the scheduler to take in other cases.
 	return ctrl.Result{}, nil
 }
```
britaniar, line 74: Add comment
```suggestion
// buildCustomPredicate creates a predicate that only triggers on deletion timestamp changes.
```

pkg/scheduler/watchers/clusterresourceplacement/watcher.go
```diff
@@ -35,45 +35,45 @@ import (
...
 
-// SetupWithManager sets up the controller with the manager.
-func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
-	customPredicate := predicate.Funcs{
+func buildCustomPredicate() predicate.Predicate {
```
britaniar, line 75: We used the same thing in different controllers. Could we move to a central place since it is duplicate code?

pkg/scheduler/watchers/clusterresourceplacement/watcher.go
zhiying-lin: comment for the internal func is not required.

pkg/scheduler/watchers/clusterresourceplacement/watcher.go
zhiying-lin: Usually what controller watches is different. I feel it could be error prone to make it as a common func.

britaniar (approved):

PR #158, opened 2025-07-31
feat: refactor rollout controller for RP

jwtty:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -135,7 +135,7 @@ func (r *Reconciler) handleUpdate(ctx context.Context, placementObj fleetv1beta1
 
 	if placementSpec.RevisionHistoryLimit != nil {
 		if revisionLimit <= 0 {
-			err := fmt.Errorf("invalid placement %s: invalid revisionHistoryLimit %d", placementKObj, revisionLimit)
+			err := fmt.Errorf("invalid placement %s/%s: invalid revisionHistoryLimit %d", placementObj.GetNamespace(), placementObj.GetName(), revisionLimit)
 			klog.ErrorS(controller.NewUnexpectedBehaviorError(err), "Invalid revisionHistoryLimit value and using default value instead", "placement", placementKObj)
 		} else {
 			revisionLimit = *placementSpec.RevisionHistoryLimit
@@ -1053,10 +1053,9 @@ func (r *Reconciler) determineRolloutStateForPlacementWithExternalRolloutStrateg
 	allRPS []fleetv1beta1.PerClusterPlacementStatus,
 	selectedResourceIDs []fleetv1beta1.ResourceIdentifier,
 ) (bool, error) {
-	placementKObj := klog.KObj(placementObj)
 	if len(selected) == 0 {
 		// This should not happen as we already checked in setPlacementStatus.
-		err := controller.NewUnexpectedBehaviorError(fmt.Errorf("selected cluster list is empty for placement %s when checking per-cluster rollout state", placementKObj))
+		err := controller.NewUnexpectedBehaviorError(fmt.Errorf("selected cluster list is empty for placement %s/%s when checking per-cluster rollout state", placementObj.GetNamespace(), placementObj.GetName()))
 		klog.ErrorS(err, "Should not happen: selected cluster list is empty in determineRolloutStateForPlacementWithExternalRolloutStrategy()")
 		return false, err
 	}
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -135,7 +135,7 @@ func (r *Reconciler) handleUpdate(ctx context.Context, placementObj fleetv1beta1
 
 	if placementSpec.RevisionHistoryLimit != nil {
 		if revisionLimit <= 0 {
-			err := fmt.Errorf("invalid placement %s: invalid revisionHistoryLimit %d", placementObj.GetName(), revisionLimit)
+			err := fmt.Errorf("invalid placement %s: invalid revisionHistoryLimit %d", placementKObj, revisionLimit)
 			klog.ErrorS(controller.NewUnexpectedBehaviorError(err), "Invalid revisionHistoryLimit value and using default value instead", "placement", placementKObj)
 		} else {
 			revisionLimit = *placementSpec.RevisionHistoryLimit
@@ -189,7 +189,8 @@ func (r *Reconciler) handleUpdate(ctx context.Context, placementObj fleetv1beta1
 			klog.ErrorS(err, "Failed to extract the resource index from the resourceSnapshot", "placement", placementKObj, "resourceSnapshot", latestResourceSnapshotKObj)
 			return ctrl.Result{}, controller.NewUnexpectedBehaviorError(err)
 		}
-		selectedResourceIDs, err = controller.CollectResourceIdentifiersUsingMasterResourceSnapshot(ctx, r.Client, placementObj.GetName(), latestResourceSnapshot, strconv.Itoa(latestResourceSnapshotIndex))
+		placementKey := controller.GetObjectKeyFromNamespaceName(placementObj.GetNamespace(), placementObj.GetName())
+		selectedResourceIDs, err = controller.CollectResourceIdentifiersUsingMasterResourceSnapshot(ctx, r.Client, placementKey, latestResourceSnapshot, strconv.Itoa(latestResourceSnapshotIndex))
 		if err != nil {
 			klog.ErrorS(err, "Failed to collect resource identifiers from the resourceSnapshot", "placement", placementKObj, "resourceSnapshot", latestResourceSnapshotKObj)
 			return ctrl.Result{}, err
@@ -459,7 +460,7 @@ func (r *Reconciler) getOrCreateResourceSnapshot(ctx context.Context, placement 
 		// check to see all that the master cluster resource snapshot and sub-indexed snapshots belonging to the same group index exists.
 		resourceSnapshotList, err := controller.ListAllResourceSnapshotWithAnIndex(ctx, r.Client, latestResourceSnapshot.GetLabels()[fleetv1beta1.ResourceIndexLabel], placement.GetName(), placement.GetNamespace())
 		if err != nil {
-			klog.ErrorS(err, "Failed to list the latest group resourceSnapshots associated with the placement", "placement", placement.GetName())
+			klog.ErrorS(err, "Failed to list the latest group resourceSnapshots associated with the placement", "placement", placementKObj)
 			return ctrl.Result{}, nil, controller.NewAPIServerError(true, err)
 		}
 		if len(resourceSnapshotList.GetResourceSnapshotObjs()) == numberOfSnapshots {
@@ -1052,9 +1053,10 @@ func (r *Reconciler) determineRolloutStateForPlacementWithExternalRolloutStrateg
 	allRPS []fleetv1beta1.PerClusterPlacementStatus,
 	selectedResourceIDs []fleetv1beta1.ResourceIdentifier,
 ) (bool, error) {
+	placementKObj := klog.KObj(placementObj)
 	if len(selected) == 0 {
 		// This should not happen as we already checked in setPlacementStatus.
-		err := controller.NewUnexpectedBehaviorError(fmt.Errorf("selected cluster list is empty for placement %s when checking per-cluster rollout state", placementObj.GetName()))
+		err := controller.NewUnexpectedBehaviorError(fmt.Errorf("selected cluster list is empty for placement %s when checking per-cluster rollout state", placementKObj))
 		klog.ErrorS(err, "Should not happen: selected cluster list is empty in determineRolloutStateForPlacementWithExternalRolloutStrategy()")
 		return false, err
 	}
@@ -1120,10 +1122,7 @@ func (r *Reconciler) determineRolloutStateForPlacementWithExternalRolloutStrateg
 	} else {
 		placementStatus.ObservedResourceIndex = observedResourceIndex
 		// Construct placement key for the resource collection function
-		placementKey := placementObj.GetName()
-		if placementObj.GetNamespace() != "" {
-			placementKey = placementObj.GetNamespace() + "/" + placementObj.GetName()
-		}
+		placementKey := controller.GetObjectKeyFromNamespaceName(placementObj.GetNamespace(), placementObj.GetName())
 		selectedResources, err := controller.CollectResourceIdentifiersFromResourceSnapshot(ctx, r.Client, placementKey, observedResourceIndex)
 		if err != nil {
 			klog.ErrorS(err, "Failed to collect resource identifiers from resourceSnapshot", "placement", klog.KObj(placementObj), "resourceSnapshotIndex", observedResourceIndex)
--- a/pkg/controllers/rollout/controller.go
+++ b/pkg/controllers/rollout/controller.go
@@ -262,20 +262,20 @@ func waitForResourcesToCleanUp(allBindings []placementv1beta1.BindingObj, placem
 				bindingMap[bindingSpec.TargetCluster] = binding
 			} else {
 				return false, controller.NewUnexpectedBehaviorError(fmt.Errorf("the same cluster `%s` has bindings `%s` and `%s` pointing to it",
-					bindingSpec.TargetCluster, bindingMap[bindingSpec.TargetCluster].GetName(), binding.GetName()))
+					bindingSpec.TargetCluster, klog.KObj(bindingMap[bindingSpec.TargetCluster]), klog.KObj(binding)))
 			}
 		}
 	}
 	// check if there are any cluster that has a binding that is both being deleted and scheduled
 	for cluster, binding := range bindingMap {
-		// check if there is a  deleting binding on the same cluster
+		// check if there is a deleting binding on the same cluster
 		if deletingBinding[cluster] {
-			klog.V(2).InfoS("Find a binding assigned to a cluster with another deleting binding", "placement", placementObjRef, "binding", binding)
+			klog.V(2).InfoS("Find a binding assigned to a cluster with another deleting binding", "placement", placementObjRef, "binding", klog.KObj(binding))
 			bindingSpec := binding.GetBindingSpec()
 			if bindingSpec.State == placementv1beta1.BindingStateBound {
 				// the rollout controller won't move a binding from scheduled state to bound if there is a deleting binding on the same cluster.
 				return false, controller.NewUnexpectedBehaviorError(fmt.Errorf(
-					"find a cluster `%s` that has a bound binding `%s` and a deleting binding point to it", bindingSpec.TargetCluster, binding.GetName()))
+					"find a cluster `%s` that has a bound binding `%s` and a deleting binding point to it", bindingSpec.TargetCluster, klog.KObj(binding)))
 			}
 			if bindingSpec.State == placementv1beta1.BindingStateUnscheduled {
 				// this is a very rare case that the resource was in the middle of being removed from a member cluster after it is unselected.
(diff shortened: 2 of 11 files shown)


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -135,7 +135,7 @@ func (r *Reconciler) handleUpdate(ctx context.Context, placementObj fleetv1beta1
 
 	if placementSpec.RevisionHistoryLimit != nil {
 		if revisionLimit <= 0 {
-			err := fmt.Errorf("invalid placement %s: invalid revisionHistoryLimit %d", placementObj.GetName(), revisionLimit)
+			err := fmt.Errorf("invalid placement %s: invalid revisionHistoryLimit %d", placementKObj, revisionLimit)
```
ryanzhang-oss: what will be the output for placementKObj  as a string? I don't get why we change every where in this  file to print out obj with %s

zhiying-lin: would be "namespace/name". only recording name is not enough to distinguish whether it's a RP or CRP.

pkg/controllers/rollout/controller.go
```diff
@@ -262,28 +262,28 @@ func waitForResourcesToCleanUp(allBindings []placementv1beta1.BindingObj, placem
 				bindingMap[bindingSpec.TargetCluster] = binding
 			} else {
 				return false, controller.NewUnexpectedBehaviorError(fmt.Errorf("the same cluster `%s` has bindings `%s` and `%s` pointing to it",
-					bindingSpec.TargetCluster, bindingMap[bindingSpec.TargetCluster].GetName(), binding.GetName()))
+					bindingSpec.TargetCluster, klog.KObj(bindingMap[bindingSpec.TargetCluster]), klog.KObj(binding)))
```
zhiying-lin: prefer not to use the func from the log package in the error string. I usually only use the klog utils for log only.

pkg/controllers/rollout/controller.go
```diff
@@ -262,28 +262,28 @@ func waitForResourcesToCleanUp(allBindings []placementv1beta1.BindingObj, placem
...
 			}
 		}
 	}
 	// check if there are any cluster that has a binding that is both being deleted and scheduled
 	for cluster, binding := range bindingMap {
-		// check if there is a  deleting binding on the same cluster
+		// check if there is a deleting binding on the same cluster
 		if deletingBinding[cluster] {
-			klog.V(2).InfoS("Find a binding assigned to a cluster with another deleting binding", "placement", placementObjRef, "binding", binding)
+			klog.V(2).InfoS("Find a binding assigned to a cluster with another deleting binding", "placement", placementObjRef, "binding", klog.KObj(binding))
```
zhiying-lin, line 275: create a bindingKOjb variable so it can be reused in multiple places?

pkg/controllers/rollout/controller.go
```diff
@@ -678,12 +678,12 @@ func (r *Reconciler) updateBindings(ctx context.Context, bindings []toBeUpdatedB
...
 
-// SetupWithManager sets up the rollout controller with the Manager.
+// SetupWithManagerForClusterResourceBinding sets up the rollout controller with the Manager for cluster scoped resources.
 // The rollout controller watches resource snapshots and resource bindings.
-// It reconciles on the CRP when a new resource resourceBinding is created or an existing resource binding is created/updated.
-func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
-	r.recorder = mgr.GetEventRecorderFor("rollout-controller")
-	return runtime.NewControllerManagedBy(mgr).Named("rollout-controller").
+// It reconciles on the CRP when a new cluster resource binding is created or an existing cluster resource binding is created/updated.
+func (r *Reconciler) SetupWithManagerForClusterResourceBinding(mgr runtime.Manager) error {
```
zhiying-lin:
```suggestion
func (r *Reconciler) SetupWithManagerForClusterResourcePlacement(mgr runtime.Manager) error {
```

using CRP or RP may be more accurate, as you watch not just binding.

pkg/controllers/rollout/controller.go
```diff
@@ -678,12 +678,12 @@ func (r *Reconciler) updateBindings(ctx context.Context, bindings []toBeUpdatedB
...
+	r.recorder = mgr.GetEventRecorderFor("cluster-scoped-rollout-controller")
+	return runtime.NewControllerManagedBy(mgr).Named("cluster-scoped-rollout-controller").
```
zhiying-lin:
```suggestion
	return runtime.NewControllerManagedBy(mgr).Named("cluster-resource-placement-rollout-controller").
```

pkg/controllers/rollout/controller.go
```diff
@@ -678,12 +678,12 @@ func (r *Reconciler) updateBindings(ctx context.Context, bindings []toBeUpdatedB
...
 		WithOptions(ctrl.Options{MaxConcurrentReconciles: r.MaxConcurrentReconciles}). // set the max number of concurrent reconciles
 		Watches(&placementv1beta1.ClusterResourceSnapshot{}, handler.Funcs{
 			CreateFunc: func(ctx context.Context, e event.CreateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
```
zhiying-lin: why not refactoring handler.Funcs as a common variable, so that it can be reused by CRP and RP?

ditto to other handler.Funcs

pkg/controllers/rollout/controller.go
```diff
@@ -880,18 +934,52 @@ func handleResourceSnapshot(snapshot client.Object, q workqueue.TypedRateLimitin
...
+
+	if ro.Spec.Placement == nil {
+		klog.V(2).InfoS("ResourceOverride does not have a placement specified, skipping enqueue", "resourceOverride", klog.KObj(ro))
+		return
+	}
+
+	namespace := ""
+	if enqueueCRP {
+		if ro.Spec.Placement.Scope == placementv1beta1.NamespaceScoped {
+			klog.V(2).InfoS("Skipping a resourceOverride delete event with resource placement", "resourceOverride", klog.KObj(ro))
```
zhiying-lin: need to update the log, assuming this util now can handle any event, other than delete

pkg/controllers/rollout/controller.go
```diff
@@ -880,18 +934,52 @@ func handleResourceSnapshot(snapshot client.Object, q workqueue.TypedRateLimitin
...
+			return
+		}
+	} else {
+		if ro.Spec.Placement.Scope != placementv1beta1.NamespaceScoped {
```
zhiying-lin, line 958: please make sure you validate this using CEL so that it can be blocked when customer creates/updates this invalid case,

pkg/controllers/rollout/controller.go
```diff
@@ -818,42 +866,48 @@ func handleResourceOverrideSnapshot(o client.Object, q workqueue.TypedRateLimiti
 	}
```
zhiying-lin, line 859: should we add a flag to indicate whether the util should enqueue the RP or CRP?

we don't expect the CRP rollout controller to handles the RP key.

ditto to other places

pkg/controllers/rollout/controller_test.go
```diff
@@ -162,12 +163,7 @@ func TestReconcilerHandleResourceSnapshot(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			queue := &controllertest.Queue{TypedInterface: workqueue.NewTypedRateLimitingQueue[reconcile.Request](workqueue.DefaultTypedItemBasedRateLimiter[reconcile.Request]())}
 			handleResourceSnapshot(tt.snapshot, queue)
-			if tt.shouldEnqueue && queue.Len() == 0 {
-				t.Errorf("handleResourceSnapshot test `%s` didn't queue the object when it should enqueue", name)
-			}
-			if !tt.shouldEnqueue && queue.Len() != 0 {
-				t.Errorf("handleResourceSnapshot test `%s` queue the object when it should not enqueue", name)
-			}
+			validateEnqueueBehavior(t, name, queue, tt.shouldEnqueue, nil)
```
zhiying-lin: need to update existing ut, just validating if shouldEnqueue is not good enough for now. We need to check the key.

pkg/controllers/rollout/controller_test.go
```diff
@@ -198,16 +194,284 @@ func TestReconcilerHandleResourceBinding(t *testing.T) {
...
+		})
+	}
+}
+
+func TestReconcilerHandleResourceOverrideSnapshot(t *testing.T) {
+	tests := map[string]struct {
+		snapshot           client.Object
+		shouldEnqueue      bool
+		expectedEnqueueKey *reconcile.Request
```
zhiying-lin:
```suggestion
		wantEnqueueKey *reconcile.Request
```

pkg/controllers/rollout/controller_test.go
```diff
@@ -198,16 +194,284 @@ func TestReconcilerHandleResourceBinding(t *testing.T) {
...
-			}
+			validateEnqueueBehavior(t, name, queue, tt.shouldEnqueue, nil)
+		})
+	}
+}
+
+func TestReconcilerHandleResourceOverrideSnapshot(t *testing.T) {
+	tests := map[string]struct {
+		snapshot           client.Object
+		shouldEnqueue      bool
```
zhiying-lin: we don't need the bool any more. Just compare the whole list.

pkg/controllers/rollout/controller_test.go
```diff
@@ -198,16 +194,284 @@ func TestReconcilerHandleResourceBinding(t *testing.T) {
...
+			queue := &controllertest.Queue{TypedInterface: workqueue.NewTypedRateLimitingQueue[reconcile.Request](workqueue.DefaultTypedItemBasedRateLimiter[reconcile.Request]())}
+			handleResourceOverride(tt.resourceOverride, queue, tt.enqueueCRP)
+			validateEnqueueBehavior(t, name, queue, tt.shouldEnqueue, tt.expectedEnqueueKey)
+		})
+	}
+}
+
+// validateEnqueueBehavior is a helper function to validate queue enqueue behavior in tests
+func validateEnqueueBehavior(t *testing.T, testName string, queue *controllertest.Queue, shouldEnqueue bool, expectedEnqueueKey *reconcile.Request) {
+	if shouldEnqueue && queue.Len() == 0 {
```
zhiying-lin: please compare the keys directly, instead of one by one,

if you're using cmp.Equal(got, want..)

you don't need this util and the readability is better.

zhiying-lin (commented):
assuming you will add IT in a separate PR

pkg/controllers/clusterresourceplacement/controller.go
jwtty: ObjectRef has the String() function: https://github.com/kubernetes/klog/blob/v2.130.1/k8s_references.go#L34

pkg/controllers/rollout/controller.go
jwtty: I originally didn't do this because there are multiple bindings in a loop. I created a kRef within each loop.

pkg/controllers/rollout/controller.go
jwtty: fixed, thanks!

pkg/controllers/rollout/controller.go
jwtty: fixed

pkg/controllers/rollout/controller.go
jwtty: fixed

pkg/controllers/rollout/controller.go
jwtty: By invalid case, you mean a CRO with NamespacedScoped right? I'm preparing a PR for that. For RO, the scope can be either Cluster or Namespaced.

pkg/controllers/rollout/controller_test.go
jwtty: fixed

pkg/controllers/rollout/controller_test.go
jwtty: Not sure I follow. Queue does not support "List()", only "Get()" by getting one resource each time.

pkg/controllers/rollout/controller_test.go
jwtty: Updated all tests.

pkg/controllers/rollout/controller_test.go
jwtty: Use cmp.Equal instead.

pkg/controllers/rollout/controller_test.go
jwtty: removed.

pkg/controllers/rollout/controller.go
jwtty: refactored.

pkg/controllers/rollout/controller.go
jwtty: The flag should only be added for resourceOverride and resourceOverrideSnapshot. For the other types, e.g. bindings, CRP rollout controller will only be triggered by clusterResourceBinding while RP rollout controller will only be triggered by resourceBinding. Only resourceOverride and resourceOverrideSnapshot are watched by both CRP and RP controllers.

pkg/controllers/rollout/controller.go
jwtty: what's the recommended way?

jwtty: something like GetNamespace()/GetName()?

pkg/controllers/rollout/controller.go
```diff
@@ -732,96 +716,105 @@ func (r *Reconciler) SetupWithManagerForClusterResourceBinding(mgr runtime.Manag
...
+			handleResourceSnapshot(e.Object, q)
+		},
+	}
+}
+
+// resourceOverrideSnapshotHandlerFuncs returns the handler functions for resource override snapshot events.
+func resourceOverrideSnapshotHandlerFuncs(enqueueCRP bool) handler.Funcs {
+	return handler.Funcs{
+		CreateFunc: func(ctx context.Context, e event.CreateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+			klog.V(2).InfoS("Handling a resourceOverrideSnapshot create event", "resourceOverrideSnapshot", klog.KObj(e.Object))
```
zhiying-lin: add the enqueueCRP bool to the log? so that we can distinguish whether the log is for CRP or RP

pkg/controllers/rollout/controller.go
```diff
@@ -884,17 +877,24 @@ func handleResourceOverrideSnapshot(o client.Object, q workqueue.TypedRateLimiti
...
-			NamespacedName: types.NamespacedName{Name: placementInOverride.Name},
-		})
+		if placementInOverride.Scope != placementv1beta1.NamespaceScoped {
+			klog.V(2).InfoS("Skipping a resourceOverrideSnapshot event with cluster resource placement", "resourceOverrideSnapshot", snapshotKRef)
+			return
+		}
+		namespace = snapshot.GetNamespace()
 	}
+
+	klog.V(2).InfoS("Handling a resourceOverrideSnapshot event", "resourceOverrideSnapshot", snapshotKRef)
```
zhiying-lin: add the enqueueCRP to the log?

pkg/controllers/rollout/controller.go
zhiying-lin: for this particular case, usually we not only create the error but also log the error.

in the error, you can just log the binding name while in the log, you can use the KObj to log the full name.

pkg/controllers/rollout/controller_test.go
```diff
@@ -155,59 +158,380 @@ func TestReconcilerHandleResourceSnapshot(t *testing.T) {
 					},
 				},
 			},
-			shouldEnqueue: false,
+			wantEnqueueKeys: nil,
```
zhiying-lin, line 189: add at least one test for the RP? to validate if the key is correct or not.

pkg/controllers/rollout/controller_test.go
```diff
@@ -155,59 +158,380 @@ func TestReconcilerHandleResourceSnapshot(t *testing.T) {
...
-			if !tt.shouldEnqueue && queue.Len() != 0 {
-				t.Errorf("handleResourceSnapshot test `%s` queue the object when it should not enqueue", name)
-			}
+			validateEnqueueBehavior(t, name, queue, tt.wantEnqueueKeys)
 		})
 	}
 }
 
-func TestReconcilerHandleResourceBinding(t *testing.T) {
+func TestReconcilerHandleBinding(t *testing.T) {
```
zhiying-lin: use the func name as the test name?

pkg/controllers/rollout/controller_test.go
```diff
@@ -155,59 +158,380 @@ func TestReconcilerHandleResourceSnapshot(t *testing.T) {
...
+							Name:  "test-placement",
+							Scope: placementv1beta1.ClusterScoped,
+						},
+					},
+				},
+			},
+			enqueueCRP:      false,
+			wantEnqueueKeys: nil,
+		},
+		"test enqueue a new latest ResourceOverrideSnapshot with namespace scoped placement": {
```
zhiying-lin:
```suggestion
		"test enqueue a new latest ResourceOverrideSnapshot with namespace scoped placement while enqueueCRP=false": {
```

the test name is not consistent with the above one

pkg/controllers/rollout/controller.go
jwtty: added

pkg/controllers/rollout/controller.go
jwtty: added

pkg/controllers/rollout/controller_test.go
jwtty: fixed, not sure why they were named this way.

pkg/controllers/rollout/controller_test.go
jwtty: added.

pkg/controllers/rollout/controller_test.go
jwtty: missed this one, added.

pkg/controllers/rollout/controller.go
jwtty: replaced

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -1120,10 +1121,7 @@ func (r *Reconciler) determineRolloutStateForPlacementWithExternalRolloutStrateg
 	} else {
 		placementStatus.ObservedResourceIndex = observedResourceIndex
 		// Construct placement key for the resource collection function
-		placementKey := placementObj.GetName()
-		if placementObj.GetNamespace() != "" {
-			placementKey = placementObj.GetNamespace() + "/" + placementObj.GetName()
-		}
+		placementKey := controller.GetObjectKeyFromNamespaceName(placementObj.GetNamespace(), placementObj.GetName())
```
ryanzhang-oss, line 1124: For next PR, I think we should change the CollectResourceIdentifiersFromResourceSnapshot  func signature to take in namespacedName instead of string

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -189,7 +189,8 @@ func (r *Reconciler) handleUpdate(ctx context.Context, placementObj fleetv1beta1
 			klog.ErrorS(err, "Failed to extract the resource index from the resourceSnapshot", "placement", placementKObj, "resourceSnapshot", latestResourceSnapshotKObj)
 			return ctrl.Result{}, controller.NewUnexpectedBehaviorError(err)
 		}
-		selectedResourceIDs, err = controller.CollectResourceIdentifiersUsingMasterResourceSnapshot(ctx, r.Client, placementObj.GetName(), latestResourceSnapshot, strconv.Itoa(latestResourceSnapshotIndex))
+		placementKey := controller.GetObjectKeyFromNamespaceName(placementObj.GetNamespace(), placementObj.GetName())
```
ryanzhang-oss, line 192: CollectResourceIdentifiersUsingMasterResourceSnapshot probably should take the namespacedName instead of string, next PR.

zhiying-lin (approved):

zhiying-lin (approved):

PR #159, opened 2025-07-31
interface: add placementScope in override API

jwtty:

Add scope field in placementRef in override API so that controller can differentiate if the placement referenced is clusterResourcePlacement or resourcePlacement.

API design followed CRD:
https://github.com/kubernetes/kubernetes/blob/032142c53e54e41aff516cf9515c175afeeffd32/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/types.go#L277C1-L283C2

Fixes #

I have:

### DIFF
--- a/apis/placement/v1alpha1/override_types.go
+++ b/apis/placement/v1alpha1/override_types.go
@@ -68,12 +68,32 @@ type ClusterResourceOverrideSpec struct {
 	Policy *OverridePolicy `json:"policy"`
 }
 
+// ResourceScope defines the scope of placement reference.
+type ResourceScope string
+
+const (
+	// ClusterScoped indicates placement is cluster-scoped.
+	ClusterScoped ResourceScope = "Cluster"
+
+	// NamespaceScoped indicates placement is namespace-scoped.
+	NamespaceScoped ResourceScope = "Namespaced"
+)
+
 // PlacementRef is the reference to a placement.
 // For now, we only support ClusterResourcePlacement.
 type PlacementRef struct {
 	// Name is the reference to the name of placement.
 	// +required
 	Name string `json:"name"`
+
+	// Scope defines the scope of the placement.
+	// A clusterResourceOverride can only reference a clusterResourcePlacement (cluster-scoped),
+	// and a resourceOverride can reference either a clusterResourcePlacement or resourcePlacement (namespaced).
+	// The referenced resourcePlacement must be in the same namespace as the resourceOverride.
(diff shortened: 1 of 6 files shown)


apis/placement/v1alpha1/override_types.go
```diff
@@ -68,12 +68,32 @@ type ClusterResourceOverrideSpec struct {
...
 	Name string `json:"name"`
+
+	// Scope defines the scope of the placement.
+	// A clusterResourceOverride can only reference a clusterResourcePlacement (cluster-scoped),
+	// and a resourceOverride can reference either a clusterResourcePlacement or resourcePlacement (namespaced).
+	// The referenced resourcePlacement must be in the same namespace as the resourceOverride.
+	// +kubebuilder:validation:Enum=Cluster;Namespaced
+	// +kubebuilder:default=Cluster
+	// +optional
+	Scope ResourceScope `json:"scope,omitempty"`
```
circy9, line 96: I would expect a reference object contains the unique ids for an object like the following. What are the reasons to add scope instead of namespace?

PlacementRef {
  namespace
  name
}

Similar with https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-reference/?utm_source=chatgpt.com

jwtty: The objectReference with uid is mostly set by controller. This is a field expecting user input, which normally only needs namespace/name. We do not add namespace because the RP referenced must have the same namespace as the RO and if a CRP is referenced, it does not need a namespace.

ryanzhang-oss: this field is not designed for a specific placement but more for an intention so we can know if it's for RP or a CRP.

circy9: SGTM

circy9 (approved):

apis/placement/v1beta1/override_types.go
```diff
@@ -67,12 +67,32 @@ type ClusterResourceOverrideSpec struct {
...
+	// ClusterScoped indicates placement is cluster-scoped.
+	ClusterScoped ResourceScope = "Cluster"
+
+	// NamespaceScoped indicates placement is namespace-scoped.
+	NamespaceScoped ResourceScope = "Namespaced"
+)
+
 // PlacementRef is the reference to a placement.
 // For now, we only support ClusterResourcePlacement.
 type PlacementRef struct {
```
zhiying-lin, line 83: could you please add a CEL validation on the CRO to make sure that the scope should be "cluster" only?

PR #169, opened 2025-08-06
feat: enable resource change detector for RP

jwtty:
Enable resource change detector to start RP controller and add the namespaced resources to skip watch list.

Fixes #

I have:

### DIFF
--- a/pkg/utils/apiresources.go
+++ b/pkg/utils/apiresources.go
@@ -65,9 +65,19 @@ var (
 		Kind:  placementv1beta1.ClusterResourcePlacementKind,
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/utils/apiresources.go
```diff
@@ -177,16 +197,19 @@ func NewResourceConfig(isAllowList bool) *ResourceConfig {
 
 	// disable all fleet placement resources except for the envelope type
 	r.AddGroupKind(ClusterResourcePlacementGK)
```
zhiying-lin, line 199: could you please add a task or TODO? we need to support group instead of listing all the apis.

zhiying-lin (approved):

PR #171, opened 2025-08-06
feat: enable scheduler for RP

jwtty:
Enable Scheduler to support resourcePlacements. Add integration tests.

Fixes #

I have:


test/scheduler/actuals_test.go
```diff
@@ -25,113 +25,160 @@ import (
...
+		if bindingCount := len(bindingList.GetBindingObjs()); bindingCount != 0 {
 			return fmt.Errorf("%d bindings have been created unexpectedly", bindingCount)
 		}
 
 		return nil
 	}
 }
 
-func crpSchedulerFinalizerAddedActual(crpName string) func() error {
+func placementSchedulerFinalizerAddedActual(placementKey string) func() error {
```
zhiying-lin, line 53: instead of passing the key, we can change the placementKey type to types.NamespacedName to avoid building/parsing from the key

jwtty: I chose this way so that existing CRP tests don't need change. Otherwise, all existing CRP tests all need to updated too. And it's too much.

ryanzhang-oss: add a TODO and leave that for the next PR?

zhiying-lin: i;m ok with addressing it in the next PR. It should not be hard for AI to do this kind of job :)

jwtty: sure, will do in next PR

PR #175, opened 2025-08-07
test: add rollout controller integration tests for RP

jwtty:
Add rollout controller IT for RP.

Fixes #

I have:

### DIFF
--- a/pkg/controllers/rollout/suite_test.go
+++ b/pkg/controllers/rollout/suite_test.go
@@ -26,6 +26,8 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/controllers/rollout/suite_test.go
```diff
@@ -104,6 +106,14 @@ var _ = BeforeSuite(func() {
 	By("set k8s client same as the controller manager")
 	k8sClient = mgr.GetClient()
 
+	// Create test namespace
```
ryanzhang-oss, line 109: do all the test run under this namespace? When will it be deleted?

zhiying-lin (approved):
We don't have any override tests in the existing it tests. Probably need to add one. So that we can cover the case when the override triggers the rollout.

pkg/controllers/rollout/suite_test.go
jwtty: All the RP related tests run under this namesapce. It will be deleted in AfterSuite when the test env is teared down.

PR #179, opened 2025-08-11
fix: set the correct value for work applier fast backoff

michaelawyu:

This PR fixes a minor issue where the default value for work applier backoff is set to the same value as the one for slow backoff.

I have:

N/A

N/A


ryanzhang-oss (commented):
it's time to add some tests?

michaelawyu:
> it's time to add some tests?

Yeah, Ryan, would you prefer if I send them in a separate PR, or would you prefer I add it to this one?

PR #181, opened 2025-08-11
feat: add feature gate support and gate property provider capabilities

michaelawyu:

This PR adds feature gate support and gate property provider capabilities so that users can turn off certain properties (e.g., costs, available resource).

I have:

- [x] Unit tests
- [x] Integration tests

Additional tests will be added in a separate PR.

### DIFF
--- a/docker/refresh-token.Dockerfile
+++ b/docker/refresh-token.Dockerfile
@@ -1,4 +1,4 @@
-# Build the hubagent binary
+# Build the refreshtoken binary
 FROM mcr.microsoft.com/oss/go/microsoft/golang:1.24.6 AS builder
 
 WORKDIR /workspace
--- a/pkg/controllers/bindingwatcher/suite_test.go
+++ b/pkg/controllers/bindingwatcher/suite_test.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"path/filepath"
 	"testing"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -122,6 +123,14 @@ var _ = BeforeSuite(func() {
 		err = mgr.Start(ctx)
 		Expect(err).Should(Succeed(), "failed to run manager")
 	}()
+
+	// Note (chenyu1): for the binding watcher integration tests, must wait for the cache to sync
+	// before moving onto the test stage, otherwise some Update events might not be caught properly.
+	//
+	// This is wrapped in an Eventually block as the manager might not have started yet.
+	Eventually(func() bool {
+		return mgr.GetCache().WaitForCacheSync(ctx)
+	}, time.Second*10, time.Second*2).To(BeTrue(), "failed to wait for cache to sync")
 })
 
 var _ = AfterSuite(func() {
--- a/pkg/controllers/bindingwatcher/watcher_integration_test.go
+++ b/pkg/controllers/bindingwatcher/watcher_integration_test.go
@@ -40,7 +40,7 @@ const (
 	testReason1                      = "testReason1"
 	testReason2                      = "testReason2"
 
-	eventuallyTimeout    = time.Second * 20
+	eventuallyTimeout    = time.Second * 10
 	consistentlyDuration = time.Second * 10
 	interval             = time.Millisecond * 250
 )
--- a/pkg/controllers/placement/controller.go
+++ b/pkg/controllers/placement/controller.go
@@ -153,7 +153,7 @@ func (r *Reconciler) handleUpdate(ctx context.Context, placementObj fleetv1beta1
 		// TODO, create a separate user type error struct to improve the user facing messages
 		scheduleCondition := metav1.Condition{
 			Status:             metav1.ConditionFalse,
-			Type:               string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType),
+			Type:               getPlacementScheduledConditionType(placementObj),
 			Reason:             condition.InvalidResourceSelectorsReason,
 			Message:            fmt.Sprintf("The resource selectors are invalid: %v", err),
 			ObservedGeneration: placementObj.GetGeneration(),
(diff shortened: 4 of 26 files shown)


pkg/utils/featuregates/featuregates.go
```diff
@@ -0,0 +1,125 @@
...
+type mutableFeatureGate struct {
+	// MapStringBool is a Kubernetes implementation of a flag Value that supports
+	// multiple comma-separated key-value pairs as CLI arguments.
+	//
+	// Note (chenyu1):
+	*k8sflag.MapStringBool
+
+	closed             bool
+	registeredFeatures map[FeatureName]FeatureSpec
+
```
ryanzhang-oss: why can't we just have one map to store all the features? Why need two maps?

michaelawyu: Hi Ryan! The `k8sflag.MapStringBool` is a `flag.Value` that does the CLI argument parsing. I added the `registered Features` map so that components can register the feature gates they need (plus their default values), and the feature implementation will use this map to cross-ref the `MapStringBool` to ensure that unspecified gates will have proper values, specified gates will override default values, and unsupported/incorrect values will be rejected when the agent starts up.

michaelawyu: This is just a minor adaptation from K8s' feature gate implementation; I didn't directly use the code from the api-machinery pkg as currently we do not have any version control in our feature gate setup so much of the functionaliy in that pkg is irrelevant to us.

pkg/utils/featuregates/featuregates.go
```diff
@@ -0,0 +1,125 @@
...
+	MemberAgentFeatureGate MutableFeatureGate = NewFeatureGate()
+)
+
+// FeatureName is the name of a feature.
+type FeatureName string
+
+// FeatureSpec is the enablement setup for a feature.
+type FeatureSpec struct {
+	// Default is the default enablement state of the feature.
+	Default bool
```
ryanzhang-oss: what other fields are we planning for the future? Are all features only need turn on and off? If not, should this be an enum?

pkg/propertyprovider/azure/provider_integration_test.go
```diff
@@ -725,39 +733,51 @@ var _ = Describe("azure property provider", func() {
 	Context("add a new node", Serial, Ordered, func() {
 		BeforeAll(shouldCreateNodes(nodes[0]))
 
-		It("should report correct properties", shouldReportCorrectPropertiesForNodes(nodes[0:1], nil))
+		It("should report correct properties", func() {
```
michaelawyu, line 736: Note: this encapsulation is needed so that Ginkgo tree building will not pass nil values to the function.

cmd/memberagent/main.go
```diff
@@ -103,7 +102,9 @@ var (
...
 	workApplierRequeueRateLimiterSkipToFastBackoffForAvailableOrDiffReportedWorkObjs = flag.Bool("work-applier-requeue-rate-limiter-skip-to-fast-backoff-for-available-or-diff-reported-work-objs", true, "If set, the rate limiter will skip the slow backoff stage and start fast backoff immediately for work objects that are available or have diff reported.")
-	featureGates                                                                     = featuregates.MemberAgentFeatureGate
+	// Azure property provider feature gates.
+	isAzurePropertyProviderCostCollectionEnabled               = flag.Bool("enable-cost-collection-in-azure-property-provider", true, "If set, the Azure property provider will collect cost information in the member cluster.")
+	isAzurePropertyProviderAvailableResourcesCollectionEnabled = flag.Bool("enable-available-resources-collection-in-azure-property-provider", true, "If set, the Azure property provider will track available resources in the member cluster.")
```
ryanzhang-oss: nit: anyway to make the flat name a bit short?

pkg/propertyprovider/azure/provider_integration_test.go
```diff
@@ -132,167 +132,175 @@ var (
...
+				}
+			}
 
-				wantProperties := map[clusterv1beta1.PropertyName]clusterv1beta1.PropertyValue{
-					propertyprovider.NodeCountProperty: {
-						Value: fmt.Sprintf("%d", len(nodes)),
+			var wantConditions []metav1.Condition
+			switch {
+			case !isCostsEnabled:
+				// Cost calculation has been disabled. No need to add a condition.
```
ryanzhang-oss, line 246: Do we want to test when the cost is disabled, we don't collect?

pkg/propertyprovider/azure/provider_integration_test.go
```diff
@@ -132,167 +132,175 @@ var (
...
+						corev1.ResourceMemory: allocatableMemoryCapacity,
+					},
+				},
+				Conditions: wantConditions,
+			}
 
-				res := p.Collect(ctx)
-				if diff := cmp.Diff(res, expectedRes, ignoreObservationTimeFieldInPropertyValue); diff != "" {
-					return fmt.Errorf("property collection response (-got, +want):\n%s", diff)
+			if isAvailableResourcesEnabled {
```
ryanzhang-oss, line 291: do we have the negative test?

ryanzhang-oss (commented):
is there no e2e for this?

cmd/memberagent/main.go
michaelawyu: Done 👌

pkg/propertyprovider/azure/provider_integration_test.go
michaelawyu: Hi Ryan! Unit tests have been added for these scenarios (`TestCollectWithDisabledFeatures` function in `provider_test.go`).

pkg/propertyprovider/azure/provider_integration_test.go
michaelawyu: Hi Ryan! Yeah, for integration tests we do; it's in the `feature gates` node of the integration test (line 988). One tests the case where costs are disabled, and the other where available resources are disabled.

michaelawyu: E2E test cases need to be added separately though; will submit subsequent PRs.

PR #183, opened 2025-08-12
fix: refactor the work applier code to reduce conditional checking + address an issue in work applier where some errors might not get properly surfaced in ReportDiff mode

michaelawyu:

This RP includes:

* Minor refactoring of the code (unifying apply and report diff result types) to reduce conditional checks, which has been a source of bugs
* Address an issue (caused by the distinction between apply and report diff result types) where in ReportDiff mode some errors (e.g., decoding, missing objects, etc.) are not being surfaced properly.

I have:

- [x] Unit tests
- [x] Integration tests
- [x] E2E tests

Additional tests will be submitted in a separate PR.


pkg/controllers/workapplier/controller.go
```diff
@@ -266,33 +264,61 @@ func NewReconciler(
...
 
-	ManifestProcessingApplyResultTypeAppliedWithFailedDriftDetectionDescription = "Manifest has been applied successfully, but drift detection has failed"
+	ManifestProcessingApplyOrReportDiffResultTypeAppliedWithFailedDriftDetectionDescription = "Manifest has been applied successfully, but drift detection has failed"
 
 	// The result type and description for successful processing attempts.
-	ManifestProcessingApplyResultTypeApplied manifestProcessingAppliedResultType = "Applied"
+	ManifestProcessingApplyOrReportDiffResultTypeApplied ManifestProcessingApplyOrReportDiffResultType = "Applied"
 
-	ManifestProcessingApplyResultTypeAppliedDescription = "Manifest has been applied successfully"
+	ManifestProcessingApplyOrReportDiffResultTypeAppliedDescription = "Manifest has been applied successfully"
```
ryanzhang-oss: nit: move the ManifestProcessingApplyOrReportDiffResultType together and group the description var to another place, the names are so long it's hard to read and figure who is who

pkg/controllers/workapplier/controller.go
```diff
@@ -266,33 +264,61 @@ func NewReconciler(
...
+	// The result type for completed diff reportings.
+	ManifestProcessingApplyOrReportDiffResultTypeFoundDiff   ManifestProcessingApplyOrReportDiffResultType = "FoundDiff"
+	ManifestProcessingApplyOrReportDiffResultTypeNoDiffFound ManifestProcessingApplyOrReportDiffResultType = "NoDiffFound"
+
+	ManifestProcessingApplyOrReportDiffResultTypeNoDiffFoundDescription = "No diff has been found between the hub cluster and the member cluster"
+	ManifestProcessingApplyOrReportDiffResultTypeFoundDiffDescription   = "Diff has been found between the hub cluster and the member cluster"
+)
+
+var (
+	// A set for all apply related result types.
```
ryanzhang-oss, line 312: why there is no report related results type list?

pkg/controllers/workapplier/controller.go
```diff
@@ -266,33 +264,61 @@ func NewReconciler(
...
+	manifestProcessingApplyResTypSet = set.New[ManifestProcessingApplyOrReportDiffResultType](
```
ryanzhang-oss:
```suggestion
	manifestProcessingApplyResTypSet = set.New(
```

pkg/controllers/workapplier/status.go
```diff
@@ -102,9 +103,9 @@ func (r *Reconciler) refreshWorkStatus(
 		if bundle.inMemberClusterObj != nil {
 			inMemberClusterObjGeneration = bundle.inMemberClusterObj.GetGeneration()
 		}
-		setManifestAppliedCondition(manifestCond, bundle.applyResTyp, bundle.applyErr, inMemberClusterObjGeneration)
+		setManifestAppliedCondition(manifestCond, isReportDiffModeOn, bundle.applyOrReportDiffResTyp, bundle.applyOrReportDiffErr, inMemberClusterObjGeneration)
 		setManifestAvailableCondition(manifestCond, bundle.availabilityResTyp, bundle.availabilityErr, inMemberClusterObjGeneration)
-		setManifestDiffReportedCondition(manifestCond, bundle.reportDiffResTyp, bundle.reportDiffErr, inMemberClusterObjGeneration)
+		setManifestDiffReportedCondition(manifestCond, isReportDiffModeOn, bundle.applyOrReportDiffResTyp, bundle.applyOrReportDiffErr, inMemberClusterObjGeneration)
```
ryanzhang-oss, line 108: why even call into setManifestAppliedCondition if isReportDiffModeOn==true? and call setManifestDiffReportedCondition if isReportDiffModeOn==false?

pkg/controllers/workapplier/controller.go
michaelawyu: Hi Ryan! The set for apply op result types is added so that the work applier can do a sanity check to make sure that we are not populating the status reasons wrong; the diff reporting status does not need it as the reason might actually be non-diff-reporting related (e.g., decoding error).

pkg/controllers/workapplier/controller.go
michaelawyu: Fixed.

pkg/controllers/workapplier/status.go
michaelawyu: Hi Ryan! This is needed to clear out the `Applied`/`DiffReported` conditions when the user switches apply strategies (CSA/SSA <-> ReportDiff). These conditions use the actual object generations (not the work generations) as the condition observed generations, so we have to drop them to ensure correctness

michaelawyu: The CRP/RP controller uses the same pattern.

pkg/controllers/workapplier/controller.go
michaelawyu: Hi Ryan! I have changed them to shorter names and re-grouped them.

PR #184, opened 2025-08-12
feat: update the resource_selector to support RP

zhiying-lin:

update the resource_selector to support RP

Fixes #

I have:

added ut

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -159,11 +159,16 @@ cmd/memberagent/        # Member agent main and setup
 ## Testing Patterns
 
 ### Unit Tests
-- Use `testify` for assertions
+- Avoid the use of ‘assert’ libraries.
 - Controllers use `envtest` for integration testing with real etcd
 - Mock external dependencies with `gomock`
 - Unit test files: `<go_file>_test.go` in same directory
 - Table-driven test style preferred
+- Use cmp.Equal for equality comparison and cmp.Diff to obtain a human-readable diff between objects.
+- Test outputs should output the actual value that the function returned before printing the value that was expected. A usual format for printing test outputs is “YourFunc(%v) = %v, want %v”.
+- If your function returns a struct, don’t write test code that performs an individual comparison for each field of the struct. Instead, construct the struct that you’re expecting your function to return, and compare in one shot using diffs or deep comparisons. The same rule applies to arrays and maps.
+- If your struct needs to be compared for approximate equality or some other kind of semantic equality, or it contains fields that cannot be compared for equality (e.g. if one of the fields is an io.Reader), tweaking a cmp.Diff or cmp.Equal comparison with cmpopts options such as cmpopts.IgnoreInterfaces may meet your needs (example); otherwise, this technique just won’t work, so do whatever works.
+- If your function returns multiple return values, you don’t need to wrap those in a struct before comparing them. Just compare the return values individually and print them.
 
 ### Integration Tests  
 - Located in `test/integration/` and `test/scheduler/`
--- a/pkg/controllers/clusterresourceplacement/resource_selector.go
+++ b/pkg/controllers/clusterresourceplacement/resource_selector.go
@@ -129,7 +129,7 @@ func convertResourceSelector(old []fleetv1alpha1.ClusterResourceSelector) []flee
 }
 
 // gatherSelectedResource gets all the resources according to the resource selector.
-func (r *Reconciler) gatherSelectedResource(placementName types.NamespacedName, selectors []fleetv1beta1.ClusterResourceSelector) ([]*unstructured.Unstructured, error) {
+func (r *Reconciler) gatherSelectedResource(placementKey types.NamespacedName, selectors []fleetv1beta1.ClusterResourceSelector) ([]*unstructured.Unstructured, error) {
 	var resources []*unstructured.Unstructured
 	var resourceMap = make(map[fleetv1beta1.ResourceIdentifier]bool)
 	for _, selector := range selectors {
@@ -145,10 +145,10 @@ func (r *Reconciler) gatherSelectedResource(placementName types.NamespacedName, 
 		}
 		var objs []runtime.Object
 		var err error
-		if gvk == utils.NamespaceGVK && placementName.Namespace == "" {
-			objs, err = r.fetchNamespaceResources(selector, placementName.Name)
+		if gvk == utils.NamespaceGVK && placementKey.Namespace == "" {
+			objs, err = r.fetchNamespaceResources(selector, placementKey.Name)
 		} else {
-			objs, err = r.fetchResources(selector, placementName)
+			objs, err = r.fetchResources(selector, placementKey)
 		}
 		if err != nil {
 			return nil, err
@@ -164,7 +164,7 @@ func (r *Reconciler) gatherSelectedResource(placementName types.NamespacedName, 
 			}
 			if _, exist := resourceMap[ri]; exist {
 				err = fmt.Errorf("found duplicate resource %+v", ri)
-				klog.ErrorS(err, "User selected one resource more than once", "resource", ri, "placement", placementName)
+				klog.ErrorS(err, "User selected one resource more than once", "resource", ri, "placement", placementKey)
 				return nil, controller.NewUserError(err)
 			}
 			resourceMap[ri] = true
@@ -231,15 +231,15 @@ func buildApplyOrderMap() map[string]int {
 }
 
 // fetchResources retrieves the objects based on the selector.
-func (r *Reconciler) fetchResources(selector fleetv1beta1.ClusterResourceSelector, placementName types.NamespacedName) ([]runtime.Object, error) {
-	klog.V(2).InfoS("Start to fetch resources by the selector", "selector", selector, "placement", placementName)
+func (r *Reconciler) fetchResources(selector fleetv1beta1.ClusterResourceSelector, placementKey types.NamespacedName) ([]runtime.Object, error) {
+	klog.V(2).InfoS("Start to fetch resources by the selector", "selector", selector, "placement", placementKey)
 	gk := schema.GroupKind{
 		Group: selector.Group,
 		Kind:  selector.Kind,
 	}
 	restMapping, err := r.RestMapper.RESTMapping(gk, selector.Version)
 	if err != nil {
-		return nil, controller.NewUserError(fmt.Errorf("invalid placement %s, failed to get GVR of the selector: %w", placementName, err))
+		return nil, controller.NewUserError(fmt.Errorf("invalid placement %s, failed to get GVR of the selector: %w", placementKey, err))
 	}
 	gvr := restMapping.Resource
 	gvk := schema.GroupVersionKind{
@@ -249,21 +249,21 @@ func (r *Reconciler) fetchResources(selector fleetv1beta1.ClusterResourceSelecto
 	}
 
 	isNamespacedResource := !r.InformerManager.IsClusterScopedResources(gvk)
-	if isNamespacedResource && placementName.Namespace == "" {
+	if isNamespacedResource && placementKey.Namespace == "" {
 		// If it's a namespace-scoped resource but placement has no namespace, return error.
-		err := fmt.Errorf("invalid placement %s: cannot select namespace-scoped resource %v without a namespace", placementName, gvr)
+		err := fmt.Errorf("invalid placement %s: cannot select namespace-scoped resource %v in a clusterResourcePlacement", placementKey, gvr)
 		klog.ErrorS(err, "Invalid resource selector", "selector", selector)
 		return nil, controller.NewUserError(err)
-	} else if !isNamespacedResource && placementName.Namespace != "" {
+	} else if !isNamespacedResource && placementKey.Namespace != "" {
 		// If it's a cluster-scoped resource but placement has a namespace, return error.
-		err := fmt.Errorf("invalid placement %s: cannot select cluster-scoped resource %v with a namespace", placementName, gvr)
+		err := fmt.Errorf("invalid placement %s: cannot select cluster-scoped resource %v in a resourcePlacement", placementKey, gvr)
 		klog.ErrorS(err, "Invalid resource selector", "selector", selector)
 		return nil, controller.NewUserError(err)
 	}
 
 	if !r.InformerManager.IsInformerSynced(gvr) {
 		err := fmt.Errorf("informer cache for %+v is not synced yet", restMapping.Resource)
-		klog.ErrorS(err, "Informer cache is not synced", "gvr", gvr, "placement", placementName)
+		klog.ErrorS(err, "Informer cache is not synced", "gvr", gvr, "placement", placementKey)
 		return nil, controller.NewExpectedBehaviorError(err)
 	}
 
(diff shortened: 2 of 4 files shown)


CLAUDE.md
```diff
@@ -159,11 +159,16 @@ cmd/memberagent/        # Member agent main and setup
...
 ### Unit Tests
-- Use `testify` for assertions
+- Avoid the use of ‘assert’ libraries.
 - Controllers use `envtest` for integration testing with real etcd
 - Mock external dependencies with `gomock`
 - Unit test files: `<go_file>_test.go` in same directory
 - Table-driven test style preferred
+- Use cmp.Equal for equality comparison and cmp.Diff to obtain a human-readable diff between objects.
+- Test outputs should output the actual value that the function returned before printing the value that was expected. A usual format for printing test outputs is “YourFunc(%v) = %v, want %v”.
+- If your function returns a struct, don’t write test code that performs an individual comparison for each field of the struct. Instead, construct the struct that you’re expecting your function to return, and compare in one shot using diffs or deep comparisons. The same rule applies to arrays and maps.
```
jwtty, line 169:
```suggestion
- If your function returns a struct, don’t write test code that performs an individual comparison for each field of the struct. Instead, construct the struct that you’re expecting your function to return, and compare in one shot using diffs or deep comparisons. The same rule applies to arrays and maps. Use the cmp.Diff function.
```

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -128,8 +129,7 @@ func convertResourceSelector(old []fleetv1alpha1.ClusterResourceSelector) []flee
 }
 
 // gatherSelectedResource gets all the resources according to the resource selector.
-// TODO: treat the RP selector differently to not allow RP to select cluster scoped resources
-func (r *Reconciler) gatherSelectedResource(placement string, selectors []fleetv1beta1.ClusterResourceSelector) ([]*unstructured.Unstructured, error) {
+func (r *Reconciler) gatherSelectedResource(placementName types.NamespacedName, selectors []fleetv1beta1.ClusterResourceSelector) ([]*unstructured.Unstructured, error) {
```
jwtty: rename to `placementKey`?

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -145,10 +145,10 @@ func (r *Reconciler) gatherSelectedResource(placement string, selectors []fleetv
 		}
 		var objs []runtime.Object
 		var err error
-		if gvk == utils.NamespaceGVK {
-			objs, err = r.fetchNamespaceResources(selector, placement)
+		if gvk == utils.NamespaceGVK && placementName.Namespace == "" {
+			objs, err = r.fetchNamespaceResources(selector, placementName.Name)
```
jwtty: This PR does not implement namespace only CRP right?

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -230,43 +230,64 @@ func buildApplyOrderMap() map[string]int {
...
 		Version: selector.Version,
 		Kind:    selector.Kind,
 	}
-	if !r.InformerManager.IsClusterScopedResources(gvk) {
-		return nil, controller.NewUserError(fmt.Errorf("invalid placement %s: %+v is not a cluster scoped resource", placeName, restMapping.Resource))
+
+	isNamespacedResource := !r.InformerManager.IsClusterScopedResources(gvk)
+	if isNamespacedResource && placementName.Namespace == "" {
+		// If it's a namespace-scoped resource but placement has no namespace, return error.
+		err := fmt.Errorf("invalid placement %s: cannot select namespace-scoped resource %v without a namespace", placementName, gvr)
```
jwtty:
```suggestion
		err := fmt.Errorf("invalid placement %s: cannot select namespace-scoped resource %v in a clusterResourcePlacement", placementName, gvr)
```

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -230,43 +230,64 @@ func buildApplyOrderMap() map[string]int {
...
+		klog.ErrorS(err, "Invalid resource selector", "selector", selector)
+		return nil, controller.NewUserError(err)
+	} else if !isNamespacedResource && placementName.Namespace != "" {
+		// If it's a cluster-scoped resource but placement has a namespace, return error.
+		err := fmt.Errorf("invalid placement %s: cannot select cluster-scoped resource %v with a namespace", placementName, gvr)
```
jwtty:
```suggestion
		err := fmt.Errorf("invalid placement %s: cannot select cluster-scoped resource %v in a namespaced resourcePlacement", placementName, gvr)
```

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -282,18 +303,26 @@ func (r *Reconciler) fetchClusterScopedResources(selector fleetv1beta1.ClusterRe
...
 	// go ahead and claim all objects by adding a finalizer and insert the placement in its annotation
 	for i := 0; i < len(objects); i++ {
-		uObj := objects[i].DeepCopyObject().(*unstructured.Unstructured)
-		if uObj.GetDeletionTimestamp() != nil {
-			// skip a to be deleted namespace
-			klog.V(2).InfoS("skip the deleting cluster scoped resources by the selector",
-				"selector", selector, "placeName", placeName, "resource name", uObj.GetName())
+		if uObj := objects[i].DeepCopyObject().(*unstructured.Unstructured); uObj.GetDeletionTimestamp() != nil {
+			// skip a to be deleted resource
+			klog.V(2).InfoS("skip the deleting resource by the selector",
```
jwtty:
```suggestion
			klog.V(2).InfoS("Skip the deleting resource by the selector",
```

pkg/controllers/clusterresourceplacement/resource_selector_test.go
```diff
@@ -807,6 +814,886 @@ func createResourceContentForTest(t *testing.T, obj interface{}) *fleetv1beta1.R
 	}
 }
 
+func TestGatherSelectedResource(t *testing.T) {
+	// Common GroupVersionResource definitions used across multiple test cases.
+	deploymentGVR := schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"}
```
jwtty: We have these defined in pkg/utils/common.go

pkg/controllers/clusterresourceplacement/resource_selector_test.go
```diff
@@ -807,6 +814,886 @@ func createResourceContentForTest(t *testing.T, obj interface{}) *fleetv1beta1.R
...
+		Object: map[string]interface{}{
+			"apiVersion": "apps/v1",
+			"kind":       "Deployment",
+			"metadata": map[string]interface{}{
+				"name":      "test-deployment",
+				"namespace": "test-ns",
+			},
+		},
+	}
+	testDeployment.SetGroupVersionKind(schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "Deployment"})
```
jwtty: We also have these GVK defined in pkg/utils/common.go

pkg/controllers/clusterresourceplacement/resource_selector_test.go
```diff
@@ -807,6 +814,886 @@ func createResourceContentForTest(t *testing.T, obj interface{}) *fleetv1beta1.R
...
+			}
+			if diff := cmp.Diff(tt.want, got); diff != "" {
+				t.Errorf("gatherSelectedResource() mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+// fakeRESTMapper is a minimal RESTMapper implementation for testing
+type fakeRESTMapper struct {
```
jwtty, line 1612: Move these to test/utils too?

test/utils/informer/manager.go
```diff
@@ -35,11 +35,24 @@ type FakeLister struct {
 	Err     error
 }
 
-func (f *FakeLister) List(_ labels.Selector) ([]runtime.Object, error) {
+func (f *FakeLister) List(selector labels.Selector) ([]runtime.Object, error) {
```
ryanzhang-oss, line 38: where is this used?

test/utils/informer/manager.go
```diff
@@ -65,19 +78,33 @@ type FakeNamespaceLister struct {
 	Err       error
 }
 
-func (f *FakeNamespaceLister) List(_ labels.Selector) ([]runtime.Object, error) {
+func (f *FakeNamespaceLister) List(selector labels.Selector) ([]runtime.Object, error) {
```
ryanzhang-oss, line 81: same

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -145,10 +145,10 @@ func (r *Reconciler) gatherSelectedResource(placement string, selectors []fleetv
 		}
 		var objs []runtime.Object
 		var err error
-		if gvk == utils.NamespaceGVK {
-			objs, err = r.fetchNamespaceResources(selector, placement)
+		if gvk == utils.NamespaceGVK && placementName.Namespace == "" {
```
ryanzhang-oss: "placementName" should not contain a namespace, maybe PlacmentKey?

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -230,43 +230,64 @@ func buildApplyOrderMap() map[string]int {
 	return ordering
 }
 
-// fetchClusterScopedResources retrieves the objects based on the selector.
-func (r *Reconciler) fetchClusterScopedResources(selector fleetv1beta1.ClusterResourceSelector, placeName string) ([]runtime.Object, error) {
-	klog.V(2).InfoS("start to fetch the cluster scoped resources by the selector", "selector", selector)
+// fetchResources retrieves the objects based on the selector.
+func (r *Reconciler) fetchResources(selector fleetv1beta1.ClusterResourceSelector, placementName types.NamespacedName) ([]runtime.Object, error) {
```
ryanzhang-oss: TODO: I will rename the ClusterResourceSelector as it no longer select clusterResource only

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -330,7 +359,8 @@ func (r *Reconciler) fetchNamespaceResources(selector fleetv1beta1.ClusterResour
 	}
 	namespaces, err := r.InformerManager.Lister(utils.NamespaceGVR).List(labelSelector)
 	if err != nil {
-		return nil, controller.NewAPIServerError(true, fmt.Errorf("cannot list all the namespaces given the label selector: %w", err))
+		klog.ErrorS(err, "Cannot list all the namespaces by the label selector", "labelSelector", labelSelector, "placement", placeName)
+		return nil, controller.NewAPIServerError(true, err)
```
ryanzhang-oss, line 363: NewAPIServerError already does a klog.error, why print it again?

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -384,10 +414,17 @@ func (r *Reconciler) fetchAllResourcesInOneNamespace(namespaceName string, place
 		lister := r.InformerManager.Lister(gvr)
 		objs, err := lister.ByNamespace(namespaceName).List(labels.Everything())
 		if err != nil {
-			return nil, controller.NewAPIServerError(true, fmt.Errorf("cannot list all the objects of type %+v in namespace %s: %w", gvr, namespaceName, err))
+			klog.ErrorS(err, "Cannot list all the objects in namespace", "gvr", gvr, "namespace", namespaceName)
+			return nil, controller.NewAPIServerError(true, err)
```
ryanzhang-oss, line 418: same

CLAUDE.md
zhiying-lin: copied from the golang style guides. whether to use the cmp.Diff depends on how complicated the struct is.

pkg/controllers/clusterresourceplacement/resource_selector.go
zhiying-lin: yeah, will do it in a separate PR

pkg/controllers/clusterresourceplacement/resource_selector_test.go
zhiying-lin: we can do if we find other controllers need this.

test/utils/informer/manager.go
zhiying-lin: it was used in the unit test, TestGatherSelectedResource

pkg/controllers/clusterresourceplacement/resource_selector.go
zhiying-lin: renamed

pkg/controllers/clusterresourceplacement/resource_selector.go
zhiying-lin: the fix is to remove the wrapper of the original api server error. Otherwise, the NewAPIServerError() will always return the unexpected one.

The logs will provide more info, though APIServerError provides the call stack.

jwtty: Not sure if I follow. `controller.NewAPIServerError` checks if the err is unexpected by calling `errors.Is()` and `errors.As()`. These functions consider the wrapped errors too?

pkg/controllers/clusterresourceplacement/resource_selector_test.go
```diff
@@ -807,6 +814,877 @@ func createResourceContentForTest(t *testing.T, obj interface{}) *fleetv1beta1.R
...
+			placementName: types.NamespacedName{Name: "test-placement"},
+			selectors: []fleetv1beta1.ClusterResourceSelector{
+				{
+					Group:   "apps",
+					Version: "v1",
+					Kind:    "Deployment",
+					Name:    "test-deployment",
+				},
+			},
+			resourceConfig: utils.NewResourceConfig(true), // deny list - empty means deny all
```
ryanzhang-oss, line 1015:
```suggestion
			resourceConfig: utils.NewResourceConfig(true), // make this allow list - nothing is allowed
```

pkg/controllers/clusterresourceplacement/resource_selector_test.go
```diff
@@ -807,6 +814,877 @@ func createResourceContentForTest(t *testing.T, obj interface{}) *fleetv1beta1.R
...
+			placementName: types.NamespacedName{Name: "test-placement", Namespace: "test-ns"},
+			selectors: []fleetv1beta1.ClusterResourceSelector{
+				{
+					Group:   "rbac.authorization.k8s.io",
+					Version: "v1",
+					Kind:    "ClusterRole",
+					Name:    "test-clusterrole",
+				},
+			},
+			resourceConfig: utils.NewResourceConfig(false), // allow all resources
```
ryanzhang-oss, line 1029:
```suggestion
			resourceConfig: utils.NewResourceConfig(false)
```

pkg/controllers/clusterresourceplacement/resource_selector_test.go
```diff
@@ -807,6 +814,877 @@ func createResourceContentForTest(t *testing.T, obj interface{}) *fleetv1beta1.R
...
+			placementName: types.NamespacedName{Name: "test-placement", Namespace: "test-ns"},
+			selectors: []fleetv1beta1.ClusterResourceSelector{
+				{
+					Group:   "apps",
+					Version: "v1",
+					Kind:    "Deployment",
+					Name:    "test-deployment",
+				},
+			},
+			resourceConfig: utils.NewResourceConfig(false), // allow all resources
```
ryanzhang-oss, line 1048:
```suggestion
			resourceConfig: utils.NewResourceConfig(false), // default deny list
```

ryanzhang-oss (approved):

pkg/controllers/clusterresourceplacement/resource_selector.go
zhiying-lin: You're right, validated the chain behavior of the as/is func. https://go.dev/play/p/nyuQ1UFtRAG

errors.Is(err, target): This function checks if err or any error in its chain matches target. It recursively unwraps the error chain until a match is found or no more errors can be unwrapped.
errors.As(err, target): This function checks if err or any error in its chain matches the type of target. If a match is found, it assigns the error from the chain to target.

Before i hit a UT failure, i guess it should not be related to the wrapper here.

PR #188, opened 2025-08-13
feat: enable resourcePlacement controllers

jwtty:
Enable resourcePlacement controllers/watchers.

Fixes #

I have:

### DIFF
--- a/cmd/hubagent/options/options.go
+++ b/cmd/hubagent/options/options.go
@@ -98,6 +98,8 @@ type Options struct {
 	EnableStagedUpdateRunAPIs bool
 	// EnableEvictionAPIs enables to agents to watch the eviction and placement disruption budget CRs.
 	EnableEvictionAPIs bool
+	// EnableResourcePlacement enables the agents to watch the ResourcePlacement APIs.
+	EnableResourcePlacement bool
 	// EnablePprof enables the pprof profiling.
 	EnablePprof bool
 	// PprofPort is the port for pprof profiling.
@@ -126,6 +128,7 @@ func NewOptions() *Options {
 		EnableV1Alpha1APIs:                      false,
 		EnableClusterInventoryAPIs:              true,
 		EnableStagedUpdateRunAPIs:               true,
+		EnableResourcePlacement:                 true,
 		EnablePprof:                             false,
 		PprofPort:                               6065,
 		ResourceSnapshotCreationMinimumInterval: 30 * time.Second,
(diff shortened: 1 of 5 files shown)


cmd/hubagent/workload/setup.go
```diff
@@ -96,6 +97,14 @@ var (
...
 	}
 
+	// There's a prerequisite that v1Beta1RequiredGVKs must be installed too.
+	rpRequiredGVKs = []schema.GroupVersionKind{
+		placementv1beta1.GroupVersion.WithKind(placementv1beta1.ResourcePlacementKind),
+		placementv1beta1.GroupVersion.WithKind(placementv1beta1.ResourceBindingKind),
+		placementv1beta1.GroupVersion.WithKind(placementv1beta1.ResourceSnapshotKind),
+		placementv1beta1.GroupVersion.WithKind(placementv1beta1.SchedulingPolicySnapshotKind),
+	}
+
```
ryanzhang-oss, line 107: just double check that those are automatically included in the CRD installer?

pkg/scheduler/watchers/membercluster/watcher.go
```diff
@@ -141,10 +144,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, controller.NewAPIServerError(true, err)
 	}
 	rpList := &placementv1beta1.ResourcePlacementList{}
-	// Empty namespace provided to list RPs across all namespaces.
-	if err := r.Client.List(ctx, rpList, client.InNamespace("")); err != nil {
-		klog.ErrorS(err, "Failed to list RPs", "memberCluster", memberClusterRef)
-		return ctrl.Result{}, controller.NewAPIServerError(true, err)
+	if r.EnableResourcePlacement {
+		// Empty namespace provided to list RPs across all namespaces.
+		if err := r.Client.List(ctx, rpList, client.InNamespace("")); err != nil {
```
ryanzhang-oss, line 149: does this get all the RPs in every namespace or just in "default"

jwtty: In all namespaces, we have integration tests that the RPs are created in different namespace other than default and are selected.

cmd/hubagent/workload/setup.go
jwtty: yes, CRD installer selects CRDs based on the group. Verified in prod that these are installed.

cmd/hubagent/workload/setup.go
```diff
@@ -220,18 +230,66 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
...
+				klog.ErrorS(err, "Unable to set up the resourceBinding watcher")
+				return err
+			}
+
+			klog.Info("Setting up schedulingPolicySnapshot watcher")
+			if err := (&clusterschedulingpolicysnapshot.Reconciler{
+				Client:              mgr.GetClient(),
+				PlacementController: resourcePlacementController,
+			}).SetupWithManagerForSchedulingPolicySnapshot(mgr); err != nil {
+				klog.ErrorS(err, "Unable to set up the schedulingPolicySnapshot watcher")
```
ryanzhang-oss, line 264: since we reuse all those controllers, we probably want to rename them in the next PR.

cmd/hubagent/options/options.go
```diff
@@ -126,6 +128,7 @@ func NewOptions() *Options {
 		EnableV1Alpha1APIs:                      false,
 		EnableClusterInventoryAPIs:              true,
 		EnableStagedUpdateRunAPIs:               true,
+		EnableResourcePlacement:                 true,
```
ryanzhang-oss, line 131: when are we going to remove this? We are adding more and more flags we realistically we won't turn most of them off like the StagedUpdateRun/EnableEvictionAPIs/EnableClusterInventoryAPIs

jwtty: Actually I wonder if we really need to set these here? The `AddFlags` has a default value anyway.

PR #182, opened 2025-08-11
test: add override test in rollout integration

jwtty:
### Description of your changes

Fixes #

I have:

- [ ] Run `make reviewable` to ensure this PR is ready for review.

PR #191, opened 2025-08-14
test: add integration tests for work applier exponential backoff behavior

michaelawyu:

This PR adds additional integration tests for verifying the work applier exponential backoff behavior.

Fixes #

I have:

N/A

N/A

### DIFF
--- a/pkg/controllers/workapplier/controller_integration_test.go
+++ b/pkg/controllers/workapplier/controller_integration_test.go
@@ -7051,7 +7051,7 @@ var _ = Describe("negative cases", func() {
 			malformedConfigMapJSON := marshalK8sObjJSON(malformedConfigMap)
 
 			// Create a Work object with all the manifest JSONs.
-			createWorkObject(workName, nil, regularNSJSON, malformedConfigMapJSON, regularConfigMapJson)
+			createWorkObject(workName, memberReservedNSName1, nil, regularNSJSON, malformedConfigMapJSON, regularConfigMapJson)
 		})
 
 		It("should add cleanup finalizer to the Work object", func() {
@@ -7071,13 +7071,13 @@ var _ = Describe("negative cases", func() {
 			regularNSObjectAppliedActual := regularNSObjectAppliedActual(nsName, appliedWorkOwnerRef)
 			Eventually(regularNSObjectAppliedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to apply the namespace object")
 
-			Expect(memberClient.Get(ctx, client.ObjectKey{Name: nsName}, regularNS)).To(Succeed(), "Failed to retrieve the NS object")
+			Expect(memberClient1.Get(ctx, client.ObjectKey{Name: nsName}, regularNS)).To(Succeed(), "Failed to retrieve the NS object")
 
 			// Ensure that the ConfigMap object has been applied as expected.
 			regularConfigMapAppliedActual := regularConfigMapObjectAppliedActual(nsName, configMapName, appliedWorkOwnerRef)
 			Eventually(regularConfigMapAppliedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to apply the ConfigMap object")
 
-			Expect(memberClient.Get(ctx, client.ObjectKey{Namespace: nsName, Name: configMapName}, regularConfigMap)).To(Succeed(), "Failed to retrieve the ConfigMap object")
+			Expect(memberClient1.Get(ctx, client.ObjectKey{Namespace: nsName, Name: configMapName}, regularConfigMap)).To(Succeed(), "Failed to retrieve the ConfigMap object")
 		})
 
 		It("should update the Work object status", func() {
@@ -7198,7 +7198,7 @@ var _ = Describe("negative cases", func() {
 
 		AfterAll(func() {
 			// Delete the Work object and related resources.
-			deleteWorkObject(workName)
+			deleteWorkObject(workName, memberReservedNSName1)
 
 			// Ensure applied manifest has been removed.
 			regularConfigMapRemovedActual := regularConfigMapRemovedActual(nsName, configMapName)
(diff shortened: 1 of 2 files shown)


pkg/controllers/workapplier/backoff_integration_test.go
```diff
@@ -0,0 +1,524 @@
...
+					ignoreDiffDetailsObsTime, ignoreDriftDetailsObsTime,
+					cmpopts.SortSlices(lessFuncPatchDetail),
+				); diff != "" {
+					return fmt.Errorf("work status diff (-got, +want):\n%s", diff)
+				}
+
+				// Track the observation timestamp of the diff details.
+				lastDiffObservedTime = &work.Status.ManifestConditions[0].DiffDetails.ObservationTime
+				return nil
+			}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update Work object status")
```
ryanzhang-oss, line 210: eventuallyInterval is 1 second, I wonder if it will make it flaky?

pkg/controllers/workapplier/backoff_integration_test.go
```diff
@@ -0,0 +1,524 @@
...
+		})
+
+		It("should wait for the fixed delay period of time before next reconciliation", func() {
+			curDiffObservedTime := &metav1.Time{}
+
+			// Need to poll a bit longer to avoid flakiness.
+			Eventually(diffObservationTimeChangedActual(workName, wantWorkStatus, curDiffObservedTime, lastDiffObservedTime), eventuallyDuration*2, eventuallyInterval).Should(Or(Succeed()))
+
+			// Fixed delay is set to 10 seconds. Give a one-sec leeway to avoid flakiness.
```
ryanzhang-oss, line 219: I thought it was 5 sec? Where is 20 second set?

pkg/controllers/workapplier/backoff_integration_test.go
michaelawyu: Hi Ryan! These tests run with a special exponential backoff setup (10 sec delay first, then 20 secs with 1.5x slow backoff, and eventually 30 seconds with 2x fast backoff). The results are expected to stay the same for at least 10 seconds after first apply, so it should be alright.

pkg/controllers/workapplier/backoff_integration_test.go
michaelawyu: Hi Ryan! Yeah, for the actual setup the default value is 5 seconds; this one has to use a longer one as the default values (not the initial 5 second part, but the slow/fast backoffs do not play well with our integration tests -> can be flaky sometimes).

pkg/controllers/workapplier/suite_test.go
```diff
@@ -128,60 +166,135 @@ var _ = BeforeSuite(func() {
...
-				memberReservedNSName: {},
+				memberReservedNSName2: {},
 			},
 		},
 		Logger: textlogger.NewLogger(textlogger.NewConfig(textlogger.Verbosity(4))),
 	})
 	Expect(err).ToNot(HaveOccurred())
 
-	workApplier = NewReconciler(
+	superLongExponentialBackoffRateLimiter := NewRequeueMultiStageWithExponentialBackoffRateLimiter(
```
michaelawyu, line 233: The test specific setup is added here; PTAL.

pkg/controllers/workapplier/backoff_integration_test.go
michaelawyu: Hi Ryan! I've pinpointed the test-specific setup in the comment down below.

michaelawyu:
Resolved the conflict.

Makefile
```diff
@@ -137,6 +137,8 @@ test: manifests generate fmt vet local-unit-test integration-test## Run tests.
 ## workaround to bypass the pkg/controllers/workv1alpha1 tests failure
 ## rollout controller tests need a bit longer to complete, so we increase the timeout
 ##
+# Set up the timeout parameters as some of the test lengths have exceeded the default 10 minute mark.
```
ryanzhang-oss, line 140: where is the timeout set in this PR?

pkg/controllers/workapplier/backoff_integration_test.go
```diff
@@ -0,0 +1,524 @@
...
+			if diff := cmp.Diff(
+				&work.Status, wantWorkStatus,
+				ignoreFieldConditionLTTMsg,
+				ignoreDiffDetailsObsTime, ignoreDriftDetailsObsTime,
+				cmpopts.SortSlices(lessFuncPatchDetail),
+			); diff != "" {
+				StopTrying("the work object status has changed unexpectedly").Wrap(fmt.Errorf("work status diff (-got, +want):\n%s", diff)).Now()
+			}
+
+			curDriftObservedTime.Time = work.Status.ManifestConditions[0].DriftDetails.ObservationTime.Time
```
ryanzhang-oss, line 96: Why pass in the `curDriftObservedTime` when its value is set inside the function. Looks like this is more of an "out" parameter  or return value?

Makefile
michaelawyu: Hi Ryan! A previous PR has already set the timeout to 20 mins, so the changes have been absolved.

michaelawyu: The default is 10 minutes, #182 has set it to 20 minutes and got merged before this PR

pkg/controllers/workapplier/backoff_integration_test.go
michaelawyu: Hi Ryan! Some of test steps later verify the currently observed timestamp:

```
It("should wait for the fixed delay period of time before next reconciliation", func() {
			curDiffObservedTime := &metav1.Time{}

			// Need to poll a bit longer to avoid flakiness.
			Eventually(diffObservationTimeChangedActual(workName, wantWorkStatus, curDiffObservedTime, lastDiffObservedTime), eventuallyDuration*2, eventuallyInterval).Should(Or(Succeed()))

			// Fixed delay is set to 10 seconds. Give a one-sec leeway to avoid flakiness.
			Expect(curDiffObservedTime.Sub(lastDiffObservedTime.Time)).To(And(
				BeNumerically(">=", time.Second*9),
				BeNumerically("<=", time.Second*11),
			), "the interval between two observations is not as expected")

			// Update the tracked observation time.
			lastDiffObservedTime = curDiffObservedTime
		})
```

You are right that it works more like a returned result rather than an passed-in arguement, but the Ginkgo Eventually  blocks make it a bit difficult to format that way 😞

michaelawyu:
Merged the PR to unblock progress -> can open new PRs if changes are needed.

PR #192, opened 2025-08-14
docs: add rp examples

jwtty:
Example resources to test RP behavior.

Fixes #

I have:

### DIFF
--- a/examples/resourceplacement/test-crp.yaml
+++ b/examples/resourceplacement/test-crp.yaml
@@ -10,7 +10,7 @@ spec:
       kind: Namespace
       name: test-ns
       version: v1
-      selectionScope: NamespaceOnly
+      selectionScope: NamespaceOnly # only namespace itself is placed, no resources within the namespace
   policy:
     placementType: PickAll
   strategy:


examples/resourceplacement/test-crp.yaml
```diff
@@ -0,0 +1,20 @@
+# This tests a CRP selecting a namespace only.
```
britaniar, line 1: Nit: Maybe add a note on the difference between a regular CRP with the namespace only in the resource selector. Something like "NOTE:  With `SelectionScope` as `NamespaceOnly`, only the namespace will be propagated."

jwtty: I added the comment next to the field.

examples/resourceplacement/rp-deploy.yaml
```diff
@@ -0,0 +1,26 @@
+# This tests selecting multiple resources in a namespace,
+# and only applying to a subset of clusters.
+# Prerequisite: create and expose a deployment named "test-nginx" in namespace "test-ns".
```
ryanzhang-oss, line 3: better to have those yaml in the directory too so it's self contained

jwtty: will add in a separate PR.

PR #195, opened 2025-08-16
feat: enable RP metrics

jwtty:
1. Move v1beta1 placement metrics from pkg/utils/controller/metrics/metrics.go to pkg/metrics/metrics.go
2. Add namespace label in placememt status gauge metric.

Fixes #

I have:

- [ x Run `make reviewable` to ensure this PR is ready for review.

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller_integration_test.go
+++ b/pkg/controllers/clusterresourceplacement/controller_integration_test.go
@@ -35,9 +35,9 @@ import (
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 
 	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
+	fleetmetrics "github.com/kubefleet-dev/kubefleet/pkg/metrics"
 	"github.com/kubefleet-dev/kubefleet/pkg/utils"
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
-	"github.com/kubefleet-dev/kubefleet/pkg/utils/controller/metrics"
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/resource"
 	metricsUtils "github.com/kubefleet-dev/kubefleet/test/utils/metrics"
 )
@@ -380,7 +380,7 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
 	Context("When creating new pickAll ClusterResourcePlacement", func() {
 		BeforeEach(func() {
 			// Reset metric before each test
-			metrics.FleetPlacementStatusLastTimeStampSeconds.Reset()
+			fleetmetrics.FleetPlacementStatusLastTimeStampSeconds.Reset()
 
 			By("Create a new crp")
 			crp = &placementv1beta1.ClusterResourcePlacement{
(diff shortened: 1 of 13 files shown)


pkg/metrics/metrics.go
```diff
@@ -43,6 +43,26 @@ var (
 		Name: "placement_apply_succeed_counter",
 		Help: "Number of successfully applied cluster resource placement",
 	}, []string{"name"})
```
britaniar, line 45: Looks like the metrics above may be old. Do we still need them?

pkg/metrics/metrics.go
```diff
@@ -43,6 +43,26 @@ var (
...
+
+	// FleetPlacementStatusLastTimeStampSeconds is a prometheus metric which keeps track of the last placement status.
+	FleetPlacementStatusLastTimeStampSeconds = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "fleet_workload_placement_status_last_timestamp_seconds",
+		Help: "Timestamp in seconds of the last current placement status condition of crp.",
```
britaniar: nit: remove CRP from the description

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -35,9 +35,9 @@ import (
 	ctrlmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 
 	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
+	fleetmetrics "github.com/kubefleet-dev/kubefleet/pkg/metrics"
```
britaniar: does it need to be fleetmetrics? Can it be just metrics like it was previously, since the other two are renamed already?

pkg/metrics/metrics.go
jwtty: `WorkApplyTime`, `PlacementApplyFailedCount` and `PlacementApplySucceedCount` are still used by v1alpha1 controllers. We should deprecate them together with v1alpha1 deprecation. `JoinResultMetrics` and `LeaveResultMetrics` are used in both v1alpha1 and v1beta1 member cluster controllers.

pkg/metrics/metrics.go
jwtty: fixed

pkg/controllers/clusterresourceplacement/controller_integration_test.go
jwtty: renamed

pkg/controllers/updaterun/suite_test.go
```diff
@@ -110,13 +112,16 @@ var _ = BeforeSuite(func() {
...
-	// setup our main reconciler
+	// Setup our main reconciler.
 	err = (&Reconciler{
 		Client:          k8sClient,
 		InformerManager: dynamicInformerManager,
 	}).SetupWithManager(mgr)
 	Expect(err).Should(Succeed())
 
+	// Register metrics.
+	ctrlmetrics.Registry.MustRegister(metrics.FleetUpdateRunStatusLastTimestampSeconds)
```
ryanzhang-oss, line 123: how did it work before without it?

britaniar: Here we are registering on the hub-agent instead of the controller (https://github.com/kubefleet-dev/kubefleet/blob/d8d26187bc186be74f47372d56929b4d54c29947/pkg/utils/controller/metrics/metrics.go#L65). The integration tests didn't need it before because it was registered on the controller itself. Where as they need it now because it is in the hub-agent and the integration test doesn't run the hub-agent.

ryanzhang-oss: shouldn't we do the init function in the new metrics location too?

PR #204, opened 2025-08-19
fix: update scheduler watchers package name

britaniar:

Fixes #

I have: update the package name to be inclusive by removing cluster from the front


cmd/hubagent/workload/setup.go
```diff
@@ -371,7 +371,7 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 
 		// Set up the watchers for the controller
 		klog.Info("Setting up the clusterResourcePlacement watcher for scheduler")
-		if err := (&schedulercrpwatcher.Reconciler{
+		if err := (&schedulerplacementwatcher.Reconciler{
```
jwtty, line 374: Are we going to change the watcher names for placement controller too? like line 260?

britaniar: Yes, opening a separate PR for that. Splitting so that it is not a lot changes at once. Have the PR opened for that https://github.com/kubefleet-dev/kubefleet/pull/206. This is just specifically for the scheduler watchers.

PR #207, opened 2025-08-20
feat: make StatusReportingScope immutable

Arvindthiru:

Fixes #

I have:

ITs for CEL validation

The // +kubebuilder:default=ClusterScopeOnly marker sets StatusReportingScope to ClusterScopeOnly,

- if the field is ommitted
- If the field is set to an empty string

I tried adding another CEL validation to ensure users can't input empty string to StatusReportingScope but internally it's set to ClusterScopeOnly due to the default marker making the CEL validation check for empty string pointless

Current implementation achieves making StatusReportingScope immutable but users can supply empty string as input but internally it gets converted to ClusterScopeOnly.

Considering we can't drop the default marker I'm proceeding with this approach

### DIFF
--- a/test/apis/placement/v1beta1/api_validation_integration_test.go
+++ b/test/apis/placement/v1beta1/api_validation_integration_test.go
@@ -27,15 +27,9 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
-)
-
-var (
-	eventuallyTimeout = time.Second * 5
-	interval          = time.Millisecond * 250
 )
 
 const (
@@ -87,7 +81,6 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 		})
 
 		It("should deny update of ClusterResourcePlacement with nil policy", func() {
-			Expect(hubClient.Get(ctx, types.NamespacedName{Name: crpName}, &crp)).Should(Succeed(), "Get CRP call failed")
 			crp.Spec.Policy = nil
 			err := hubClient.Update(ctx, &crp)
 			var statusErr *k8sErrors.StatusError
@@ -96,7 +89,6 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 		})
 
 		It("should deny update of ClusterResourcePlacement with different placement type", func() {
-			Expect(hubClient.Get(ctx, types.NamespacedName{Name: crpName}, &crp)).Should(Succeed(), "Get CRP call failed")
 			crp.Spec.Policy.PlacementType = placementv1beta1.PickAllPlacementType
 			err := hubClient.Update(ctx, &crp)
 			var statusErr *k8sErrors.StatusError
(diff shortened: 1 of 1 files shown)


test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -331,121 +398,291 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
...
+
+		It("should deny update of ClusterResourcePlacement StatusReportingScope to NamespaceAccessible due to immutability", func() {
+			Eventually(func(g Gomega) error {
+				err := hubClient.Get(ctx, types.NamespacedName{Name: crpName}, &crp)
+				if err != nil {
+					return err
+				}
+				crp.Spec.StatusReportingScope = placementv1beta1.NamespaceAccessible
+				err = hubClient.Update(ctx, &crp)
+				if k8sErrors.IsConflict(err) {
```
britaniar: why is this needed?

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -331,121 +398,291 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
...
+					return err
+				}
+				err = hubClient.Get(ctx, types.NamespacedName{Name: crpName}, &crp)
+				if err != nil {
+					return err
+				}
+				g.Expect(crp.Spec.StatusReportingScope).To(Equal(placementv1beta1.ClusterScopeOnly), "CRP should have default StatusReportingScope ClusterScopeOnly")
+				return nil
+			}, eventuallyTimeout, interval).Should(Succeed(), "Failed to deny update on CRP")
```
britaniar:
```suggestion
			}, eventuallyTimeout, interval).Should(Succeed(), "Failed to update on CRP")

File: test/apis/placement/v1beta1/api_validation_integration_test.go [hunk 3]
```diff
@@ -331,121 +398,291 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
...
+					},
+					{
+						Group:   "",
+						Version: "v1",
+						Kind:    "PersistentVolume",
+						Name:    "test-pv",
+					},
+				}...)
+				return hubClient.Update(ctx, &crp)
+			}, eventuallyTimeout, interval).Should(Succeed(), "Failed to deny update on CRP")
```
britaniar: ditto

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -331,121 +398,291 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
...
+					{
+						Group:   "",
+						Version: "v1",
+						Kind:    "PersistentVolume",
+						Name:    "test-pv",
+					},
+				}...)
+				crp.Spec.StatusReportingScope = placementv1beta1.ClusterScopeOnly
+				return hubClient.Update(ctx, &crp)
+			}, eventuallyTimeout, interval).Should(Succeed(), "Failed to deny update on CRP")
```
britaniar: ditto

test/apis/placement/v1beta1/api_validation_integration_test.go
Arvindthiru: Ideally we won't run it conflicts, but it's just a safety measure to ensure if we do we throw error and retry

britaniar (dismissed):

examples/test-crp1.yaml
```diff
@@ -11,4 +11,4 @@ spec:
   policy:
     placementType: PickN
   strategy:
-    type: RollingUpdate
\ No newline at end of file
+    type: RollingUpdate
```
ryanzhang-oss: what's the change here?

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -33,6 +33,11 @@ import (
 	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
 )
 
+var (
+	eventuallyTimeout = time.Second * 5
+	interval          = time.Millisecond * 250
+)
+
```
ryanzhang-oss: why adding those here, we have general e2es. Is this different?

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -331,121 +398,291 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 			Expect(hubClient.Delete(ctx, &crp)).Should(Succeed())
 		})
 
-		It("should allow update of ClusterResourcePlacement with StatusReportingScope NamespaceAccessible, one namespace selector", func() {
-			crp.Spec.StatusReportingScope = placementv1beta1.NamespaceAccessible
-			Expect(hubClient.Update(ctx, &crp)).Should(Succeed())
+		It("should allow update of ClusterResourcePlacement with empty string for StatusReportingScope", func() {
```
ryanzhang-oss:
```suggestion
		It("should allow add empty string for StatusReportingScope in a ClusterResourcePlacement without one set", func() {
```

ryanzhang-oss: as you can't replace something with an empty string

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -331,121 +398,291 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
...
+				err = hubClient.Get(ctx, types.NamespacedName{Name: crpName}, &crp)
+				if err != nil {
+					return err
+				}
+				g.Expect(crp.Spec.StatusReportingScope).To(Equal(placementv1beta1.ClusterScopeOnly), "CRP should have default StatusReportingScope ClusterScopeOnly")
+				return nil
+			}, eventuallyTimeout, interval).Should(Succeed(), "Failed to update CRP")
+		})
+
+		It("should allow update of ClusterResourcePlacement with default StatusReportingScope, multiple namespace resource selectors", func() {
```
ryanzhang-oss:
```suggestion
		It("should allow update of ClusterResourcePlacement which has default StatusReportingScope, multiple namespace resource selectors", func() {
```

test/apis/placement/v1beta1/api_validation_integration_test.go
ryanzhang-oss: why we only have this here in this test?

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -331,121 +398,291 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
...
+						Kind:    "Namespace",
+						Name:    "test-ns-2",
+					},
+					{
+						Group:   "rbac.authorization.k8s.io",
+						Version: "v1",
+						Kind:    "ClusterRole",
+						Name:    "test-cluster-role",
+					},
+				}...)
```
ryanzhang-oss: when you pt them together, how do I know which one caused the deny?

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -33,6 +33,11 @@ import (
 	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
 )
 
+var (
```
ryanzhang-oss: can we add ResourcePlacement test too as the PlacementSpec is shared with RP

test/apis/placement/v1beta1/api_validation_integration_test.go
Arvindthiru: There are needed for the eventually blocks in this file

test/apis/placement/v1beta1/api_validation_integration_test.go
Arvindthiru: We do have conflict checks in other specs in this file

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -331,121 +394,419 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 			Expect(hubClient.Delete(ctx, &crp)).Should(Succeed())
 		})
 
-		It("should allow update of ClusterResourcePlacement with StatusReportingScope NamespaceAccessible, one namespace selector", func() {
-			crp.Spec.StatusReportingScope = placementv1beta1.NamespaceAccessible
-			Expect(hubClient.Update(ctx, &crp)).Should(Succeed())
+		It("should allow empty string for StatusReportingScope in a ClusterResourcePlacement when StatusReportingScope is not set", func() {
+			Eventually(func(g Gomega) error {
```
jwtty: You don't need to get the crp again as it will be returned in the `create` function. Then you don't need the `Eventually` anymore.

test/apis/placement/v1beta1/api_validation_integration_test.go
jwtty: May I ask which test has IsConflict check?

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -412,14 +489,7 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 					Kind:    "Namespace",
 					Name:    "test-ns-2",
 				},
-				{
```
jwtty, line 415: I have a question: do we allow switching from one namespace to another?

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -441,12 +511,220 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
...
+							Name:    "test-secret",
+						},
+					},
+					StatusReportingScope: placementv1beta1.NamespaceAccessible,
+				},
+			}
+			Expect(hubClient.Create(ctx, &rp)).Should(Succeed())
+		})
+
+		It("should allow creation of ResourcePlacement with StatusReportingScope ClusterScopeOnly", func() {
```
jwtty, line 580: This should have no effect for RP right?

test/apis/placement/v1beta1/api_validation_integration_test.go
Arvindthiru: We won't allow it

jwtty: Per offline discussion, we either need to add webhook to block changing the namespace, or we need to make sure the old CRPS in the original namespace is cleaned up after switching.

jwtty (approved):

PR #209, opened 2025-08-20
feat: add support for degraded diff drift calculation

michaelawyu:

KubeFleet currently uses server-side apply in the dry-run mode to calculate diffs/drifts in the partial comparison mode. It has been identified that there are cases where server-side apply dry-runs might fail when the supplied manifest itself is not valid (e.g., when mutable scheduling directives are used). This PR addresses this issue by enabling degraded diff/drift calculations -> the work applier will fall back to the full comparison mode if it is sure that server-side apply dry-runs will never succeed given the current manifest.

Fixes #

I have:

- [N/A] Unit tests
- [x] Integration tests

Additional tests will be submitted in separate PRs to control the PR size.

### DIFF
--- a/apis/cluster/v1beta1/zz_generated.deepcopy.go
+++ b/apis/cluster/v1beta1/zz_generated.deepcopy.go
@@ -21,7 +21,7 @@ limitations under the License.
 package v1beta1
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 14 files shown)


michaelawyu:
Note: partial comparison mode requires server-side apply to function; unfortunately the mock clients we use in our unit tests do not support server-side apply ops, and as a result all partial comparison mode related code are only tested via integration tests, not unit tests.

PR #217, opened 2025-08-21
test: enable rollout e2e tests for RP

jwtty:
Enable RP e2e for rollout scenarios.

Fixes #

I have:

### DIFF
--- a/test/e2e/resource_placement_rollout_test.go
+++ b/test/e2e/resource_placement_rollout_test.go
@@ -1,0 +1,1193 @@
+/*
+Copyright 2025 The KubeFleet Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


test/e2e/resource_placement_rollout_test.go
```diff
@@ -0,0 +1,1193 @@
...
+			Eventually(func() error {
+				rp := &placementv1beta1.ResourcePlacement{}
+				if err := hubClient.Get(ctx, rpKey, rp); err != nil {
+					return fmt.Errorf("failed to get RP: %w", err)
+				}
+
+				if diff := cmp.Diff(rp.Status.SelectedResources, wantSelectedResources, cmpopts.SortSlices(utils.LessFuncResourceIdentifier)); diff != "" {
+					return fmt.Errorf("selected resources mismatched (-got, +want): %s", diff)
+				}
+				// Use the fresh observed resource index to verify the RP status later.
```
zhiying-lin, line 1080: just wondering why the observedResourceIndex is not predictable, should it be 0?

jwtty: Since this is a custom resource, and our resource discovery has a 15s period, there's chance that even the resources are already created at the beginning of the test, the custom resource has not been detected at the time when rp gets created. So it may take 2 resourceSnapshots: the 1st one is empty without the custom resource, then 2nd one has it.

jwtty: I'm not sure why the CRP rollout test does not need this in this test, probably because this is the 2nd test and the resource gets populated quickly. I'm adding this to avoid flakiness.

zhiying-lin (approved):

PR #219, opened 2025-08-22
test: add RP E2Es for diff/drift

britaniar:

Fixes #

I have:

### DIFF
--- a/test/e2e/resource_placement_drift_diff_test.go
+++ b/test/e2e/resource_placement_drift_diff_test.go
@@ -36,62 +36,82 @@ import (
 )
 
 var _ = Describe("take over existing resources using RP", Label("resourceplacement"), func() {
+	var crpName, nsName, cm1Name, cm2Name string
+
+	BeforeEach(OncePerOrdered, func() {
+		nsName = fmt.Sprintf(workNamespaceNameTemplate, GinkgoParallelProcess())
+		cm1Name = fmt.Sprintf(appConfigMapNameTemplate, GinkgoParallelProcess())
+		cm2Name = fmt.Sprintf(appConfigMapNameTemplate+"-%d", GinkgoParallelProcess(), 2)
+
+		By("Create the resources on the hub cluster")
+		createWorkResources()
+		cm2 := appConfigMap()
+		cm2.Name = cm2Name
+		Expect(hubClient.Create(ctx, &cm2)).To(Succeed())
+
+		By("Create the resources on one of the member clusters")
+		ns := appNamespace()
+		Expect(memberCluster1EastProdClient.Create(ctx, &ns)).To(Succeed())
+
+		cm1 := appConfigMap()
+		// Update the configMap data (managed field).
+		cm1.Data = map[string]string{
+			managedDataFieldKey: managedDataFieldVal1,
+		}
+		Expect(memberCluster1EastProdClient.Create(ctx, &cm1)).To(Succeed())
+
+		// Update the configMap labels (unmanaged field).
+		cm2 = appConfigMap()
+		cm2.Name = cm2Name
+		cm2.Labels = map[string]string{
+			unmanagedLabelKey: unmanagedLabelVal1,
+		}
+		Expect(memberCluster1EastProdClient.Create(ctx, &cm2)).To(Succeed())
+
+		By("Create the CRP with Namespace-only selector")
+		crpName = fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+		crp := &placementv1beta1.ClusterResourcePlacement{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: crpName,
+				// Add a custom finalizer; this would allow us to better observe
+				// the behavior of the controllers.
+				Finalizers: []string{customDeletionBlockerFinalizer},
+			},
+			Spec: placementv1beta1.PlacementSpec{
+				ResourceSelectors: namespaceOnlySelector(),
+				Policy: &placementv1beta1.PlacementPolicy{
+					PlacementType: placementv1beta1.PickAllPlacementType,
+				},
+				Strategy: placementv1beta1.RolloutStrategy{
+					Type: placementv1beta1.RollingUpdateRolloutStrategyType,
+					RollingUpdate: &placementv1beta1.RollingUpdateConfig{
+						UnavailablePeriodSeconds: ptr.To(2),
+					},
+				},
+			},
+		}
+		Expect(hubClient.Create(ctx, crp)).To(Succeed(), "Failed to create CRP")
+
+		By("Validate CRP status is as expected")
+		crpStatusUpdatedActual := crpStatusUpdatedActual(workNamespaceIdentifiers(), allMemberClusterNames, nil, "0")
+		Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP status as expected")
+	})
+
+	AfterEach(OncePerOrdered, func() {
+		// The pre-existing resource that have not been taken over and must be deleted manually.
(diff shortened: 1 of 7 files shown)


test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -0,0 +1,3045 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
+	"github.com/kubefleet-dev/kubefleet/test/e2e/framework"
+)
+
+var _ = Describe("take over existing resources using RP", Label("resourceplacement"), func() {
```
jwtty, line 38: So the original design of the drift diff e2e for CRP is that we can test each combination of ApplyStrategy on both `unmanaged` field (tested by the namespace) and `managed` field (tested by the configMap). Right now in our RP test, the unmanaged scenario is stilled tested by CRP but RP only tests the managed field. For each of the test cases, could you please either add another configMap with an unmanaged field change in the placed resources or just add another test case that tests configMap + unmanaged field?

test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -0,0 +1,3045 @@
...
+				if err := hubClient.Get(ctx, types.NamespacedName{Name: crpName}, crp); err != nil {
+					return err
+				}
+				wantCRPStatus := buildWantCRPStatus(crp.Generation)
+
+				if diff := cmp.Diff(crp.Status, *wantCRPStatus, placementStatusCmpOptions...); diff != "" {
+					return fmt.Errorf("CRP status diff (-got, +want): %s", diff)
+				}
+				return nil
+				// Use a longer eventually timeout, as the Fleet is set to detect drifts
```
michaelawyu: Hi Britania! Sorry for jumping in and adding a comment, but after we switch to exponential backoff this should no longer apply -> we could still use the 30 seconds delay (or switch to `longEventuallyDuration` for simplicity), as it's a bit difficult for us to determine how much the current delay is right now (though it should still be small at this stage), but the comment here is no longer applicable I am afraid.

michaelawyu: The same applies to other occurrences as well. Feel free to just drop the comment -> I will see if there's a slightly more sophisticated way to handle this

test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -0,0 +1,2467 @@
...
+
+		It("should place the resources on the rest of the member clusters", func() {
+			targetClusters := []*framework.Cluster{
+				memberCluster2EastCanary,
+				memberCluster3WestProd,
+			}
+			for idx := range targetClusters {
+				memberCluster := targetClusters[idx]
+
+				workResourcesPlacedActual := workNamespaceAndConfigMapPlacedOnClusterActual(memberCluster)
```
jwtty, line 194: This function only checks the appConfigMap(), we should check the other one too. ditto to other tests.

test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -0,0 +1,2467 @@
...
+					},
+					Strategy: placementv1beta1.RolloutStrategy{
+						Type: placementv1beta1.RollingUpdateRolloutStrategyType,
+						RollingUpdate: &placementv1beta1.RollingUpdateConfig{
+							UnavailablePeriodSeconds: ptr.To(2),
+						},
+					},
+				},
+			}
+			Expect(hubClient.Create(ctx, crp)).To(Succeed(), "Failed to create CRP")
```
jwtty: Can we move the common resource creation and crp creation and validation part out to the `BeforeEach(OncePerOrdered...` block as the other tests? ditto to other tests.

test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -0,0 +1,2467 @@
...
+				return nil
+			}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update RP status as expected")
+		})
+
+		It("should take over existing resources with no diff on clusters", func() {
+			cm1 := &corev1.ConfigMap{}
+			Expect(memberCluster1EastProdClient.Get(ctx, client.ObjectKey{Name: cm1Name, Namespace: nsName}, cm1)).To(Succeed())
+
+			// The difference has been overwritten.
+			wantCM := appConfigMap()
```
jwtty, line 319: This part is for "should not take over existing resources with diff on clusters"?

test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -0,0 +1,2467 @@
...
+				memberCluster := targetClusters[idx]
+
+				workResourcesPlacedActual := workNamespaceAndConfigMapPlacedOnClusterActual(memberCluster)
+				Eventually(workResourcesPlacedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place work resources on member cluster %s", memberCluster.ClusterName)
+			}
+		})
+
+		AfterAll(func() {
+			// The pre-existing namespace has not been taken over and must be deleted manually.
+			cleanWorkResourcesOnCluster(memberCluster1EastProd)
```
jwtty: Does this function remove both configMaps?

test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -0,0 +1,2467 @@
...
+			for idx := range targetClusters {
+				memberCluster := targetClusters[idx]
+
+				workResourcesPlacedActual := workNamespaceAndConfigMapPlacedOnClusterActual(memberCluster)
+				Eventually(workResourcesPlacedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place work resources on member cluster %s", memberCluster.ClusterName)
+			}
+		})
+
+		AfterAll(func() {
+			ensureRPAndRelatedResourcesDeleted(types.NamespacedName{Name: rpName, Namespace: nsName}, allMemberClusters)
```
jwtty, line 736: You need to make sure the other configMap is deleted too. You can simply append that configMap in this function. Please check the signature. ditto to other functions too.

jwtty (commented):
Thanks for making the change!

test/e2e/utils_test.go
```diff
@@ -1612,6 +1640,24 @@ func ensureRPAndRelatedResourcesDeleted(rpKey types.NamespacedName, memberCluste
 	cleanupConfigMap()
 }
 
+func ensureRPCleanUp(rpKey types.NamespacedName) {
```
jwtty: Can we use `ensureRPAndRelatedResourcesDeleted` directly?

test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -0,0 +1,2342 @@
...
+		cm1Name = fmt.Sprintf(appConfigMapNameTemplate, GinkgoParallelProcess())
+		cm2Name = fmt.Sprintf(appConfigMapNameTemplate+"-%d", GinkgoParallelProcess(), 2)
+
+		By("Create the resources on the hub cluster")
+		createWorkResources()
+		cm2 := appConfigMap()
+		cm2.Name = cm2Name
+		Expect(hubClient.Create(ctx, &cm2)).To(Succeed())
+
+		By("Create the resources on one of the member clusters")
```
jwtty: No need to create resources on the member cluster for this test?

test/e2e/utils_test.go
britaniar: If I use it, then it deletes the configmap on the hub cluster and I will need to recreate it every time.

PR #220, opened 2025-08-22
fix: fix the resource change bug

zhiying-lin:

only trigger the ns scoped resource when the CRP is selecting it in the status or the spec selects the resource.

Fixes #

I have:

### DIFF
--- a/pkg/controllers/resourcechange/resourcechange_controller.go
+++ b/pkg/controllers/resourcechange/resourcechange_controller.go
@@ -436,7 +436,7 @@ func collectAllAffectedPlacementsV1Beta1(key keys.ClusterWideKey, res *unstructu
 			// For the clusterResourcePlacement, we skip the namespace scoped resources if the placement is cluster scoped.
 			if key.Namespace != "" && isSelectNamespaceOnly(selector) && placement.GetNamespace() == "" {
 				// If the selector is namespace only, we skip the namespace scoped resources.
-				klog.V(2).InfoS("Skipping namespace scoped resource for namespace only selector", "obj", klog.KRef(res.GetNamespace(), res.GetName()), "selector", selector, "placement", klog.KObj(placement))
+				klog.V(2).InfoS("Skipping namespace scoped resource for namespace only selector", "key", key, "obj", klog.KRef(res.GetNamespace(), res.GetName()), "selector", selector, "placement", klog.KObj(placement))
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/resourcechange/resourcechange_controller.go
```diff
@@ -433,7 +434,7 @@ func collectAllAffectedPlacementsV1Beta1(isClusterScoped bool, res *unstructured
 		// will validate the resource placement's namespace matches the resource's namespace.
 		for _, selector := range placement.GetPlacementSpec().ResourceSelectors {
 			// For the clusterResourcePlacement, we skip the namespace scoped resources if the placement is cluster scoped.
-			if !isClusterScoped && isSelectNamespaceOnly(selector) && placement.GetNamespace() == "" {
+			if key.Namespace != "" && isSelectNamespaceOnly(selector) && placement.GetNamespace() == "" {
 				// If the selector is namespace only, we skip the namespace scoped resources.
 				klog.V(2).InfoS("Skipping namespace scoped resource for namespace only selector", "obj", klog.KRef(res.GetNamespace(), res.GetName()), "selector", selector, "placement", klog.KObj(placement))
```
jwtty: This line of log should refer the key.

pkg/controllers/resourcechange/resourcechange_controller_test.go
```diff
@@ -671,20 +671,41 @@ func TestCollectAllAffectedPlacementsV1Beta1_ClusterResourcePlacement(t *testing
...
-			Name: "test-nameSpace",
+			Name: "test-namespace",
 			Labels: map[string]string{
 				"region":  rand.String(10),
 				"version": rand.String(4),
 			},
 		},
 	}
+
+	// Common ResourceIdentifier for Namespace tests (cluster-scoped)
```
ryanzhang-oss, line 682: Where is the test case for the namespaceOnly CRP that should not match the resource?

zhiying-lin: existing test, "Skip a placement with selecting ns only"

ryanzhang-oss: I don't get it, if a test already exists then how come it passed?  I was looking for changes in the existing tests to catch the bug but all I see is just addin a key.

zhiying-lin: The bug is in where we compare the resource with the status. Before, we used the ns so any namespace scoped resource will match with the ns only CRP. Now, i changed to use the original key for the namespace scoped resource.

These kind of tests were missing in the original UT.

zhiying-lin: or you're talking about, adding a test case for the namespaceOnly CRP that should not match the resource in the status?

PR #223, opened 2025-08-22
fix: fix msi client used by refresh-token to not reuse tcp connection

nibooy:

Fix msi client used by refresh-token to not reuse tcp connection to prevent race condition where a tcp connection is established and continues being used before a side car inserts an iptables rule to intercepts the calls.

Fixes #

I have:

### DIFF
--- a/pkg/authtoken/providers/azure/azure_msi.go
+++ b/pkg/authtoken/providers/azure/azure_msi.go
@@ -18,6 +18,7 @@ package azure
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
@@ -50,7 +51,25 @@ func New(clientID, scope string) authtoken.Provider {
 // FetchToken gets a new token to make request to the associated fleet' hub cluster.
 func (a *AuthTokenProvider) FetchToken(ctx context.Context) (authtoken.AuthToken, error) {
 	token := authtoken.AuthToken{}
-	opts := &azidentity.ManagedIdentityCredentialOptions{ID: azidentity.ClientID(a.ClientID)}
+	opts := &azidentity.ManagedIdentityCredentialOptions{
+		// We may race at startup with a sidecar which inserts an iptables rule
+		// to intercept IMDS calls.  If we get here before such an iptables rule
+		// is inserted, we will inadvertently connect to real IMDS, which won't
+		// be able to service our request.  IMDS does not set 'Connection:
+		// close' on 400 errors.  Default Go HTTP client behavior will keep the
+		// underlying TCP connection open for re-use, unaffected by iptables,
+		// causing all further requests to continue to be sent to real IMDS and
+		// fail.  Use a separate HTTP client for IMDS calls, where connection
+		// re-use is disabled.
+		ClientOptions: azcore.ClientOptions{
+			Transport: &http.Client{
+				Transport: &http.Transport{
+					DisableKeepAlives: true,
+				},
+			},
+		},
+		ID: azidentity.ClientID(a.ClientID),
+	}
 
 	klog.V(2).InfoS("FetchToken", "client ID", a.ClientID)
 	credential, err := azidentity.NewManagedIdentityCredential(opts)


jim-minter (dismissed):
LGTM

pkg/authtoken/providers/azure/azure_msi.go
```diff
@@ -50,7 +51,25 @@ func New(clientID, scope string) authtoken.Provider {
 // FetchToken gets a new token to make request to the associated fleet' hub cluster.
 func (a *AuthTokenProvider) FetchToken(ctx context.Context) (authtoken.AuthToken, error) {
 	token := authtoken.AuthToken{}
-	opts := &azidentity.ManagedIdentityCredentialOptions{ID: azidentity.ClientID(a.ClientID)}
+	opts := &azidentity.ManagedIdentityCredentialOptions{
+		// We may race at startup with a sidecar which inserts an iptables rule
```
ryanzhang-oss: can we move the msi-adapter to init container?

ryanzhang-oss (dismissed):

pkg/authtoken/providers/azure/azure_msi.go
zhiying-lin: echo what ryan commented, is it possible to solve this issue by making sure the sidecar order? disabling reusing connections means overhead of the connection establishment, slow start, long latency, more server load.

nibooy:
> is there anyway to verify this works?

@ryanzhang-oss, We verified the end-to-end process by testing the fleet member join with a new refresh-token image that has the change. 

Before the fix, we saw occasional join failures due to the msi-adapter losing the race to refresh-token and iptables rule not being set up before refresh-token establishes connection.

After the fix, 6/6 member joins were successful. We also checked that the packet count on the iptables rule (Chain aad-metadata) added by the adapter was incremented (1). 

packet

pkg/authtoken/providers/azure/azure_msi.go
nibooy: I understand your concern. AFAIK, we can't just change the existing msi-adapter image to an init container. We would need to move just the logic of the msi-adapter that adds the iptables rule to the init container and keep rest of the logic that serves the token requests as a regular sidecar.

This requires changes in the source code of both AKS and Arc's msi-adapter. I can start a discussion to get momentum on this, but I think it will take a while to be delivered since it is an image used by many teams. 

What are your thoughts on merging this fix first to fix the flakiness for Arc delivery in the short term while we work on the long term change on the msi-adapter?

pkg/authtoken/providers/azure/azure_msi.go
```diff
@@ -50,7 +51,25 @@ func New(clientID, scope string) authtoken.Provider {
...
+		// be able to service our request.  IMDS does not set 'Connection:
+		// close' on 400 errors.  Default Go HTTP client behavior will keep the
+		// underlying TCP connection open for re-use, unaffected by iptables,
+		// causing all further requests to continue to be sent to real IMDS and
+		// fail.  Use a separate HTTP client for IMDS calls, where connection
+		// re-use is disabled.
+		ClientOptions: azcore.ClientOptions{
+			Transport: &http.Client{
+				Transport: &http.Transport{
+					DisableKeepAlives: true,
```
jwtty: Can we try calling `transport.CloseIdleConnections()` when an err happens on line 89? So that the connection is only closed after a failed request but still can be reused when succeeding.

nibooy: Successfully tested the change and updated the PR.
Confirmed using netstat that there are just two TCP connections for the msi-adapter--the 9090 port connection is for the msi-adapter's readiness probe. 
msi

pkg/authtoken/providers/azure/azure_msi.go
```diff
@@ -50,7 +51,28 @@ func New(clientID, scope string) authtoken.Provider {
...
+	// is inserted, we will inadvertently connect to real IMDS, which won't
+	// be able to service our request.  IMDS does not set 'Connection:
+	// close' on 400 errors.  Default Go HTTP client behavior will keep the
+	// underlying TCP connection open for re-use, unaffected by iptables,
+	// causing all further requests to continue to be sent to real IMDS and
+	// fail.  Use a separate HTTP client for IMDS calls, where connection
+	// re-use is disabled.
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			DisableKeepAlives: true,
```
jwtty: I thought that if we call CloseIdleConnections below, we don't need to disable keepalives.

nibooy: Verified that CloseIdleConnections is sufficient by running join tests and updated the PR to remove DisableKeepAlives.

jwtty (approved):

PR #229, opened 2025-08-28
fix: validate if the resource should be placed

zhiying-lin:

- check if the resource is reserved or not
- add more test cases

Fixes #

I have:

### DIFF
--- a/pkg/controllers/placement/resource_selector.go
+++ b/pkg/controllers/placement/resource_selector.go
@@ -284,13 +284,15 @@ func (r *Reconciler) fetchResources(selector fleetv1beta1.ResourceSelectorTerm, 
 			klog.ErrorS(err, "Cannot get the resource", "gvr", gvr, "name", selector.Name, "namespace", placementKey.Namespace)
 			return nil, controller.NewAPIServerError(true, client.IgnoreNotFound(err))
 		}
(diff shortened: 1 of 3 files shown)


pkg/controllers/placement/resource_selector.go
```diff
@@ -418,16 +441,8 @@ func (r *Reconciler) fetchAllResourcesInOneNamespace(namespaceName string, place
...
 		for _, obj := range objs {
-			uObj := obj.DeepCopyObject().(*unstructured.Unstructured)
-			if uObj.GetDeletionTimestamp() != nil {
-				// skip a to be deleted resource
-				klog.V(2).InfoS("skip the deleting resource by the selector",
-					"placeName", placeName, "namespace", namespaceName, "object", klog.KObj(uObj))
-				continue
-			}
-			shouldInclude, err := utils.ShouldPropagateObj(r.InformerManager, uObj)
+			shouldInclude, err := r.shouldPropagateObj(namespaceName, placeName, obj)
```
jwtty, line 444: Should be named to `placementName`.

jwtty (approved):

pkg/controllers/placement/resource_selector.go
zhiying-lin: it was original code, so merged it first and we can do the refactor later

PR #234, opened 2025-08-29
feat: (re-)enable parallel processing in the work applier

michaelawyu:

This PR (re-)enables parallel processing in the work applier.

I have:

See notes.

### DIFF
--- a/pkg/controllers/workapplier/waves.go
+++ b/pkg/controllers/workapplier/waves.go
@@ -1,0 +1,180 @@
+/*
+Copyright 2025 The KubeFleet Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workapplier
+
+import (
+	"slices"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+)
+
+type waveNumber int
+
+const (
+	lastWave waveNumber = 999
+)
+
+var (
+	// The default wave number for all known Kubernetes resource type.
+	//
+	// Note (chenyu1): the waves below are based on the Helm resource installation
+	// order (see also the Helm source code). Similar objects are grouped together
+	// to achieve best performance.
+	defaultWaveNumberByResourceType = map[string]waveNumber{
+		// Apply namespaces and priority classes first.
+		"namespaces":      0,
+		"priorityclasses": 0,
+		// Apply policies, configuration data, and other static resources second.
+		"networkpolicies":           1,
+		"resourcequotas":            1,
+		"limitranges":               1,
+		"podsecuritypolicies":       1,
+		"poddisruptionbudgets":      1,
+		"serviceaccounts":           1,
+		"secrets":                   1,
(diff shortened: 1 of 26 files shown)


michaelawyu:
Due to a test complication; no tests are submitted in this PR at the moment. Will move additional tests to this PR soon.

pkg/controllers/workapplier/waves.go
```diff
@@ -0,0 +1,180 @@
...
+	lastWave waveNumber = 999
+)
+
+var (
+	// The default wave number for all known Kubernetes resource type.
+	//
+	// Note (chenyu1): the waves below are based on the Helm resource installation
+	// order (see also the Helm source code). Similar objects are grouped together
+	// to achieve best performance.
+	defaultWaveNumberByResourceType = map[string]waveNumber{
```
zhiying-lin, line 38: we also defined the apply order in https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/controllers/placement/resource_selector.go#L47

the orders are not consistent. for example, the ingressclasses.

Can we try to merge them? so that it's always consistent

michaelawyu: Hi Zhiying! Yeah, I was a bit concerned about this too when composing the PR. The reason why we did it separately in this PR was that, if we do not group the resource types but process them invididually, there's a high chance that in each batch there's only 1-2 objects, which kind of defeats the purpose of doing the parallelization.

michaelawyu: The PR also kinds of renders the apply order sorting we did on the hub cluster side redundant (we still need to sort the resources for stability reasons, but they do not have to be done in a specific order -> similar to how the work generator sorts enveloped objects).

michaelawyu: At this moment I am leaning on grouping the resource types for larger batch sizes, but I do not have a particularly strong opinion on the subject matter -> happy to discuss about the topic further.

michaelawyu: As for the apply order part on the hub cluster, will submit another PR to keep things clean after this one is merged.

michaelawyu: (A side note: another reason I am a bit reluctant to change the apply order on the hub cluster side was that it might trigger a rollout on existing workloads, as it will register as a new resource snapshot IIRC? -> we might need to be a bit careful on this 😞)

michaelawyu:
> discussed offline, will update the applyOrder comment in a separate PR.

Will do 👌

pkg/controllers/workapplier/waves.go
ryanzhang-oss: these two orders serve different purposes. The hub one was the apply order before this PR but now it's more to keep the uniqueness for a list of resources (or we don't think there is any change) while the member side is more on the real apply order. Probably should change the name of the hub side list to reflect the fact that it's no longer the order that member will apply.

pkg/controllers/workapplier/waves.go
```diff
@@ -0,0 +1,180 @@
...
+	bundles []*manifestProcessingBundle
+}
+
+// organizeBundlesIntoProcessingWaves organizes the list of bundles into different
+// waves of bundles for parallel processing based on their GVR information.
+func organizeBundlesIntoProcessingWaves(bundles []*manifestProcessingBundle, workRef klog.ObjectRef) []*bundleProcessingWave {
+	// Pre-allocate the map; 7 is the total count of default wave numbers, though
+	// not all wave numbers might be used.
+	waveByNum := make(map[waveNumber]*bundleProcessingWave, 7)
+
```
ryanzhang-oss, line 121: ni: shouldn't this be a bundleProcessingWave array instead? One can precreate all the bundleProcessingWave so there is no need for the getOrAddWave function anymore (but we can keep to save memory).  I am not sure why we need to sort the map to produce an array at the end.

PR #238, opened 2025-09-03
feat: security: obscure drift/diff reportings on secret object data fields

michaelawyu:

This PR will set the work applier to obscure any drift/diff reportings on the data fields of secret objects, so as to protect sensitive information from being revealed outside cluster boundaries.

This behavior is consistent with what ArgoCD does with similar functions.

I have:

- [x] Unit tests
- [x] Integration tests


ryanzhang-oss:
I have a hard time to find the connection between "secret object data fields" and the only place I can see setting this diffCalculatedInDegradedMode to true which is 

	case errors.IsInvalid(err):

PR #244, opened 2025-09-09
test: add additional integration tests for waved work applier manifest processing

michaelawyu:

This PR adds additional integration tests for waved work applier manifest processing.

I have:

N/A

N/A

### DIFF
--- a/pkg/controllers/workapplier/waves_integration_test.go
+++ b/pkg/controllers/workapplier/waves_integration_test.go
@@ -1,0 +1,1218 @@
+/*
+Copyright 2025 The KubeFleet Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workapplier
+
+import (
+	"fmt"
+	"math/rand/v2"
+	"slices"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	schedulingv1 "k8s.io/api/scheduling/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	fleetv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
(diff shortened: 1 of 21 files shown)


michaelawyu:
Note: this PR is a follow-up to #234.

pkg/controllers/workapplier/waves_integration_test.go
```diff
@@ -0,0 +1,1218 @@
...
+					wantWorkConds[idx].ObservedGeneration = work.Generation
+				}
+				if diff := cmp.Diff(
+					work.Status.Conditions, wantWorkConds,
+					ignoreFieldConditionLTTMsg,
+				); diff != "" {
+					return fmt.Errorf("Work status conditions mismatch (-got, +want):\n%s", diff)
+				}
+				return nil
+			}, eventuallyDuration*10, eventuallyInterval*5).Should(Succeed(), "Failed to update work status")
```
weng271190436, line 1157: How long is this test supposed to take? Might be helpful to comment on the x10 wait time for people reading the code

pkg/controllers/workapplier/waves_integration_test.go
```diff
@@ -0,0 +1,1218 @@
...
+			Eventually(appliedWorkRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove the AppliedWork object")
+
+			workRemovedActual := workRemovedActual(workName)
+			Eventually(workRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove the Work object")
+			// The environment prepared by the envtest package does not support namespace
+			// deletion; consequently this test suite would not attempt to verify its deletion.
+		})
+	})
+
+	Context("all waves", Ordered, func() {
```
weng271190436, line 501: The other 3 tests (e.g. single wave, two non-consecutive waves) all have some cleanup steps in AfterAll. Why "all waves" don't have something similar?

pkg/controllers/workapplier/waves_integration_test.go
```diff
@@ -0,0 +1,1218 @@
...
+			gap := placedRole.CreationTimestamp.Sub(placedNS.CreationTimestamp.Time)
+			Expect(gap).To(BeNumerically(">=", parallelizerFixedDelay), "The creation time gap between resources in different waves is less than the fixed delay")
+			Expect(gap).To(BeNumerically("<", parallelizerFixedDelay*2), "The creation time gap between resources in different waves is at least twice as large as the fixed delay")
+		})
+
+		AfterAll(func() {
+			// Delete the Work object and related resources.
+			deleteWorkObject(workName, memberReservedNSName3)
+
+			// Remove the Role object if it still exists.
```
weng271190436, line 471: there are 3 "if it still exists" object deletions, which are Role, PriorityClass, ConfigMap. Are those supposed to be deleted by the controller after deleteWorkObject but in practice the controller failed to delete them? If so why?

michaelawyu: Hi Wei! This is a standard Ginkgo practice where we try to clean up the objects anyhow in the AfterAll block -> this make sures that even if the current test spec fails in-between, other tests can still run with no worries for alreadyExists types of errors (note that our Ginkgo setup allows parallelization and specs assigned to the same Ginkgo process may use the same namespace/resource names).

michaelawyu: Hi Wei! This is a standard Ginkgo practice where we try to clean up the objects anyhow in the AfterAll block -> this make sures that even if the current test spec fails in-between, other tests can still run with no worries for alreadyExists types of errors or any other side-effects (though admittedly the chances for this to occur in this specific integration test setup are quite low with the random suffix + the fact that envtest does not allow namespace GC).

pkg/controllers/workapplier/waves_integration_test.go
michaelawyu: Ah, let me add this; must have missed it somehow.

pkg/controllers/workapplier/waves_integration_test.go
michaelawyu: Hi Wei! As the setup goes for this specific setup between each processing wave there has been a long delay (5 secs) so that we could verify if objects are being applied in batch as expected -> but I do agree that a comment here would make things clearer; let me add this

pkg/controllers/workapplier/waves_integration_test.go
weng271190436: Thanks for explaining why cleanup is necessary!

I still have a question on why only need to clean up the role but not the namespace? From what I can see you applied both role and namespace to the member cluster but in the cleanup you explicitly delete the role only

I was thinking maybe deleteWorkObject handles the namespace cleanup. But then shouldn't deleteWorkObject also handle the role cleanup?

michaelawyu: Hi Wei! The reason why we do not clean up the namespace was that the `envtest` package we use for running integration tests do not support namespace deletion -> if the Work object includes a namespace manifest, after it is applied, it cannot be deleted (and the deletion won't trigger any errors, it's just stuck there) -> so there's no point in deleting the NS. In the environment we do not have GC controllers either so deleting the NS will not delete objects within in cascade.

weng271190436 (approved):

PR #251, opened 2025-09-17
interface: support skipping validation when deleting member cluster

zhiying-lin:

skip the validation when force-deleting member cluster

Fixes #

I have:


jim-minter (approved):
LGTM

apis/cluster/v1beta1/membercluster_types.go
```diff
@@ -74,6 +74,30 @@ type MemberClusterSpec struct {
...
+	// DeleteValidationModeStrict performs strict validation when deleting a MemberCluster.
+	DeleteValidationModeStrict = "Strict"
+)
+
+type DeleteOptions struct {
+	// Mode of validation. Can be "Skip", or "Strict". Default is Strict.
+	// +kubebuilder:validation:Enum=Skip;Strict
+	// +kubebuilder:default=Strict
+	// +kubebuilder:validation:Optional
+	ValidationMode DeleteValidationMode `json:"validationMode,omitempty"`
```
ryanzhang-oss, line 100: Cince we added the delete option as a struct instead of just an Enum, I assume we plan to add or think there is high chance of adding other information here? Anything hint on what that might be?

zhiying-lin: eg, ForceDeleteWaitTime, //  the wait time in minutes before we force delete a member cluster.

Today it is not configurable.

PR #253, opened 2025-09-18
feat: update mc webhook to support skipping validation

zhiying-lin:

Fixes #

I have:


test/e2e/join_and_leave_test.go
```diff
@@ -330,7 +330,63 @@ var _ = Describe("Test member cluster join and leave flow", Label("joinleave"),
...
+					mcObj := &clusterv1beta1.MemberCluster{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: memberCluster.ClusterName,
+						},
+					}
+					err := hubClient.Delete(ctx, mcObj)
+					Expect(err).ShouldNot(Succeed(), "Want the deletion to be denied")
+					var statusErr *apierrors.StatusError
+					Expect(errors.As(err, &statusErr)).To(BeTrue(), fmt.Sprintf("Delete memberCluster call produced error %s. Error type wanted is %s.", reflect.TypeOf(err), reflect.TypeOf(&apierrors.StatusError{})))
+					Expect(statusErr.ErrStatus.Message).Should(MatchRegexp("Please delete serviceExport test-namespace/test-svc in the member cluster before leaving, request is denied"))
```
serbrech, line 376: is there a code to identify the error as a ServiceExport error?
how do we differentiate between the different reasons that prevented the deletion?
parsing the message string is going to break sooner or later

zhiying-lin: unfortunately, the webhook does not provide such feature :(

https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/admission/response.go#L35

to prevent the regression, the only way is to validate the error message

PR #255, opened 2025-09-23
interface: experimental: status back-reporting

michaelawyu:

This PR adds some experimental API changes that allow status back-reporting.

I have:

N/A

N/A

### DIFF
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -542,6 +542,7 @@ type RolloutStrategy struct {
 
 	// ReportBackStrategy describes how to report back the status of applied resources on the member cluster.
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="(self == null) || (self.type == 'Mirror' ? size(self.destination) != 0 : true)",message="when reportBackStrategy.type is 'Mirror', a destination must be specified"
 	ReportBackStrategy *ReportBackStrategy `json:"reportBackStrategy,omitempty"`
 }
 
--- a/config/crd/bases/placement.kubernetes-fleet.io_clusterresourceplacements.yaml
+++ b/config/crd/bases/placement.kubernetes-fleet.io_clusterresourceplacements.yaml
@@ -1987,6 +1987,11 @@ spec:
                     required:
                     - type
                     type: object
+                    x-kubernetes-validations:
+                    - message: when reportBackStrategy.type is 'Mirror', a destination
+                        must be specified
+                      rule: '(self == null) || (self.type == ''Mirror'' ? size(self.destination)
+                        != 0 : true)'
                   rollingUpdate:
                     description: Rolling update config params. Present only if RolloutStrategyType
                       = RollingUpdate.
--- a/config/crd/bases/placement.kubernetes-fleet.io_resourceplacements.yaml
+++ b/config/crd/bases/placement.kubernetes-fleet.io_resourceplacements.yaml
@@ -922,6 +922,11 @@ spec:
                     required:
                     - type
                     type: object
+                    x-kubernetes-validations:
+                    - message: when reportBackStrategy.type is 'Mirror', a destination
+                        must be specified
+                      rule: '(self == null) || (self.type == ''Mirror'' ? size(self.destination)
+                        != 0 : true)'
                   rollingUpdate:
                     description: Rolling update config params. Present only if RolloutStrategyType
                       = RollingUpdate.


apis/placement/v1beta1/work_types.go
```diff
@@ -142,7 +146,7 @@ type DriftDetails struct {
 	// +kubebuilder:validation:Required
 	ObservedInMemberClusterGeneration int64 `json:"observedInMemberClusterGeneration"`
 
-	// FirsftDriftedObservedTime is the timestamp when the drift was first detected.
+	// FirftDriftedObservedTime is the timestamp when the drift was first detected.
```
ryanzhang-oss:
```suggestion
	// FirstDriftedObservedTime is the timestamp when the drift was first detected.
```

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -1469,6 +1473,52 @@ const (
...
+
+// ReportBackStrategy describes how to report back the resource status from member clusters.
+type ReportBackStrategy struct {
+	// Type dictates the type of the report back strategy to use.
+	//
+	// Available options include:
+	//
+	// * Disabled: status back-reporting is disabled. This is the default behavior.
+	//
+	// * MirrorIfPossible: status back-reporting is enabled by copying the status fields verbatim to
```
ryanzhang-oss: we should use CEL to block so it is just "Mirror"

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -1469,6 +1473,52 @@ const (
...
+	//   the corresponding objects on the hub cluster side. This is only performed when the placement
+	//   object has a scheduling policy that selects exactly one member cluster (i.e., a pickFixed
+	//   scheduling policy with exactly one cluster name, or a pickN scheduling policy with the
+	//   numberOfClusters field set to 1). If multiple member clusters are selected, KubeFleet will
+	//   fall back to the Copy strategy.
+	//
+	// * Copy: status back-reporting is enabled by copying the status fields verbatim via the Work API
```
ryanzhang-oss: could use a better name

apis/placement/v1beta1/work_types.go
```diff
@@ -196,6 +200,25 @@ type DiffDetails struct {
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Format=date-time
+	ObservationTime metav1.Time `json:"observationTime"`
+
+	// ObservedInMemberClusterGeneration is the generation of the applied resource on the member
+	// cluster side.
+	//
+	// +kubebuilder:validation:Optional
+	ObservedInMemberClusterGeneration int64 `json:"observedInMemberClusterGeneration"`
```
ryanzhang-oss: those fields are not useful when the mode is "Mirror", right?

apis/placement/v1beta1/clusterresourceplacement_types.go
michaelawyu: Done.

apis/placement/v1beta1/clusterresourceplacement_types.go
michaelawyu: Yeah, I wasn't too confident/keen on this name either. Ryan, do you happen to have a name that you prefer? I was thinking about `CopyToWork` but it feels a bit weird -> would `RetainPerCluster` sounds better?

apis/placement/v1beta1/work_types.go
michaelawyu: Hi Ryan! Yeah, for the Mirror mode it does not matter, but I assume users might still want to peek at the Work object for troubleshooting reasons

apis/placement/v1beta1/work_types.go
michaelawyu: Done 😞 I thought AI would catch this; I was wrong 😞

apis/placement/v1beta1/clusterresourceplacement_types.go
michaelawyu: Hi Ryan! I've updated this to `ReplicatePerCluster`; a few alternatives proposed by copilot include `ReplicateViaWork`, `Relay`, `RelayPerCluster`, `RelayViaWork` -> if there's one which you think would be better, please let me know

apis/placement/v1beta1/work_types.go
michaelawyu: Hi Ryan! Per earlier discussions, I've dropped the `ObservedInMemberClusterGeneration` field.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -1480,6 +1484,52 @@ const (
...
+	//
+	// * ReplicatePerCluster: status back-reporting is enabled by copying the status fields verbatim via the Work API
+	//   on the hub cluster side. Users may look up the status of a specific resource applied to a
+	//   specific member cluster by inspecting the corresponding Work object on the hub cluster side.
+	//
+	// +kubebuilder:default=Disabled
+	// +kubebuilder:validation:Enum=Disabled;Mirror;ReplicatePerCluster
+	// +kubebuilder:validation:Required
+	Type ReportBackStrategyType `json:"type"`
+}
```
ryanzhang-oss, line 1546: can this be spread out in a different way

const (
	
    // ReportBackStrategyTypeDisabled disables status back-reporting from the member clusters.
	ReportBackStrategyTypeDisabled ReportBackType = "Disabled"

    // will copy the status back
	ReportBackStrategyTypeMirror ReportBackType = "Copy"
)

(
  
    // Propagate to the original resource
	ReportBackPropagateTypeOriginal ReportBackPropagateType = "Original"

    // Propagate to work only
	ReportBackPropagateTypeWork ReportBackPropagateType = "Work"
)

// ReportBackStrategy describes how to report back the resource status from member clusters.
type ReportBackStrategy struct {
	
     // Type dictates the type of the report back type to use.
	Type ReportBackType `json:"type"`

	// Type dictates the type of the report back strategy to use.
	Type ReportBackPropagateType `json:"type"`
}

ryanzhang-oss (commented):
btw, it occurred to me that although we don't use v1 work, it's better to copy all the fields there

apis/placement/v1beta1/clusterresourceplacement_types.go
michaelawyu: Yeah, that sounds great; let me make the edits

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -1480,6 +1484,66 @@ const (
...
+	// * OriginalResource: the status fields will be copied verbatim to the original resource on the hub cluster side.
+	//   This is only performed when the placement object has a scheduling policy that selects exactly one member cluster
+	//   (i.e., a pickFixed scheduling policy with exactly one cluster name, or a pickN scheduling policy with the numberOfClusters
+	//   field set to 1).
+	//
+	// * WorkAPI: the status fields will be copied verbatim via the Work API on the hub cluster side. Users may look up
+	//   the status of a specific resource applied to a specific member cluster by inspecting the corresponding Work object
+	//   on the hub cluster side. This is the default behavior.
+	//
+	// +kubebuilder:validation:Enum=OriginalResource;WorkAPI
```
ryanzhang-oss, line 1543: What's the default if user doesn't set it?

apis/placement/v1beta1/clusterresourceplacement_types.go
michaelawyu: Resolving this conversation as it's no longer applicable with the new changes.

apis/placement/v1beta1/clusterresourceplacement_types.go
michaelawyu: Hi Ryan! I changed it to a pointer field so that when the `Type` is `Disabled`, the `Destination` part does not need to be set. We could add a CEL expression on the topper level to make sure that `Destination` cannot be empty when `Type` is `Mirror` -> would this be alright?

michaelawyu: Or we could use a default value here but then `Destination` would be set even if `Type` is Disabled -> it should be just an appearance thing though, the actual behavior would be fine as I understand

ryanzhang-oss: CEL sounds better

michaelawyu: Sure, let me add the CEL expression

michaelawyu: Added.

PR #258, opened 2025-09-26
feat: security: obscure drift/diff reportings on secret object data fields (re-submission)

michaelawyu:

This PR will set the work applier to obscure any drift/diff reportings on the data fields of secret objects, so as to protect sensitive information from being revealed outside cluster boundaries.

This behavior is consistent with what ArgoCD does with similar functions.

I have:

- [x] Unit tests
- [x] Integration tests

This PR is a re-submission of #238, in an attempt to avoid the unexpected rebasing complications.

### DIFF
--- a/pkg/controllers/workapplier/drift_detection_takeover.go
+++ b/pkg/controllers/workapplier/drift_detection_takeover.go
@@ -19,6 +19,7 @@ package workapplier
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/qri-io/jsonpointer"
 	"github.com/wI2L/jsondiff"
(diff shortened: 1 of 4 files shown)


michaelawyu:
To control the PR size and to avoid conflicts, this PR features only unit tests and integration tests for the change. E2E tests will be added as separate PRs.

pkg/controllers/workapplier/drift_detection_takeover.go
```diff
@@ -369,3 +377,42 @@ func (r *Reconciler) removeLeftBehindAppliedWorkOwnerRefs(ctx context.Context, o
...
+//
+// Note (chenyu1): there are other Kubernetes API objects that also feature sensitive data,
+// such as TokenRequest and CertificateSigningRequest; these objects are not included in the list
+// as they have the sensitive information in the status, which are not accounted for in the
+// drift/diff calculation in the very beginning.
+func obscureSensitiveFieldsInPatchDetails(
+	srcObj *unstructured.Unstructured, details []fleetv1beta1.PatchDetail,
+) (sanitizedPatchDetails []fleetv1beta1.PatchDetail) {
+	// Verify if the object is a Secret.
+	if srcObj.GetAPIVersion() != "v1" || srcObj.GetKind() != "Secret" {
```
ryanzhang-oss, line 396: do we even need to compare/diff the secret?

michaelawyu: Hi Ryan! I thought about omitting secret objects as a whole as well, but I fear that some users might still want to check for diffs on the labels/annotations? Another side effect would be that if there's any CRP blocked on secret diffs now, omission will unblock them.

PR #260, opened 2025-09-30
chore: Add scheduler and e2e debugging info 

ryanzhang-oss:

Fixes #

I have:

### DIFF
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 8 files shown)


pkg/scheduler/framework/framework.go
```diff
@@ -603,8 +604,19 @@ type filteredClusterWithStatus struct {
 	status  *Status
 }
 
+// helper type to pretty print a list of filteredClusterWithStatus
+type filteredClusterWithStatusList []*filteredClusterWithStatus
+
+func (cs filteredClusterWithStatusList) String() string {
```
michaelawyu, line 613: Hi Ryan! This is going to be a quite expensive call for large fleets I assume?

michaelawyu: Though I agree that there might not be much for us to do if we do want detailed information....

pkg/scheduler/framework/framework.go
```diff
@@ -906,7 +918,7 @@ func (f *framework) runSchedulingCycleForPickNPlacementType(
 	}
 
 	// Pick the top scored clusters.
-	klog.V(2).InfoS("Picking clusters", "policySnapshot", policyRef)
+	klog.V(2).InfoS("Picking clusters", "policySnapshot", policyRef, "filtered", filtered, "scored", scored)
```
michaelawyu, line 927: Hi Ryan! There should be a similar call on the pickAll part as well.

michaelawyu: `runSchedulingCycleForPickAllPlacementType`, to be specific.

pkg/scheduler/framework/score.go
```diff
@@ -111,3 +114,11 @@ func (sc ScoredClusters) Less(i, j int) bool {
 func (sc ScoredClusters) Swap(i, j int) {
 	sc[i], sc[j] = sc[j], sc[i]
 }
+
+func (sc ScoredClusters) String() string {
```
michaelawyu, line 118: Same performance concern for this... but yeah, there's not much that we could do if the details are needed.

pkg/scheduler/framework/framework.go
michaelawyu: (And our log outputs will be cut off anyway if the list gets too long)

test/e2e/rollout_test.go
```diff
@@ -612,18 +612,20 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
-					meta.SetStatusCondition(&work.Status.Conditions, metav1.Condition{
-						Type:   placementv1beta1.WorkConditionTypeAvailable,
-						Status: metav1.ConditionFalse,
-						Reason: "WorkNotAvailable",
-					})
-				}
-				Expect(hubClient.Status().Update(ctx, &work)).Should(Succeed(), "Failed to update the work")
+				Eventually(func() error {
+					work := placementv1beta1.Work{}
+					Expect(hubClient.Get(ctx, types.NamespacedName{Name: workName, Namespace: namespaceName}, &work)).Should(Succeed(), "Failed to get the work")
```
michaelawyu: Hi Ryan! `Expect` cannot be nested in an `Eventually` block without passed-in Gomega object I think: https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode

PR #277, opened 2025-10-09
test: fix a flakiness issue in E2E test suite (1/)

michaelawyu:

This PR resolves an issue in the RP E2E test suite where a namespace is taken over by the CRP and thus cannot be deleted successfully.

I have:

N/A

N/A

### DIFF
--- a/test/e2e/resource_placement_drift_diff_test.go
+++ b/test/e2e/resource_placement_drift_diff_test.go
@@ -1579,7 +1579,14 @@ var _ = Describe("report diff mode using RP", Label("resourceplacement"), func()
 		})
 
 		AfterAll(func() {
-			ensureRPAndRelatedResourcesDeleted(types.NamespacedName{Name: rpName, Namespace: nsName}, allMemberClusters)
+			// Clean up the created resources.
+
+			// For the RP-related resources, ignore the first member cluster as it has pre-existing resources.
+			ensureRPAndRelatedResourcesDeleted(types.NamespacedName{Name: rpName, Namespace: nsName}, []*framework.Cluster{
+				memberCluster2EastCanary,
+				memberCluster3WestProd,
+			})
+			// Note that the pre-existing namespace (not the configMaps) has been taken over by the CRP.
 			ensureCRPAndRelatedResourcesDeleted(crpName, allMemberClusters)
 		})
 	})


test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -1577,11 +1579,14 @@ var _ = Describe("report diff mode using RP", Label("resourceplacement"), func()
...
 
 		AfterAll(func() {
-			// The pre-existing namespace resources that have not been taken over and must be deleted manually.
-			cleanWorkResourcesOnCluster(memberCluster1EastProd)
-			cleanupAnotherConfigMapOnMemberCluster(types.NamespacedName{Name: cm2Name, Namespace: nsName}, memberCluster1EastProd)
+			// Clean up the created resources.
 
-			ensureRPAndRelatedResourcesDeleted(types.NamespacedName{Name: rpName, Namespace: nsName}, allMemberClusters)
+			// For the RP-related resources, ignore the first member cluster as it has pre-existing resources.
+			ensureRPAndRelatedResourcesDeleted(types.NamespacedName{Name: rpName, Namespace: nsName}, []*framework.Cluster{
```
michaelawyu, line 1585: Hi Ryan! I understand that earlier we discussed about why we do not need to ignore the 1st member cluster here, but upon closer inspection it seems necessary to do so.

michaelawyu: The RP created in this test suite is a `ReportDiff` one with takeover disabled -> it will not own the two pre-existing `configMaps` in the 1st member cluster but the deletion step, if the 1st member cluster is included, will attempt to verify if the two `configMaps` will be deleted, and thus will always fail.

ryanzhang-oss: now I am confused, isn't ensureCRPAndRelatedResourcesDeleted supposed to clean up everything anyway. In this case, even if the resources are no taken over, the namespace is so it shoud delete the entire namespace?

michaelawyu: Hi Ryan! Sorry if I wasn't being clear enough -> what I am trying to say is that, with the original `AfterAll` block:

* `cleanWorkResourcesOnCluster(memberCluster1EastProd)` (line 1581)  will attempt to delete the namespace (and all configmaps under it) -> but this might never complete as the CRP has taken over the namespace and will attempt to re-create it
* line 1582 is redundant in the sense that if 1581 succeeds, it is guaranteed to pass

and in the new setup, for `ensureRPAndRelatedResourcesDeleted` we are checking only cluster 2 and 3 because,

* in the `Before` part of the test logic two configMaps are created -> they will not be taken over by the RP (takeover is disabled)
*  `ensureRPAndRelatedResourcesDeleted` will attempt to verify if all configMaps have been removed, and this will never succeed, as the removal of the RP will not trigger their deletion

michaelawyu: TO-DO (chenyu1): add a step to verify cms will not be deleted.

PR #279, opened 2025-10-09
test: fix a flakiness issue in E2E test suite (2/)

michaelawyu:

This PR resolves:
* an issue in the RP E2E test suite where a namespace is taken over by a CRP and thus cannot be deleted successfully.
* an issue in the CRP E2E test suite where a namespace might get re-created by a CRP.

I have:

N/A

N/A

### DIFF
--- a/test/e2e/setup.sh
+++ b/test/e2e/setup.sh
@@ -127,7 +127,6 @@ helm install hub-agent ../../charts/hub-agent/ \
     --set forceDeleteWaitTime="1m0s" \
     --set clusterUnhealthyThreshold="3m0s" \
     --set logFileMaxSize=1000000 \
-    --set MaxConcurrentClusterPlacement=200 \
     --set resourceSnapshotCreationMinimumInterval=$RESOURCE_SNAPSHOT_CREATION_MINIMUM_INTERVAL \
     --set resourceChangesCollectionDuration=$RESOURCE_CHANGES_COLLECTION_DURATION
 


test/e2e/placement_drift_diff_test.go
```diff
@@ -504,10 +504,14 @@ var _ = Describe("take over existing resources", func() {
 		})
 
 		AfterAll(func() {
+			// The CRP must be deleted first, otherwise the pre-existing namespace might get re-created.
+			//
+			// Also, do not attempt to verify deletion on the first member cluster, as a pre-existing namespace
+			// lives there.
+			ensureCRPAndRelatedResourcesDeleted(crpName, []*framework.Cluster{memberCluster2EastCanary, memberCluster3WestProd})
+
```
ryanzhang-oss, line 512: so if we delete the resources on the 1EastProd first, the CRP might place the resource there but the ensureCRPAndRelatedResourcesDeleted() function still should clean it up afterwards. What did I miss?

test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -98,13 +99,7 @@ var _ = Describe("take over existing resources using RP", Label("resourceplaceme
...
 
 	AfterEach(OncePerOrdered, func() {
-		// The pre-existing resource that have not been taken over and must be deleted manually.
-		cleanupAnotherConfigMapOnMemberCluster(types.NamespacedName{Name: cm2Name, Namespace: nsName}, memberCluster1EastProd)
-		cleanWorkResourcesOnCluster(memberCluster1EastProd)
-
-		// Clean up second configMap on the hub cluster.
-		cleanupAnotherConfigMap(types.NamespacedName{Name: cm2Name, Namespace: nsName})
-
+		// No need to clean up pre-existing resources as the namespace has been taken over.
```
ryanzhang-oss, line 102: this is just an optimization, right?

test/e2e/placement_drift_diff_test.go
michaelawyu: Hi Ryan! The logic in `cleanWorkResourcesOnCluster(memberCluster1EastProd)` will attempt to delete the namespace in use, and then use an Eventually block to wait for the deletion to complete -> but with the CRP constantly re-creating the namespace, there's a chance that this Eventually block might never succeed

test/e2e/resource_placement_drift_diff_test.go
michaelawyu: Hi Ryan! The bug is the same as the one above -> where the logic attempts to remove resources and verify its deletion with an Eventually block, but with the CRP constantly re-creating the namespace, the Eventually block might never get completed.

PR #283, opened 2025-10-14
test: add E2E tests for using SKU based properties for scheduling (limited) + apply minor changes

michaelawyu:

This PR:

* applied a validation logic change on the webhook side to allow usage of SKU properties
* applied a minor naming fix on the SKU property name
* added E2E tests for using SKU based properties for scheduling

I have:

N/A

Some E2E test steps cannot be added now as the SKU usage is random in the current setup. Will add them when we modified the environment setup steps to allow consistent SKU usage.

### DIFF
--- a/test/e2e/placement_pickall_test.go
+++ b/test/e2e/placement_pickall_test.go
@@ -1005,7 +1005,7 @@ var _ = Describe("placing resources using a CRP of PickAll placement type", func
 
 	// TO-DO (chenyu1): modify the E2E setup to allow fixed node SKU assignment, so that the test suite can include
 	// a case where SKU-based scheduling can yield consistent non-empty results.
-	FContext("with affinities, property selectors only (VM size), no matching clusters", Ordered, func() {
+	Context("with affinities, property selectors only (VM size), no matching clusters", Ordered, func() {
 		crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
 		// Cross-ref with the setup script to ensure that this VM size is not used at all.
 		absentVMSize := "Standard_D8s_v3"
--- a/test/e2e/resource_placement_pickall_test.go
+++ b/test/e2e/resource_placement_pickall_test.go
@@ -850,7 +850,7 @@ var _ = Describe("placing namespaced scoped resources using a RP with PickAll po
 		It("should not place resources on any cluster", checkIfRemovedConfigMapFromAllMemberClusters)
 	})
 
-	FContext("with affinities, label and property selectors (VM size), no matching clusters", Ordered, func() {
+	Context("with affinities, label and property selectors (VM size), no matching clusters", Ordered, func() {
 		It("creating the RP should succeed", func() {
 			if !isAzurePropertyProviderEnabled {
 				Skip("Skipping this test spec as Azure property provider is not enabled in the test environment")
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,6 @@ require (
 	go.goms.io/fleet-networking v0.3.3
 	go.uber.org/atomic v1.11.0
 	go.uber.org/zap v1.27.0
-	golang.org/x/exp v0.0.0-20250305212735-054e65f0b394
 	golang.org/x/sync v0.15.0
 	golang.org/x/time v0.11.0
 	gomodules.xyz/jsonpatch/v2 v2.4.0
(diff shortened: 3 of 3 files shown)


test/e2e/placement_pickall_test.go
```diff
@@ -1002,4 +1002,81 @@ var _ = Describe("placing resources using a CRP of PickAll placement type", func
 			ensureCRPAndRelatedResourcesDeleted(crpName, nil)
 		})
 	})
+
+	// TO-DO (chenyu1): modify the E2E setup to allow fixed node SKU assignment, so that the test suite can include
+	// a case where SKU-based scheduling can yield consistent non-empty results.
+	Context("with affinities, property selectors only (VM size), no matching clusters", Ordered, func() {
+		crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+		// Cross-ref with the setup script to ensure that this VM size is not used at all.
+		absentVMSize := "Standard_D8s_v3"
```
weng271190436, line 1011: Does this have to a real SKU name? Might be safer to make it something that is never going to appear in setup script

test/e2e/resource_placement_pickall_test.go
```diff
@@ -849,4 +849,69 @@ var _ = Describe("placing namespaced scoped resources using a RP with PickAll po
...
 		It("should not place resources on any cluster", checkIfRemovedConfigMapFromAllMemberClusters)
 	})
+
+	Context("with affinities, label and property selectors (VM size), no matching clusters", Ordered, func() {
+		It("creating the RP should succeed", func() {
+			if !isAzurePropertyProviderEnabled {
+				Skip("Skipping this test spec as Azure property provider is not enabled in the test environment")
+			}
+			// Cross-ref with the setup script to ensure that this VM size is not used at all.
+			absentVMSize := "Standard_D8s_v3"
```
weng271190436, line 859: same here

pkg/utils/validator/clusterresourceplacement_test.go
```diff
@@ -1242,7 +1242,100 @@ func TestValidateClusterResourcePlacement_PickNPlacementPolicy(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			wantErr: false,
+		},
+		"valid placement policy - PickN with valid property selector name (non-resource, multiple segments)": {
+			policy: &placementv1beta1.PlacementPolicy{
```
weng271190436, line 1260: Is the convention to always create explicit test objects instead of refactor the creation of PlacementPolicy into a function that takes in PlacementType and property selector name as arguments?

test/e2e/placement_pickall_test.go
michaelawyu: Hi Wei! Yeah, for the E2E as there are Karpenter client libs in use for cost calculations we must use real SKU names. I have a PR in preparation that aims to move out property-related tests to its own test setup/suite for simplicity reasons.

pkg/utils/validator/clusterresourceplacement_test.go
michaelawyu: Hi Wei! Yeah, it's difficult to strike a good balance between nice (decoupling/flexibility/simplicity) and reusability. In our test suite we do have cases where certain setup is all wrapped in one common function (e.g., `createCRP`, `createCRPWithApplyStrategy`, etc.), and there are also cases where setup is spelled out directly in the test spec as it is very specific.

michaelawyu: I do not particularly have a strong opinion on this though. We could set up a common function of course, but this might make things a bit difficult to change/reason about in the future. As you have probably already found out in the test code, we already have some occurrences of utility functions being nested in layers to pad in flexibility.

weng271190436: Got it! I don't have strong opinion here. Just trying to learn what I should do when writing tests

weng271190436 (approved):

apis/cluster/v1/membercluster_types.go
```diff
@@ -75,7 +75,20 @@ type MemberClusterSpec struct {
...
+// A valid name should be a string of one or more segments, separated by slashes (/) if applicable.
+//
+// Each segment must be 63 characters or less, start and end with an alphanumeric character,
+// and can include dashes (-), underscores (_), dots (.), and alphanumerics in between.
+//
+// Optionally, the property name can have a prefix, which must be a DNS subdomain up to 253 characters,
+// followed by a slash (/).
+//
+// Examples include:
+// - "avg-resource-pressure"
```
ryanzhang-oss, line 89: do we actually allow this? Do we have any existing property without any prefix?

michaelawyu: Hi Ryan! We do not have any existing property names like this, but it is a valid Kubernetes label name and was allowed with our previous validation rule/specificiation. Would you prefer if I drop this requirement (i.e., must have a prefix)?

ryanzhang-oss: yeah, I think it's better to not allow a property without prefix unless there is a reason for allowing it

PR #292, opened 2025-10-15
fix: fix missing updaterun metrics and refactor

jwtty:

The PR fixes a bug introduced in #195 where I missed the `FleetUpdateRunStatusLastTimestampSeconds` metrics in hub-agent registration.
The PR separate the pkg/metrics module into 3 sub-modules: hub, member, and shared, which contains metrics emitted by hub-agent, member-agent, and both respectively. The metrics registration is done in the `init()` function within each package. I think this makes future addition of metrics easier as code is self-contained in one place and it's easier to identify which metrics are referenced by which component(s).

Fixes #

I have:

### DIFF
--- a/cmd/hubagent/main.go
+++ b/cmd/hubagent/main.go
@@ -49,8 +49,6 @@ import (
 	"github.com/kubefleet-dev/kubefleet/cmd/hubagent/workload"
 	mcv1alpha1 "github.com/kubefleet-dev/kubefleet/pkg/controllers/membercluster/v1alpha1"
 	mcv1beta1 "github.com/kubefleet-dev/kubefleet/pkg/controllers/membercluster/v1beta1"
-	_ "github.com/kubefleet-dev/kubefleet/pkg/metrics/hub"
-	_ "github.com/kubefleet-dev/kubefleet/pkg/metrics/shared"
 	"github.com/kubefleet-dev/kubefleet/pkg/webhook"
 	// +kubebuilder:scaffold:imports
 )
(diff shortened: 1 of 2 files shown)


cmd/hubagent/main.go
```diff
@@ -50,7 +49,8 @@ import (
 	"github.com/kubefleet-dev/kubefleet/cmd/hubagent/workload"
 	mcv1alpha1 "github.com/kubefleet-dev/kubefleet/pkg/controllers/membercluster/v1alpha1"
 	mcv1beta1 "github.com/kubefleet-dev/kubefleet/pkg/controllers/membercluster/v1beta1"
-	fleetmetrics "github.com/kubefleet-dev/kubefleet/pkg/metrics"
+	_ "github.com/kubefleet-dev/kubefleet/pkg/metrics/hub"
```
Arvindthiru: Instead of doing this implict call to Init function within metrics package, why not explicitly define a public function registering metrics and calling that public function within init func of main.go, wherever we need to register metrics ?

jwtty: We don't need to load this at all, removed.

pkg/metrics/hub/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+// Package hub contains metrics exclusively emitted by the hub-agent.
+package hub
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	// These 2 metrics are used in v1alpha1 controller, should be soon deprecated.
```
ryanzhang-oss, line 26: cc @weng271190436

jwtty: And there's one in member metrics too:
image

PR #308, opened 2025-10-28
interface: beforeStageTask for updateRun

Arvindthiru:

- Renaming AfterStageTask to StageTask just the struct to be reused by by both AfterStageTasks, BeforeStageTasks
- Adding new optional field BeforeStageTasks to StageConfig
- No breaking changes were introduced as part of the PR, all JSON names for updated fields remain the same

Fixes #

I have:

### DIFF
--- a/apis/placement/v1beta1/stageupdate_types.go
+++ b/apis/placement/v1beta1/stageupdate_types.go
@@ -292,7 +292,7 @@ type StageConfig struct {
 	BeforeStageTasks []StageTask `json:"beforeStageTasks,omitempty"`
 }
 
-// StageTask is the collection of pre or post stage tasks that ALL need to be completed before starting or moving to the next stage.
+// StageTask is the pre or post stage task that needs to be completed before starting or moving to the next stage.
 type StageTask struct {
 	// The type of the before or after stage task.
 	// +kubebuilder:validation:Enum=TimedWait;Approval
--- a/config/crd/bases/placement.kubernetes-fleet.io_clusterstagedupdateruns.yaml
+++ b/config/crd/bases/placement.kubernetes-fleet.io_clusterstagedupdateruns.yaml
@@ -1878,9 +1878,9 @@ spec:
                             The collection of tasks that each stage needs to complete successfully before moving to the next stage.
                             Each task is executed in parallel and there cannot be more than one task of the same type.
                           items:
-                            description: StageTask is the collection of pre or post
-                              stage tasks that ALL need to be completed before starting
-                              or moving to the next stage.
+                            description: StageTask is the pre or post stage task that
+                              needs to be completed before starting or moving to the
+                              next stage.
                             properties:
                               type:
                                 description: The type of the before or after stage
@@ -1912,9 +1912,9 @@ spec:
                             The collection of tasks that needs to completed successfully by each stage before starting the stage.
                             Each task is executed in parallel and there cannot be more than one task of the same type.
                           items:
-                            description: StageTask is the collection of pre or post
-                              stage tasks that ALL need to be completed before starting
-                              or moving to the next stage.
+                            description: StageTask is the pre or post stage task that
+                              needs to be completed before starting or moving to the
+                              next stage.
                             properties:
                               type:
                                 description: The type of the before or after stage
--- a/config/crd/bases/placement.kubernetes-fleet.io_clusterstagedupdatestrategies.yaml
+++ b/config/crd/bases/placement.kubernetes-fleet.io_clusterstagedupdatestrategies.yaml
@@ -197,9 +197,9 @@ spec:
                         The collection of tasks that each stage needs to complete successfully before moving to the next stage.
                         Each task is executed in parallel and there cannot be more than one task of the same type.
                       items:
-                        description: StageTask is the collection of pre or post stage
-                          tasks that ALL need to be completed before starting or moving
-                          to the next stage.
+                        description: StageTask is the pre or post stage task that
+                          needs to be completed before starting or moving to the next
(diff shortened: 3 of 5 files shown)


apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -282,14 +282,18 @@ type StageConfig struct {
...
+	BeforeStageTask *BeforeStageTask `json:"beforeStageTask,omitempty"`
 }
 
 // AfterStageTask is the collection of post-stage tasks that ALL need to be completed before moving to the next stage.
 type AfterStageTask struct {
 	// The type of the after-stage task.
 	// +kubebuilder:validation:Enum=TimedWait;Approval
 	// +kubebuilder:validation:Required
-	Type AfterStageTaskType `json:"type"`
+	Type StageTaskType `json:"type"`
```
britaniar, line 300: Overall LGTM just worried about the name change. Wont existing updateRuns run into issues with this change?

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -282,14 +282,18 @@ type StageConfig struct {
 	// +kubebuilder:validation:XValidation:rule="!self.exists(e, e.type == 'Approval' && has(e.waitTime))",message="AfterStageTaskType is Approval, waitTime is not allowed"
 	// +kubebuilder:validation:XValidation:rule="!self.exists(e, e.type == 'TimedWait' && !has(e.waitTime))",message="AfterStageTaskType is TimedWait, waitTime is required"
 	AfterStageTasks []AfterStageTask `json:"afterStageTasks,omitempty"`
+
+	// The task that needs to be completed successfully before starting the stage.
+	// +kubebuilder:validation:Optional
+	BeforeStageTask *BeforeStageTask `json:"beforeStageTask,omitempty"`
```
ryanzhang-oss: is it possible to have more than one task?

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -298,6 +302,14 @@ type AfterStageTask struct {
 	WaitTime *metav1.Duration `json:"waitTime,omitempty"`
 }
 
+// BeforeStageTask is the pre-stage task that needs to be completed before starting the stage.
+type BeforeStageTask struct {
```
ryanzhang-oss: do we need to create a different type?

apis/placement/v1beta1/stageupdate_types.go
Arvindthiru: It is possible in the future, hence made BeforeStageTasks a list but we only support Approval as a BeforeStageTask so we currently only support one item within the list

apis/placement/v1beta1/stageupdate_types.go
Arvindthiru: Changed to StageTask for both BeforeStageTask, AfterStageTask

apis/placement/v1beta1/stageupdate_types.go
ryanzhang-oss: we didn't change the json name "type". I verified in the CRD definition

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -281,15 +281,23 @@ type StageConfig struct {
...
 }
 
-// AfterStageTask is the collection of post-stage tasks that ALL need to be completed before moving to the next stage.
-type AfterStageTask struct {
-	// The type of the after-stage task.
+// StageTask is the pre or post stage task that needs to be completed before starting or moving to the next stage.
+type StageTask struct {
+	// The type of the before or after stage task.
 	// +kubebuilder:validation:Enum=TimedWait;Approval
 	// +kubebuilder:validation:Required
```
serbrech, line 299: How do you handle the fact that you are adding a new property that is required?
I would expect this to fail get+re-put of existing updateruns?
at least any client that does not bump the SDK will fail manipulating the updatrun object, as it won't be able to serialize the Type property?
Is there a way to configure defaults on added properties? is there a way to test this?

serbrech: I think that this will work as existing updateruns would have no value, so it will get a default value on read of "TimedWait", which is the correct expectation.
new updateruns are required to specify the type because of validation.

serbrech (commented):
I think it would really help if you clearly outlined in the PR description what APIs you are changing, what is the behavior expectation with existing data.
how defaulting and validation is handled, instead of leaving that to the reviewer.
It's not clear if what I infered from reading the code is the intended behavior.

Same with the change of type and why it's fine/what are the implications

apis/placement/v1beta1/stageupdate_types.go
serbrech: are these changes not in yet?

apis/placement/v1beta1/stageupdate_types.go
Arvindthiru: This ia an existing object I'm re-naming for re-use

apis/placement/v1beta1/stageupdate_types.go
Arvindthiru: They are in as part of this commit https://github.com/kubefleet-dev/kubefleet/pull/308/commits/d9bc4c7e99806bc140293619c8fd39d704cd98ad

apis/placement/v1beta1/stageupdate_types.go
serbrech: the Type property on the object is new, and is required.

serbrech (approved):

PR #314, opened 2025-10-29
interface: start/stop API for updateRun

Arvindthiru:

This PR aims to add new API changes to allow users to set the following intent on an updateRun,

- Started
- Stopped
- Abandoned

Adding a new spec field called state which is an enum
Adding another condition type for updateRun status called Abandoned

Fixes #

I have:

### DIFF
--- a/test/apis/placement/v1beta1/api_validation_integration_test.go
+++ b/test/apis/placement/v1beta1/api_validation_integration_test.go
@@ -1664,11 +1664,6 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 			updateRun.Spec.State = placementv1beta1.StateAbandoned
 			Expect(hubClient.Update(ctx, updateRun)).Should(Succeed())
 		})
-
-		It("should allow staying in NotStarted state", func() {
-			updateRun.Spec.State = placementv1beta1.StateNotStarted
-			Expect(hubClient.Update(ctx, updateRun)).Should(Succeed())
-		})
 	})
 
 	Context("Test ClusterStagedUpdateRun State API validation - valid Started state transitions", func() {
@@ -1698,11 +1693,6 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 
 		It("should allow transition from Started to Abandoned", func() {
 			updateRun.Spec.State = placementv1beta1.StateAbandoned
-			Expect(hubClient.Update(ctx, updateRun)).Should(Succeed())
-		})
-
-		It("should allow staying in Started state", func() {
-			updateRun.Spec.State = placementv1beta1.StateStarted
 			Expect(hubClient.Update(ctx, updateRun)).Should(Succeed())
 		})
 	})
@@ -1736,37 +1726,6 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 		})
 
 		It("should allow transition from Stopped to Abandoned", func() {
-			updateRun.Spec.State = placementv1beta1.StateAbandoned
-			Expect(hubClient.Update(ctx, updateRun)).Should(Succeed())
-		})
-
-		It("should allow staying in Stopped state", func() {
-			updateRun.Spec.State = placementv1beta1.StateStopped
-			Expect(hubClient.Update(ctx, updateRun)).Should(Succeed())
-		})
-	})
-
-	Context("Test ClusterStagedUpdateRun State API validation - valid Abandoned state transitions", func() {
-		var updateRun *placementv1beta1.ClusterStagedUpdateRun
-		updateRunName := fmt.Sprintf(validupdateRunNameTemplate, GinkgoParallelProcess())
-
-		BeforeEach(func() {
-			updateRun = &placementv1beta1.ClusterStagedUpdateRun{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: updateRunName,
-				},
-				Spec: placementv1beta1.UpdateRunSpec{
-					State: placementv1beta1.StateAbandoned,
-				},
-			}
-			Expect(hubClient.Create(ctx, updateRun)).Should(Succeed())
-		})
-
-		AfterEach(func() {
-			Expect(hubClient.Delete(ctx, updateRun)).Should(Succeed())
-		})
-
-		It("should allow staying in Abandoned state", func() {
 			updateRun.Spec.State = placementv1beta1.StateAbandoned
 			Expect(hubClient.Update(ctx, updateRun)).Should(Succeed())
 		})


.github/.copilot/breadcrumbs/2025-10-29-1500-start-stop-updaterun-api.md
```diff
@@ -0,0 +1,104 @@
+# Start/Stop UpdateRun API Implementation
```
ryanzhang-oss: do we need to check this in?

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -167,6 +168,13 @@ type UpdateRunSpec struct {
 	// and recorded in the status field.
 	// +kubebuilder:validation:Required
 	StagedUpdateStrategyName string `json:"stagedRolloutStrategyName"`
+
+	// Started indicates whether the update run should be started.
+	// When false or nil, the update run will initialize but not execute.
+	// When true, the update run will begin execution.
+	// Changing from true to false will gracefully stop the update run.
+	// +kubebuilder:validation:Optional
+	Started *bool `json:"started,omitempty"`
```
ryanzhang-oss: please do not use Boolean in the API

jwtty: Do we allow resume?

Arvindthiru: Start -> Stop -> Start (this would be resuming ?)

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -167,6 +180,13 @@ type UpdateRunSpec struct {
 	// and recorded in the status field.
 	// +kubebuilder:validation:Required
 	StagedUpdateStrategyName string `json:"stagedRolloutStrategyName"`
+
+	// ExecutionState indicates the desired execution state of the update run.
+	// When nil or "Stopped", the update run will initialize but not execute.
+	// When "Started", the update run will begin or continue execution.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=Started;Stopped
```
ryanzhang-oss: why make this optional instead of having a default?

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -146,6 +147,18 @@ func (c *ClusterStagedUpdateRun) SetUpdateRunStatus(status UpdateRunStatus) {
...
+// +enum
+type UpdateRunExecutionState string
+
+const (
+	// UpdateRunExecutionStateStarted indicates the update run should be started and actively executing.
+	UpdateRunExecutionStateStarted UpdateRunExecutionState = "Started"
+
+	// UpdateRunExecutionStateStopped indicates the update run should be stopped and not executing.
+	UpdateRunExecutionStateStopped UpdateRunExecutionState = "Stopped"
+)
```
ryanzhang-oss, line 170: this is in the spec, so they are desired states

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -167,6 +180,13 @@ type UpdateRunSpec struct {
 	// and recorded in the status field.
 	// +kubebuilder:validation:Required
 	StagedUpdateStrategyName string `json:"stagedRolloutStrategyName"`
+
+	// ExecutionState indicates the desired execution state of the update run.
+	// When nil or "Stopped", the update run will initialize but not execute.
+	// When "Started", the update run will begin or continue execution.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=Started;Stopped
+	ExecutionState *UpdateRunExecutionState `json:"executionState,omitempty"`
```
ryanzhang-oss: I am not sure if this is the best naming, maybe ask copilot?

Arvindthiru: Changed it to state to make it clean, simple

apis/placement/v1beta1/stageupdate_types.go
Arvindthiru: Changed it default, but now default is Stop which seems awkward. Also thought about having the states as Active/Inactive

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -167,6 +180,14 @@ type UpdateRunSpec struct {
 	// and recorded in the status field.
 	// +kubebuilder:validation:Required
 	StagedUpdateStrategyName string `json:"stagedRolloutStrategyName"`
+
+	// State indicates the desired state of the update run.
+	// When "Stop", the update run will initialize but not execute.
```
ryanzhang-oss: what about pause?

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -364,6 +385,12 @@ const (
 	// - "Unknown": The staged update run initialization has started.
 	StagedUpdateRunConditionInitialized StagedUpdateRunConditionType = "Initialized"
 
+	// StagedUpdateRunConditionStarted indicates whether the staged update run has been started.
+	// Its condition status can be one of the following:
+	// - "True": The staged update run has been started and is ready to progress.
+	// - "False": The staged update run is stopped or not yet started.
+	StagedUpdateRunConditionStarted StagedUpdateRunConditionType = "Started"
```
ryanzhang-oss: how to express pause/resume?

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -147,6 +148,25 @@ func (c *ClusterStagedUpdateRun) SetUpdateRunStatus(status UpdateRunStatus) {
...
+
+const (
+	// StateNotStarted indicates the update run has been initialized but execution has not started.
+	// This is the default state after initialization completes.
+	StateNotStarted State = "NotStarted"
+
+	// StateStarted indicates the update run should execute (or resume execution if paused).
+	StateStarted State = "Started"
+
+	// StateStopped indicates the update run should pause execution.
```
ryanzhang-oss:
```suggestion
	// StateStopped indicates the update run should pause execution. The rollout can be resumed by marking the state as "started" again.
```

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -147,6 +148,25 @@ func (c *ClusterStagedUpdateRun) SetUpdateRunStatus(status UpdateRunStatus) {
...
+	StateStopped State = "Stopped"
+
+	// StateAbandoned indicates the update run should be abandoned and terminated.
```
ryanzhang-oss:
```suggestion
	// StateAbandoned indicates the update run should be abandoned and terminated. An Abandoned rollout can not be started again. It's a terminal state.
```

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -147,6 +148,25 @@ func (c *ClusterStagedUpdateRun) SetUpdateRunStatus(status UpdateRunStatus) {
 	c.Status = status
 }
 
+// State represents the desired state of an update run.
+// +enum
+type State string
+
+const (
+	// StateNotStarted indicates the update run has been initialized but execution has not started.
+	// This is the default state after initialization completes.
```
ryanzhang-oss: The state here is the "desired" state set by the user but the whole description  makes this look like a status instead of spec.

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -396,6 +426,11 @@ const (
 	// - "True": The staged update run is completed successfully.
 	// - "False": The staged update run encountered an error and stopped.
 	StagedUpdateRunConditionSucceeded StagedUpdateRunConditionType = "Succeeded"
+
+	// StagedUpdateRunConditionAbandoned indicates whether the staged update run has been abandoned.
+	// Its condition status can be one of the following:
+	// - "True": The staged update run has been abandoned by user request.
+	StagedUpdateRunConditionAbandoned StagedUpdateRunConditionType = "Abandoned"
```
ryanzhang-oss: do we need this?

stopped is "progressing == false" and "reason is stopped"

Abandoned is "progressing == false" and "reason is Abandoned"

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -746,6 +781,7 @@ func (c *ClusterApprovalRequestList) GetApprovalRequestObjs() []ApprovalRequestO
 // +kubebuilder:printcolumn:JSONPath=`.spec.resourceSnapshotIndex`,name="Resource-Snapshot-Index",type=string
 // +kubebuilder:printcolumn:JSONPath=`.status.policySnapshotIndexUsed`,name="Policy-Snapshot-Index",type=string
 // +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Initialized")].status`,name="Initialized",type=string
+// +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Progressing")].status`,name="Progressing",type=string
 // +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Succeeded")].status`,name="Succeeded",type=string
```
ryanzhang-oss, line 789: why only add progressing but not abanoned if we want that condition?

test/e2e/cluster_staged_updaterun_test.go
```diff
@@ -1668,6 +1668,7 @@ func createClusterStagedUpdateRunSucceed(updateRunName, crpName, resourceSnapsho
 			PlacementName:            crpName,
 			ResourceSnapshotIndex:    resourceSnapshotIndex,
 			StagedUpdateStrategyName: strategyName,
+			State:                    placementv1beta1.StateStarted,
```
ryanzhang-oss: shouldn't the default be "stateNotStarted"?  It's the default so we don't need to add it?

Arvindthiru: Field validation fails if value is state is not specified 

Image

I tried applying the object using an yaml file it set the default value correctly but when it's go struct field validation kicks in before the default value is set. Since our updateRuns start by default I set it to StateStarted for now

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -168,6 +191,16 @@ type UpdateRunSpec struct {
...
+
+	// State indicates the desired state of the update run.
+	// NotStarted: The update run is initialized but execution has not started (default).
+	// Started: The update run should execute or resume execution.
+	// Stopped: The update run should pause execution.
+	// Abandoned: The update run should be abandoned and terminated.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=NotStarted;Started;Stopped;Abandoned
+	// +kubebuilder:default=NotStarted
+	State State `json:"state"`
```
ryanzhang-oss: we need to guard against a few invalid state transition like start -> NotStarted. Abandoned->*  it seems CEL can do that?

ryanzhang-oss: we also need to block any run that start with other than "NotStarted" state.  I am not even sure if this is something that user should set. It seems like every run always starts with this state so there is no need for user to specify this state ever.

ryanzhang-oss: as the previous vesion shows, NotStarted is more like a status than a state.

test/e2e/cluster_staged_updaterun_test.go
Arvindthiru: Needed to add omitempty, fixex

apis/placement/v1beta1/stageupdate_types.go
Arvindthiru: Other alternative suggested by AI Pending or Ready. 

But AI seems to think NotStarted already offers Explicit Intent Declaration from the user

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -147,27 +147,66 @@ func (c *ClusterStagedUpdateRun) SetUpdateRunStatus(status UpdateRunStatus) {
...
+	StateStopped State = "Stopped"
+
+	// StateAbandoned describes user intent to abandon the update run.
+	// This is a terminal state; once set, it cannot be changed.
+	StateAbandoned State = "Abandoned"
+)
+
 // UpdateRunSpec defines the desired rollout strategy and the snapshot indices of the resources to be updated.
 // It specifies a stage-by-stage update process across selected clusters for the given ResourcePlacement object.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.state) || oldSelf.state != 'NotStarted' || self.state != 'Stopped'",message="invalid state transition: cannot transition from NotStarted to Stopped"
```
ryanzhang-oss, line 174: would the logic blow more straightforward to read?

rule="!(has(oldSelf.state) && oldSelf.state == 'NotStarted' && self.state == 'Stopped')"

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -147,27 +147,66 @@ func (c *ClusterStagedUpdateRun) SetUpdateRunStatus(status UpdateRunStatus) {
...
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.state) || oldSelf.state != 'Started' || self.state != 'NotStarted'",message="invalid state transition: cannot transition from Started to NotStarted"
```
ryanzhang-oss, line 175: rule="!(has(oldSelf.state) &&  (oldSelf.state == 'Started' || oldSelf.state == 'Stopped' ) && self.state == 'Notstarted')"

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -147,27 +147,66 @@ func (c *ClusterStagedUpdateRun) SetUpdateRunStatus(status UpdateRunStatus) {
...
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="stagedRolloutStrategyName is immutable"
 	StagedUpdateStrategyName string `json:"stagedRolloutStrategyName"`
+
+	// State indicates the desired state of the update run.
+	// NotStarted: The update run is initialized but execution has not started (default).
+	// Started: The update run should execute or resume execution.
+	// Stopped: The update run should pause execution.
+	// Abandoned: The update run should be abandoned and terminated.
+	// +kubebuilder:validation:Optional
```
ryanzhang-oss, line 206: I wonder why the validation is optional?

Arvindthiru: Users allowed to not specify state, so we can apply the default state

PR #315, opened 2025-10-30
feat: add CEL validation to make placement type immutable for RP

### LINKED ISSUE #311: [BUG] RP allows bad creation and update which the controller can't handle
### **Describe the bug**
For example, one can create an RP with PickN placement strategy but no number of clusters property.

### **Environment**
Please provide the following: 
- Hub cluster details
- Member cluster details

### **To Reproduce**
Create an RP with PickN placement strategy without number of clusters property and the controller will fail to place the resource

### **Expected behavior**
Some validation should stop bad RP from being created/updated

weng271190436:

Fixes #311 

I have:

### DIFF
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -135,7 +135,7 @@ type PlacementSpec struct {
 	// Policy defines how to select member clusters to place the selected resources.
 	// If unspecified, all the joined member clusters are selected.
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:XValidation:rule="!(has(self.placementType) && has(oldSelf.placementType) && self.placementType != oldSelf.placementType)",message="placement type is immutable"
+	// +kubebuilder:validation:XValidation:rule="!(self.placementType != oldSelf.placementType)",message="placement type is immutable"
 	Policy *PlacementPolicy `json:"policy,omitempty"`
 
 	// The rollout strategy to use to replace existing placement with new ones.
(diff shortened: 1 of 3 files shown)


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -1628,6 +1628,7 @@ type ResourcePlacement struct {
 
 	// The desired state of ResourcePlacement.
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="!((has(oldSelf.policy) && !has(self.policy)) || (has(oldSelf.policy) && has(self.policy) && has(self.policy.placementType) && has(oldSelf.policy.placementType) && self.policy.placementType != oldSelf.policy.placementType))",message="placement type is immutable"
```
ryanzhang-oss: can we move this to the PlacementSpec which is shared by RP and CRP?

weng271190436: the first part of the check cannot be moved to Spec

moved the 2nd part

```yaml
!(has(oldSelf.policy) && !has(self.policy))
```

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -135,6 +135,7 @@ type PlacementSpec struct {
 	// Policy defines how to select member clusters to place the selected resources.
 	// If unspecified, all the joined member clusters are selected.
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="!(has(self.placementType) && has(oldSelf.placementType) && self.placementType != oldSelf.placementType)",message="placement type is immutable"
```
ryanzhang-oss: do we need has(self.placementType) && has(oldSelf.placementType)  part?

PlacementType is not an optional field.

weng271190436: PlacementType is technically "optional" though it has a default value PickAll

But after removing this validation still able to pass e2e tests so removing

PR #321, opened 2025-11-06
feat: add clusterProfile status fields

ryanzhang-oss:

Add k8s version and cluster api-server to the clusterProfile status by fetching them from the member agent side. 
This is to implement [kep5339](https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/5339-clusterprofile-plugin-credentials/README.md)

Fixes #

I have:

### DIFF
--- a/.github/workflows/chart.yml
+++ b/.github/workflows/chart.yml
@@ -18,7 +18,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           submodules: true
           fetch-depth: 0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
 
       - name: Check out code into the Go module directory
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
               
       - name: Set up Ginkgo CLI
         run: |
@@ -91,7 +91,7 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
 
       - name: Check out code into the Go module directory
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
       
       - name: Install Ginkgo CLI
         run: |
--- a/.github/workflows/code-lint.yml
+++ b/.github/workflows/code-lint.yml
@@ -43,7 +43,7 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
 
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
         with:
           submodules: true
 
@@ -64,7 +64,7 @@ jobs:
         go-version: ${{ env.GO_VERSION }}
 
     - name: Check out code into the Go module directory
-      uses: actions/checkout@v5
+      uses: actions/checkout@v6
 
     - name: golangci-lint
       run: make lint
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -38,7 +38,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v5
+      uses: actions/checkout@v6
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -16,7 +16,7 @@ jobs:
         with:
           egress-policy: audit
 
-      - uses: actions/checkout@ff7abcd0c3c05ccf6adc123a8cd1fd4fb30fb493 # v4.1.7
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v4.1.7
       - uses: codespell-project/actions-codespell@8f01853be192eb0f849a5c7d721450e7a467c579 # master
         with:
           check_filenames: true
--- a/.github/workflows/markdown-lint.yml
+++ b/.github/workflows/markdown-lint.yml
@@ -10,7 +10,7 @@ jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - uses: tcort/github-action-markdown-link-check@v1
       with:
         # this will only show errors in the output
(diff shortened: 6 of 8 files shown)


pkg/propertyprovider/azure/provider.go
```diff
@@ -423,6 +465,46 @@ func (p *PropertyProvider) collectAvailableResource(_ context.Context, usage *cl
 	usage.Available = available
 }
 
+// collectK8sVersion collects the Kubernetes server version information.
+// It uses a cache with a 15-minute TTL to minimize API calls to the discovery client.
+func (p *PropertyProvider) collectK8sVersion(_ context.Context, properties map[clusterv1beta1.PropertyName]clusterv1beta1.PropertyValue) {
+	now := time.Now()
+
+	// Check if we have a cached version that is still valid.
+	p.k8sVersionMutex.RLock()
```
michaelawyu: Hi Ryan! Not really a comment or something to change, but it seems that a simple mutex would suffice here :)

pkg/propertyprovider/azure/provider_test.go
```diff
@@ -949,3 +964,192 @@ func TestCollectWithDisabledFeatures(t *testing.T) {
 		})
 	}
 }
+
+// mockDiscoveryClient is a mock implementation of the discovery.ServerVersionInterface.
+type mockDiscoveryClient struct {
```
michaelawyu: Hi Ryan! Just a nit: discoveryClient seems to have a fake implementation in the client-go package: https://github.com/kubernetes/client-go/blob/d033c497ffef47be9b4f81abde5c3d94dd78089a/discovery/fake/discovery.go#L38.

pkg/propertyprovider/commons.go
```diff
@@ -24,6 +24,12 @@ const (
 	// NodeCountProperty is a property that describes the number of nodes in the cluster.
 	NodeCountProperty = "kubernetes-fleet.io/node-count"
 
+	// K8sVersionProperty is a property that describes the Kubernetes version of the cluster.
+	K8sVersionProperty = "k8s.io/k8s-version"
```
michaelawyu, line 28: Hi Ryan! Just a nit here: I vaguely recall that earlier we were told not to use anything with `k8s.io` or `kubernetes.io` in the names; instead all non-Azure properties switched to `kubernetes-fleet.io`; I don't remember the reasoning behind this though...

ryanzhang-oss: this is designed as an upstream property following the ClusterProperty API KEP

pkg/propertyprovider/azure/provider.go
ryanzhang-oss: I thought that this function will be called concurrently? Using a reader writer lock can theorically increase the thoroughput?

pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -167,10 +170,43 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	}
+	// TODO throw and unexpected error if clusterEntryPoint is not found
+	// We don't have a way to get it yet
+	clusterEntry, exists := mc.Status.Properties[propertyprovider.ClusterEntryPointProperty]
+	if exists {
+		klog.V(3).InfoS("Get Kubernetes cluster entry point from member cluster status", "clusterEntryPoint", clusterEntry.Value, "clusterProfile", klog.KObj(cp))
+		cp.Status.AccessProviders[0].Cluster.Server = clusterEntry.Value
+	}
+	// Get the CA Data
+	certificateAuthorityData, exists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
```
serbrech, line 197: why are we storing/exposing the CAauthority here?
what makes this necessary?

ryanzhang-oss: This is to implement [kep5339](https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/5339-clusterprofile-plugin-credentials/README.md)

ryanzhang-oss: the hub CA data is not a secret, right?

pkg/propertyprovider/azure/provider.go
```diff
@@ -423,6 +481,42 @@ func (p *PropertyProvider) collectAvailableResource(_ context.Context, usage *cl
...
+		}
+		klog.V(2).InfoS("Using cached Kubernetes version", "version", p.cachedK8sVersion, "cacheAge", now.Sub(p.cachedK8sVersionObservedTime))
+		return
+	}
+
+	// Cache is expired or empty, fetch the version from the discovery client.
+	klog.V(2).Info("Fetching Kubernetes version from discovery client")
+	serverVersion, err := p.discoveryClient.ServerVersion()
+	if err != nil {
+		klog.ErrorS(err, "Failed to get Kubernetes server version from discovery client")
```
michaelawyu, line 506: Hi Ryan! The provider does a mechanism where we could populate some conditions for failed properties; we do not have a hard requirement for this though

pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -167,10 +170,43 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	if exists {
+		klog.V(3).InfoS("Get Kubernetes cluster certificate authority data from member cluster status", "clusterProfile", klog.KObj(cp))
+		cp.Status.AccessProviders[0].Cluster.CertificateAuthorityData = []byte(certificateAuthorityData.Value)
+	} else {
+		// throw an alert
+		_ = controller.NewUnexpectedBehaviorError(fmt.Errorf("cluster certificate authority data not found in member cluster %s status", mc.Name))
+		cp.Status.AccessProviders[0].Cluster.InsecureSkipTLSVerify = true
```
michaelawyu, line 204: Hi Ryan! Maybe this should be controlled by another property (or flag) instead of falling back to the no TLS option (as it might not be available)?

test/e2e/join_and_leave_test.go
```diff
@@ -476,6 +477,71 @@ var _ = Describe("Test member cluster force delete flow", Label("joinleave"), Or
 	})
 })
 
+var _ = Describe("Test member cluster join and leave with clusterProfile", Label("joinleave"), Ordered, Serial, func() {
+	clusterProfileList := &clusterinventory.ClusterProfileList{}
+	Context("Test cluster profile and ", Label("joinleave"), Ordered, Serial, func() {
```
michaelawyu, line 482: Hi Ryan! Just a nit: the spec name seems to be missing a part.

PR #326, opened 2025-11-10
test: add e2e test using label to manage CRP and RP

zhiying-lin:

add a test case, when the admin manages the ns using the CRP, and the developers manages the application using RP based on the CRP result via cluster labeling.

Fixes #

I have:

e2e tests

### DIFF
--- a/test/e2e/mixed_placement_test.go
+++ b/test/e2e/mixed_placement_test.go
@@ -412,7 +412,7 @@ var _ = Describe("mixed ClusterResourcePlacement and ResourcePlacement positive 
 			}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove labels from member clusters")
 		})
 
-		It("picking N clusters with no affinities/topology spread constraints (pick by cluster names in alphanumeric order)", func() {
+		It("picking fixed cluster", func() {
 			crp := &placementv1beta1.ClusterResourcePlacement{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: crpName,


test/e2e/mixed_placement_test.go
```diff
@@ -373,3 +375,164 @@ var _ = Describe("mixed ClusterResourcePlacement and ResourcePlacement negative
...
+							UnavailablePeriodSeconds: ptr.To(2),
+						},
+					},
+				},
+			}
+			Expect(hubClient.Create(ctx, crp)).To(Succeed(), "Failed to create CRP %s", crpName)
+		})
+
+		It("should update CRP status as expected", func() {
+			crpStatusUpdatedActual := crpStatusUpdatedActual(workNamespaceIdentifiers(), wantSelectedClusters, nil, "0")
```
ryanzhang-oss, line 441: how can we guarantee that we alwasy select this cluster? IIRC, we just did pick 1, it's up to the pick algorithm. Why not just use pick fix to ping it?

zhiying-lin: it was explained in line 415, "picking N clusters with no affinities/topology spread constraints (pick by cluster names in alphanumeric order)"

our decision is consistent right not, though it may lead to the hot spot problem.

ryanzhang-oss: does our API specify that "with no affinities/topology spread constraints we will pick by cluster names in alphanumeric order"? If not, then this is implementation detail which we can change in the future. As you pointed out, pick the first according to alphabetic is probably not the best.

PR #327, opened 2025-11-11
feat: add status back-reporting support (1/, work applier side)

michaelawyu:

This PR adds support for status back-reporting on the work applier side.

I have:

- [x] Unit tests
- [x] Integration tests

E2E tests are only possible when the full flow has been implemented.


pkg/controllers/workapplier/status.go
```diff
@@ -643,3 +657,54 @@ func prepareRebuiltManifestCondQIdx(bundles []*manifestProcessingBundle) map[str
...
+) {
+	if inMemberClusterObj == nil || inMemberClusterObj.Object == nil {
+		// Do a sanity check; normally this will never occur (as status back-reporting
+		// only applies to objects that have been successfully applied).
+		//
+		// Should this unexpected situation occurs, the work applier does not register
+		// it as an error; the object shall be ignored for the status back-reporting
+		// part of the reconciliation loop.
+		wrapperErr := fmt.Errorf("attempted to back-report status for a manifest that has not been applied yet or cannot be found on the member cluster side")
+		_ = controller.NewUnexpectedBehaviorError(wrapperErr)
```
weng271190436, line 726: what does this do? Does "controller.NewUnexpectedBehaviorError" have any side effect that you want?

pkg/controllers/workapplier/status.go
```diff
@@ -643,3 +657,54 @@ func prepareRebuiltManifestCondQIdx(bundles []*manifestProcessingBundle) map[str
...
+	// have trouble serializing/deserializing the wrapper object when it's written/read to/from
+	// the API server.
+	statusBackReportingWrapper["apiVersion"] = inMemberClusterObj.GetAPIVersion()
+	statusBackReportingWrapper["kind"] = inMemberClusterObj.GetKind()
+	statusBackReportingWrapper["status"] = inMemberClusterObj.Object["status"]
+	statusData, err := json.Marshal(statusBackReportingWrapper)
+	if err != nil {
+		// This normally should never occur.
+		wrappedErr := fmt.Errorf("failed to marshal wrapped back-reported status: %w", err)
+		_ = controller.NewUnexpectedBehaviorError(wrappedErr)
```
weng271190436, line 748: why discard the newly created error?

pkg/controllers/workapplier/controller_integration_test.go
```diff
@@ -9631,3 +9642,412 @@ var _ = Describe("negative cases", func() {
...
+				if err := json.Unmarshal(backReportedDeployStatusWrapper, deployWithBackReportedStatus); err != nil {
+					return fmt.Errorf("failed to unmarshal wrapped back-reported deployment status: %w", err)
+				}
+				currentDeployWithStatus := &appsv1.Deployment{}
+				if err := memberClient1.Get(ctx, client.ObjectKey{Namespace: nsName, Name: deployName}, currentDeployWithStatus); err != nil {
+					return fmt.Errorf("failed to retrieve Deployment object from member cluster side: %w", err)
+				}
+
+				if diff := cmp.Diff(deployWithBackReportedStatus.Status, currentDeployWithStatus.Status); diff != "" {
+					return fmt.Errorf("back-reported deployment status mismatch (-got, +want):\n%s", diff)
```
weng271190436: is this supposed to be `(-want +got)` or `(-got, +want)`?

In this repo sometimes people do `(-want, +got)` sometimes `(-got, +want)`

Image

pkg/controllers/workapplier/controller_integration_test.go
```diff
@@ -9631,3 +9642,412 @@ var _ = Describe("negative cases", func() {
...
+			Eventually(regularDeployRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove the deployment object")
+
+			regularCMRemovedActual := regularConfigMapRemovedActual(nsName, configMapName)
+			Eventually(regularCMRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove the configMap object")
+
+			// Kubebuilder suggests that in a testing environment like this, to check for the existence of the AppliedWork object
+			// OwnerReference in the Namespace object (https://book.kubebuilder.io/reference/envtest.html#testing-considerations).
+			checkNSOwnerReferences(workName, nsName)
+
+			// Ensure that the AppliedWork object has been removed.
```
weng271190436, line 10008: if the AppliedWork object is supposed to be gone at this point. And you check the owner reference of the namespace right before, which includes getting the applied work from memberClient1.

What happened between `checkNSOwnerReferences(workName, nsName)` and `appliedWorkRemovedActual := appliedWorkRemovedActual(workName, nsName)` that caused the AppliedWork to be removed between these 2 statements?

pkg/controllers/workapplier/controller_integration_test.go
```diff
@@ -9631,3 +9642,412 @@ var _ = Describe("negative cases", func() {
...
+				if err := memberClient1.Status().Update(ctx, deploy); err != nil {
+					return fmt.Errorf("failed to update the Deployment status: %w", err)
+				}
+				return nil
+			}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to refresh the status of the Deployment")
+		})
+
+		It("should back-report refreshed deployment status to the Work object", func() {
+			Eventually(func() error {
+				workObj := &fleetv1beta1.Work{}
```
weng271190436: is this chunk of code exactly the same as the chunk that starts at line 9820?

pkg/controllers/workapplier/controller_integration_test.go
weng271190436: Think it should be (-want, +got) because you put want as the first argument of cmp.Diff

weng271190436: Read a bit more. It seems that if you put `got` as the first argument then it is `(-got, +want)` otherwise it is `(-want, +got)`?

It would be nice if we agree to stick to one way out of two

pkg/controllers/workapplier/status.go
ryanzhang-oss: raise alert

pkg/controllers/workapplier/status.go
```diff
@@ -151,6 +156,15 @@ func (r *Reconciler) refreshWorkStatus(
 			}
 		}
 
+		// Back-report the status from the member cluster side, if applicable.
+		//
+		// Back-reporting is only performed when:
+		// a) the ReportBackStrategy is of the type Mirror; and
+		// b) the manifest object has been applied successfully.
+		if isStatusBackReportingOn && isManifestObjectApplied(bundle.applyOrReportDiffResTyp) {
+			backReportStatus(bundle.inMemberClusterObj, manifestCond, now, klog.KObj(work))
```
ryanzhang-oss: can we get into situations that the combined resource status size is over the etcd limit?

pkg/controllers/workapplier/status.go
michaelawyu: Hi Wei! As Ryan mentioned, we have this wrapper added a long while ago so that the log collector can classify errors as needed. This was set to not return the wrapped error as the method itself does not return any errors; the underscore was added to silence the linter.

pkg/controllers/workapplier/controller_integration_test.go
michaelawyu: Hi Wei! Yeah, as you mentioned, the `got/want` order depends on the order we passed in the objects to compare. I don't think we have any form of agreement/rule on this though as semantics wise things remain the same...

pkg/controllers/workapplier/controller_integration_test.go
michaelawyu: Hi Wei! This was kind of related to an earlier issue we discussed a while ago. The integration test uses the `envtest` package, which runs a somewhat trimmed down Kubernetes environment. Specifically in this environment, IIRC when we do the foreground deletion there is no controller to actually delete the child objects, and owner references will not be updated as well (even if the owner itself is already gone), which is the reason why the corrent code formation works. I am happy to explain this a bit further if needed -> Britania might have a bit more context though as she added this part of the code.

pkg/controllers/workapplier/controller_integration_test.go
michaelawyu: Ah, yeah; let me add a common block in this test spec and simplify things a bit. Thanks.

pkg/controllers/workapplier/status.go
michaelawyu: Ah, the 1(.5) MB size limit can be a concern, let me add some logic to double-check -> this can be fixed by further sharding but we do not have that protocol yet...

pkg/controllers/workapplier/controller_integration_test.go
weng271190436: My confusion is that this function

```go
func appliedWorkRemovedActual(workName, nsName string) func() error {
	return func() error {
		// Retrieve the AppliedWork object.
		appliedWork := &fleetv1beta1.AppliedWork{}
		if err := memberClient1.Get(ctx, client.ObjectKey{Name: workName}, appliedWork); err != nil {
			if errors.IsNotFound(err) {
				// The AppliedWork object has been deleted, which is expected.
				return nil
			}
			return fmt.Errorf("failed to retrieve the AppliedWork object: %w", err)
		}
		if !appliedWork.DeletionTimestamp.IsZero() && controllerutil.ContainsFinalizer(appliedWork, metav1.FinalizerDeleteDependents) {
			// The AppliedWork object is being deleted, but the finalizer is still present. Remove the finalizer as there
			// are no real built-in controllers in this test environment to handle garbage collection.
			controllerutil.RemoveFinalizer(appliedWork, metav1.FinalizerDeleteDependents)
			Expect(memberClient1.Update(ctx, appliedWork)).To(Succeed(), "Failed to remove the finalizer from the AppliedWork object")
		}
		return fmt.Errorf("appliedWork object still exists")
	}
}
```

doesn't delete anything unless it is the removal of the finalizer that resulted in a successful deletion?

pkg/controllers/workapplier/status.go
ryanzhang-oss: is the logic added?

michaelawyu: Hi Ryan! The logic has been added; let me rebase this PR.

pkg/controllers/workapplier/controller_integration_test.go
michaelawyu: Hi Wei! Yeah, the work applier, when seeing that a work object has been deleted, will attempt to remove its corresponding AppliedWork object.

pkg/controllers/workapplier/status.go
```diff
@@ -151,6 +156,15 @@ func (r *Reconciler) refreshWorkStatus(
...
+		}
+
 		// Tally the stats.
 		if isManifestObjectApplied(bundle.applyOrReportDiffResTyp) {
```
ryanzhang-oss, line 168: Nit:  the flow is a bit more smooth if we move below inside the  isManifestObjectApplied block
        if isStatusBackReportingOn {
			backReportStatus(bundle.inMemberClusterObj, manifestCond, now, klog.KObj(work))
		}

ryanzhang-oss (dismissed):

pkg/controllers/workapplier/status.go
michaelawyu: Done ✅

pkg/controllers/workapplier/controller_integration_test.go
michaelawyu: Done ✅

PR #329, opened 2025-11-13
feat: add status back-reporting supprt (2/, new controller)

michaelawyu:

This PR adds a new controller that enables mirroring of status between the hub cluster side and the member cluster side.

It is a part of a series of PRs to enable status back-reporting in KubeFleet.

I have:

- [x] Unit tests
- [x] Integration tests (limited) 

To control the PR size, additional integration tests will be submitted in separate PRs.


pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+		// Retrieve the RP object.
+		placementObj = &placementv1beta1.ResourcePlacement{}
+		if err := r.hubClient.Get(ctx, client.ObjectKey{Namespace: parentPlacementNamespace, Name: parentPlacementName}, placementObj); err != nil {
+			wrappedErr := fmt.Errorf("failed to retrieve RP object: %w", err)
+			return nil, controller.NewAPIServerError(true, wrappedErr)
+		}
+	}
+
+	// Validate the scheduling policy of the placement object.
+	schedulingPolicy := placementObj.GetPlacementSpec().Policy
```
weng271190436, line 215: Will you do these validation in the CRP validator when someone specifies the report back strategy? If so does it make sense to move some of these validations into the common validator package?

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	// The scheduling policy is valid for back-reporting statuses to original resources.
+	return placementObj, nil
+}
+
+// formatResourceIdentifier formats a ResourceIdentifier object to a string for keying purposes.
+//
+// The format in use is `[API-GROUP]/[API-VERSION]/[API-KIND]/[NAMESPACE]/[NAME]`, e.g., `/v1/Namespace//work`.
+func formatResourceIdentifier(resourceIdentifier *placementv1beta1.ResourceIdentifier) string {
```
weng271190436, line 261: this might be useful beyond this package? I am wondering if this belongs to the pkg/utils

similarly formatWorkResourceIdentifier might also belong there?

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+			wrappedErr := fmt.Errorf("failed to unmarshal back-reported status: %w", err)
+			klog.ErrorS(err, "Failed to unmarshal back-reported status", "work", workRef, "resourceIdentifier", resIdentifier)
+			errs[pieces] = wrappedErr
+			return
+		}
+
+		// Note that if the applied resource has a status sub-resource, it is usually safe for us to assume that
+		// the original resource should also have a status sub-resource of the same format.
+		unstructured.Object["status"] = statusWrapper["status"]
+		_, err = r.hubDynamicClient.Resource(gvr).Namespace(nsName).UpdateStatus(ctx, unstructured, metav1.UpdateOptions{})
```
weng271190436, line 167: wouldn't some controller in the hub cluster overwritten the status you reported from applied work?

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+	isResEnvelopedByIdStr := prepareIsResEnvelopedMap(placementObj)
+
+	// Back-report statuses to original resources.
+
+	// Prepare a child context.
+	// Cancel the child context anyway to avoid leaks.
+	childCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	errs := make([]error, len(work.Status.ManifestConditions))
+	doWork := func(pieces int) {
```
weng271190436: nit: I think this sounds more like an `index` rather than `pieces`. Curious what does pieces mean in the context?

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+	// type with the number of clusters set to 1.
+	//
+	// TO-DO (chenyu1): at this moment the controller will attempt to locate the CRP/RP object that is
+	// associated with a Work object by parsing its name; KubeFleet does have a label (`kubernetes-fleet.io/parent-CRP`)
+	// that tracks the name of the placement object associated with a Work object, however, for RP objects
+	// the name does not include the owner namespace, which makes it less useful. Once the label situation
+	// improves, the statuses back-reporter should read the label instead for CRP/RP tracking here.
+	placementObj, err := r.validatePlacementObjectForOriginalResourceStatusBackReporting(ctx, work)
+	if err != nil {
+		klog.ErrorS(err, "Failed to validate the placement object associated with the Work object for back-reporting statuses to original resources", "work", workRef)
```
weng271190436, line 86: Is this a user facing error that we should report using EventRecorder?

pkg/controllers/statusbackreporter/controller_integration_test.go
```diff
@@ -0,0 +1,448 @@
...
+			Kind:       "Namespace",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: nsName,
+		},
+	}
+)
+
+// createWorkObject creates a new Work object with the given name, manifests, and apply strategy.
```
weng271190436: Copy/paste error from workapplier/controller_integration_test.go? Should we consolidate the 2 functions?

pkg/controllers/statusbackreporter/controller_integration_test.go
```diff
@@ -0,0 +1,448 @@
...
+			Workload: placementv1beta1.WorkloadTemplate{
+				Manifests: manifests,
+			},
+			ReportBackStrategy: reportBackStrategy,
+		},
+	}
+	Expect(hubClient.Create(ctx, work)).To(Succeed())
+}
+
+func marshalK8sObjJSON(obj runtime.Object) []byte {
```
weng271190436, line 80: The exact function exists in workapplier/controller_integration_test.go. Shall we consolidate?

pkg/controllers/statusbackreporter/controller_integration_test.go
```diff
@@ -0,0 +1,448 @@
...
+	statusBackReportingWrapper["kind"] = unstructuredObj.GetKind()
+	statusBackReportingWrapper["status"] = unstructuredObj.Object["status"]
+	statusBackReportingWrapperData, err := json.Marshal(statusBackReportingWrapper)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal status back-reporting wrapper data: %w", err)
+	}
+	return statusBackReportingWrapperData, nil
+}
+
+func workObjectRemovedActual(workName string) func() error {
```
weng271190436: also a duplicated utility from workapplier/controller_integration_test.go

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	// Prepare a map for quick lookup of whether a resource is enveloped.
+	isResEnvelopedByIdStr := prepareIsResEnvelopedMap(placementObj)
+
+	// Back-report statuses to original resources.
+
+	// Prepare a child context.
+	// Cancel the child context anyway to avoid leaks.
+	childCtx, cancel := context.WithCancel(ctx)
```
weng271190436, line 101: why create a childCtx instead of just passing ctx into ParallelizeUntil?

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+		return nil, controller.NewUserError(wrappedErr)
+	}
+
+	// The scheduling policy is valid for back-reporting statuses to original resources.
+	return placementObj, nil
+}
+
+// formatResourceIdentifier formats a ResourceIdentifier object to a string for keying purposes.
+//
+// The format in use is `[API-GROUP]/[API-VERSION]/[API-KIND]/[NAMESPACE]/[NAME]`, e.g., `/v1/Namespace//work`.
```
weng271190436: in the example `/v1/Namespace//work` you don't have API-GROUP

pkg/controllers/statusbackreporter/controller_integration_test.go
```diff
@@ -0,0 +1,448 @@
...
+		},
+	}
+	Expect(hubClient.Delete(ctx, work)).To(Succeed(), "Failed to delete the Work object")
+
+	workObjRemovedActual := workObjectRemovedActual(workName)
+	Eventually(workObjRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove work object")
+}
+
+var _ = Describe("back-reporting status", func() {
+	Context("back-report status for deployments (CRP)", Ordered, func() {
```
weng271190436, line 118: do we need a test for RP? Or is their code paths the same with regards to this feature so don't need to test RP separately?

pkg/controllers/statusbackreporter/suite_test.go
```diff
@@ -0,0 +1,142 @@
...
+	statusBackReporter *Reconciler
+
+	ctx    context.Context
+	cancel context.CancelFunc
+)
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecs(t, "Work Applier Integration Test Suite")
```
weng271190436: copy paste error from Work Applier integration test

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+			klog.ErrorS(err, "Failed to update status to the target resource", "work", workRef, "resourceIdentifier", resIdentifier)
+			errs[pieces] = wrappedErr
+			return
+		}
+	}
+	r.parallelizer.ParallelizeUntil(childCtx, len(work.Status.ManifestConditions), doWork, "backReportStatusToOriginalResources")
+	return ctrl.Result{}, errorsutil.NewAggregate(errs)
+}
+
+// validatePlacementObjectForOriginalResourceStatusBackReporting validatess whether
```
weng271190436: nit: validatess typo

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Hi Wei! We are switching to CEL-based validation as I understand (a separate PR will be prepared to add the CEL rule for this part). The CRP validator methods are currently only used for the webhooks as I understand.

michaelawyu: Though honestly speaking I do not have a very strong opinion on this -> adding the logic to the webhook part as well definitely helps, though because of the reason that webhooks are not 100% bullet-proof we still need to do some sanity check here.

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Thanks; let me fix that. -> Done.

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Yeah, I guess... though at this moment I think only this controller and the work applier need the formatter (the work applier only uses the work resource identifier one); and for the work applier the logic is bit different (it needs to handle cases where manifests cannot be decoded).

pkg/controllers/statusbackreporter/controller_integration_test.go
michaelawyu: Yeah, let me fix the comment. Thanks!

michaelawyu: For the consolidation part, we do not have a test utilities pkg yet and due to the usage of Gomega/Ginkgo deps such pkgs are going to look a bit tricky -> allow me a bit more time to think this through if it is OK

pkg/controllers/statusbackreporter/controller_integration_test.go
michaelawyu: Hi Wei! Yeah, we do need tests for the RP API as well. As mentioned the in PR notes not all tests are checked in in this PR to control the PR size -> I will submit them when this one is merged first.

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Hi Wei! A bit of background on this (and apologies, there should have been a TO-DO comment added in the PR; I forgot about that when migrating the code) ->

For the Azure Fleet Manager environment, we disabled a large number of built-in K8s controllers, so API objects like Deployments, StatefulSets, etc. are not going to have status populated at all. For the upstream installations, we recommend that users follow the same pattern, though this is not mandated.

We do need a whitelist kind of mechanism so that users can specify which APIs we/they know are safe for status back-reporting. As it is a logic component on its own it's not included in this PR (and that's partly the reason why this controller is not yet enabled) -> I will submit them separately with proper E2E tests.

michaelawyu: If there's any confusion/concern, please let me know 🙏

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Hi Wei! For this method there are some errors (e.g., API GET errors) that are not user-facing; and for the sanity check part the plan is to have CEL expressions in place so users are going to see errors when they create the CRP object.

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Hi Wei! The child context is created to allow this parallelization to be cancelled without impacting the main context (the reconciliation loop context) (and the parent context's cancellation will still propagate to the child context). Though I totally understand your concern that currently the cancel method is not invoked explicitly...

pkg/controllers/statusbackreporter/suite_test.go
michaelawyu: Thanks! Let me fix that. -> Done.

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+		klog.V(2).InfoS("Reconciliation loop ends", "controller", "statusBackReporter", "work", workRef, "latency", latency)
+	}()
+
+	work := &placementv1beta1.Work{}
+	if err := r.hubClient.Get(ctx, req.NamespacedName, work); err != nil {
+		klog.ErrorS(err, "Failed to retrieve Work object", "work", workRef)
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// TO-DO (chenyu1): verify status back-reporting strategy from the source (CRP or RP objects),
```
ryanzhang-oss: There is a way to find a work CR's placement parent?

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+	ctx context.Context, work *placementv1beta1.Work) (placementv1beta1.PlacementObj, error) {
+	// Read the `kubernetes-fleet.io/parent-CRP` label to retrieve the CRP/RP name.
+	parentPlacementName, ok := work.Labels[placementv1beta1.PlacementTrackingLabel]
+	if !ok || len(parentPlacementName) == 0 {
+		// Normally this should never occur.
+		wrappedErr := fmt.Errorf("the placement tracking label is absent or invalid (label value: %s)", parentPlacementName)
+		return nil, controller.NewUnexpectedBehaviorError(wrappedErr)
+	}
+
+	// Parse the name of the Work object to retrieve the namespace of the placement object (if applicable).
```
ryanzhang-oss: work has a "ParentNamespaceLabel" on the work

pkg/controllers/statusbackreporter/controller.go
ryanzhang-oss, line 167: I wonder this will trigger resource change dector to trigger another placement reconcile?

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+			return
+		}
+
+		// Note that if the applied resource has a status sub-resource, it is usually safe for us to assume that
+		// the original resource should also have a status sub-resource of the same format.
+		unstructured.Object["status"] = statusWrapper["status"]
+		_, err = r.hubDynamicClient.Resource(gvr).Namespace(nsName).UpdateStatus(ctx, unstructured, metav1.UpdateOptions{})
+		if err != nil {
+			// TO-DO (chenyu1): check for cases where the API definition is inconsistent between the member cluster
+			// side and the hub cluster side, and single out the errors as user errors instead.
```
ryanzhang-oss, line 170: good point,  it seems that we probably should start small? like only start with job/deployment i.e. something with stable API.

pkg/controllers/statusbackreporter/controller.go
```diff
@@ -0,0 +1,309 @@
...
+		if err := json.Unmarshal(manifestCond.BackReportedStatus.ObservedStatus.Raw, &statusWrapper); err != nil {
+			wrappedErr := fmt.Errorf("failed to unmarshal back-reported status: %w", err)
+			klog.ErrorS(err, "Failed to unmarshal back-reported status", "work", workRef, "resourceIdentifier", resIdentifier)
+			errs[pieces] = wrappedErr
+			return
+		}
+
+		// Note that if the applied resource has a status sub-resource, it is usually safe for us to assume that
+		// the original resource should also have a status sub-resource of the same format.
+		unstructured.Object["status"] = statusWrapper["status"]
```
ryanzhang-oss, line 166: do we not check the GKV that we put in the statusWrapper?

pkg/controllers/statusbackreporter/controller.go
weng271190436: I think it will trigger a change detector action but will be ignored by placement controller when it finds out that there is no meaningful diff

pkg/controllers/statusbackreporter/controller.go
michaelawyu: I will add a TO-DO comment to note the possible factoring opportunity should the situation materializes.

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Hi Ryan! The GVK/GVR we put in the wrapper are there only for completeness reasons (otherwise the serialization/deserialization would fail) -> the info are directly copied from the work resource identifier, which we already use in this controller.

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Hi Ryan! Yeah, there will be a whitelist based mechanism that controls which APIs can have their status back-reported; as mentioned to Wei earlier this will be submitted as a separate PR.

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Yeah, it will trigger the placement controller for once. We could update the `onResourceUpdated` method, but it would need a deep equal (current the method uses resource versions), which might be a bit too resource-consuming considering that this method is on the critical path? I could add a TO-DO here just in case we need to evaluate this possibility.

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Ah, I forgot about this; let me update the logic.

pkg/controllers/statusbackreporter/controller.go
michaelawyu: Will make the edits.

pkg/controllers/statusbackreporter/controller_integration_test.go
michaelawyu: Made the edits, thanks 🙏

pkg/controllers/statusbackreporter/controller_integration_test.go
michaelawyu: I made some edits that do some wrapping; it's a bit overkilling but I think it is better than introducing Gomega pkg into regular code.

pkg/controllers/statusbackreporter/controller_integration_test.go
michaelawyu: I moved them to a separate pkg though I have to say I have some minor reservation about this. It's nothing major, obviously, but this feels like adding another layer of redirection just for the sake of adding redirection. It just brings so little benefits (not really making the code more maintainable/readable) and at this moment does not seem to be a well-used pattern.

test/utils/actuals/actuals.go
```diff
@@ -0,0 +1,44 @@
+/*
```
ryanzhang-oss, line 1: why adding this file here? We have all the actuals in another file. Is the plan to move them all here?

michaelawyu: Hi Ryan! The actuals are currently kept in `_test.go` files, which cannot be imported in another test file. It's related to the discussion here: https://github.com/kubefleet-dev/kubefleet/pull/329#discussion_r2631422623.

There might be more actuals that can be shared, but I am not 100% sure if this is really a sustainable pattern (sometimes it's OK to duplicate things?).

michaelawyu: Anyway, I will merge this first to unblock progress; if there's any concern or suggestion for a better solution, please let me know -> I could submit separate PRs.

PR #343, opened 2025-11-20
fix: abandon the work applier reconciliation loop when the main context exits

michaelawyu:

This PR sets up the work applier to abandon the reconciliation loop when the main context exits in a proper way so that partially process manifests will not trigger error unexpectedly.

I have:

N/A

N/A

### DIFF
--- a/pkg/utils/parallelizer/parallelizer.go
+++ b/pkg/utils/parallelizer/parallelizer.go
@@ -60,7 +60,7 @@ func (p *parallelizer) ParallelizeUntil(ctx context.Context, pieces int, doWork 
 	// Note (chenyu1): the ParallelizeUntil method is essentially a thin wrapper around the
 	// workqueue.ParallelizeUntil method. Note that the workqueue.ParallelizeUntil method
 	// right now does not return any error; it returns when the context is cancelled, possibly
-	// in a willingly manner. Some of the KubeFleet code makes use of this to faciliate a
+	// in a willingly manner. Some of the KubeFleet code makes use of this to facilitate a
 	// fail-fast pattern (i.e., pass in a child context to the parallelizer; if one worker
 	// has exited, cancel the child context in the worker and consequently the whole parallelization).
 	// As only the caller knows why a context is cancelled (willingly by a worker or not), we leave it to the


pkg/controllers/workapplier/availability_tracker.go
```diff
@@ -83,6 +83,15 @@ func (r *Reconciler) trackInMemberClusterObjAvailability(ctx context.Context, bu
 
 	// Run the availability check in parallel.
 	r.parallelizer.ParallelizeUntil(childCtx, len(bundles), doWork, "trackInMemberClusterObjAvailability")
+
+	// The workqueue.ParallelizeUntil utility does not return errors even if its context has been
+	// cancelled (and some availability checks might not be completed yet); to catch such
+	// premature termination, the work applier checks for context cancellation directly.
+	if err := ctx.Err(); err != nil {
```
weng271190436, line 92: Why not do this check in framework.go and preprocess.go?

If this is applicable everywhere, maybe do this check and return error from 

func (p *parallelizer) ParallelizeUntil(ctx context.Context, pieces int, doWork workqueue.DoWorkPieceFunc, operation string)

whenever ctx is cancelled?

jwtty: +1, ParallelizeUtil is an interface we own. We can probably return error from there.

weng271190436: The reason that I raised this originally is "can we make it easier for the next person using ParallelizeUntil so that they won't forget to check for context cancellation if we can encapsulate the handling inside the parallelizer package." But it seems that there are some nuances so okay with this approach.

For example, in framework.go we are already handling the context cancellation error separately

michaelawyu: Hi Wei and Wantong! Yeah, actually that was my first attempt as well. The concern I had was that, there are cases where we run `ParallelizeUntil` in its own child context (primarily for the reason that the child context can be cancelled on its own to terminate the parallelization promptly); and if we do the checking with our wrapper, e.g., 

```go
func (p *parallelizer) ParallelizeUntil(ctx context.Context, pieces int, doWork workqueue.DoWorkPieceFunc, operation string) {
	...
	workqueue.ParallelizeUntil(ctx, p.numOfWorkers, pieces, doWorkWithLogs)

        if err := ctx.Done(); err != nil {...}
}
```

two possibilities exist:

a) the child context is cancelled willingly by us, and the method returns an error;
b) the parent context gets cancelled by factors outside our control, which in turn cancels the child context, and we also receive an error

The complication is that, we couldn't really tell between the two possibilities by looking at the error, so we still need to inspect the parent context at the caller level to find out why things fail, which kind of defeats the purpose of putting the logic inside the wrapper. This is why in this PR the check is done at the caller level instead at the moment. If there's any concern or things that I've missed, please let me know 🙏

PR #346, opened 2025-11-25
feat: implement initialize and execute states within update run

britaniar:

I have:
- Update state variable names to match the value.
- Update the update run controller to use State from the Spec. Implementation mostly is for update runs in Initialize state or Execute state or moving from Initialize state to Execute state.
- Updated conditional checks to disregard generation checks as update run is immutable but the State field.
- Added/update integrations tests and e2e tests.

- Integration Tests
- E2E Test

This change covers when update run starts as state `Initialize` then to `Execute` or starts at state `Execute`.
The initialized condition should be successfully initialized once. 

Generation will change each time update run state is updated and only when the state is updated. The rest of update run spec is immutable. Can correlate generation with state change.

### DIFF
--- a/apis/placement/v1beta1/stageupdate_types.go
+++ b/apis/placement/v1beta1/stageupdate_types.go
@@ -152,14 +152,14 @@ func (c *ClusterStagedUpdateRun) SetUpdateRunStatus(status UpdateRunStatus) {
 type State string
 
 const (
-	// StateInitialize describes user intent to initialize but not execute the update run.
+	// StateInitialized describes user intent to initialize but not execute the update run.
 	// This is the default state when an update run is created.
 	// Users can subsequently set the state to Execute or Abandon.
-	StateInitialize State = "Initialize"
+	StateInitialized State = "Initialize"
 
-	// StateExecute describes user intent to execute (or resume execution if paused).
+	// StateExecuted describes user intent to execute (or resume execution if paused).
 	// Users can subsequently set the state to Pause or Abandon.
-	StateExecute State = "Execute"
+	StateExecuted State = "Execute"
 
 	// StateStopped describes user intent to pause the update run.
 	// Users can subsequently set the state to Execute or Abandon.
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -159,7 +159,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	}
 
 	// Execute the updateRun.
-	if state == placementv1beta1.StateExecute {
+	if state == placementv1beta1.StateExecuted {
 		klog.V(2).InfoS("Continue to execute the updateRun", "state", state, "updatingStageIndex", updatingStageIndex, "updateRun", runObjRef)
 		finished, waitTime, execErr := r.execute(ctx, updateRun, updatingStageIndex, toBeUpdatedBindings, toBeDeletedBindings)
 		if errors.Is(execErr, errStagedUpdatedAborted) {
--- a/pkg/controllers/updaterun/execution_integration_test.go
+++ b/pkg/controllers/updaterun/execution_integration_test.go
@@ -1518,11 +1518,11 @@ var _ = Describe("UpdateRun execution tests - single stage", func() {
 		})
 	})
 
-	Context("Cluster staged update run should update clusters one by one - different states (Initialized -> Execute)", Ordered, func() {
+	Context("Cluster staged update run should update clusters one by one - different states (Initialized -> Executed)", Ordered, func() {
 		var wantMetrics []*promclient.Metric
 		BeforeAll(func() {
 			By("Creating a new clusterStagedUpdateRun")
-			updateRun.Spec.State = placementv1beta1.StateInitialize
+			updateRun.Spec.State = placementv1beta1.StateInitialized
 			Expect(k8sClient.Create(ctx, updateRun)).To(Succeed())
 
 			By("Validating the initialization succeeded and but not execution started")
@@ -1552,7 +1552,7 @@ var _ = Describe("UpdateRun execution tests - single stage", func() {
 
 		It("Should start execution after changing the state to Execute", func() {
 			By("Updating the updateRun state to Execute")
-			updateRun.Spec.State = placementv1beta1.StateExecute
+			updateRun.Spec.State = placementv1beta1.StateExecuted
 			Expect(k8sClient.Update(ctx, updateRun)).Should(Succeed(), "failed to update the updateRun state")
 
 			By("Validating the execution has started")
--- a/test/e2e/cluster_staged_updaterun_test.go
+++ b/test/e2e/cluster_staged_updaterun_test.go
@@ -314,7 +314,7 @@ var _ = Describe("test CRP rollout with staged update run", func() {
 		})
 
 		It("Should create a cluster staged update run successfully", func() {
-			createClusterStagedUpdateRunSucceed(updateRunNames[0], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecute)
+			createClusterStagedUpdateRunSucceed(updateRunNames[0], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecuted)
 		})
 
 		It("Should rollout resources to member-cluster-2 only and complete stage canary", func() {
@@ -378,7 +378,7 @@ var _ = Describe("test CRP rollout with staged update run", func() {
 		})
 
 		It("Should create a new cluster staged update run successfully", func() {
-			createClusterStagedUpdateRunSucceed(updateRunNames[1], crpName, resourceSnapshotIndex2nd, strategyName, placementv1beta1.StateExecute)
+			createClusterStagedUpdateRunSucceed(updateRunNames[1], crpName, resourceSnapshotIndex2nd, strategyName, placementv1beta1.StateExecuted)
 		})
 
 		It("Should rollout resources to member-cluster-2 only and complete stage canary", func() {
@@ -426,7 +426,7 @@ var _ = Describe("test CRP rollout with staged update run", func() {
 		})
 
 		It("Should create a new staged update run with old resourceSnapshotIndex successfully to rollback", func() {
-			createClusterStagedUpdateRunSucceed(updateRunNames[2], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecute)
+			createClusterStagedUpdateRunSucceed(updateRunNames[2], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecuted)
 		})
 
 		It("Should rollback resources to member-cluster-2 only and completes stage canary", func() {
@@ -539,7 +539,7 @@ var _ = Describe("test CRP rollout with staged update run", func() {
 		})
 
 		It("Should create a cluster staged update run successfully", func() {
-			createClusterStagedUpdateRunSucceed(updateRunNames[0], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecute)
+			createClusterStagedUpdateRunSucceed(updateRunNames[0], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecuted)
 		})
 
 		It("Should rollout resources to member-cluster-2 only and complete stage canary", func() {
@@ -593,7 +593,7 @@ var _ = Describe("test CRP rollout with staged update run", func() {
 		})
 
 		It("Should create a cluster staged update run successfully", func() {
-			createClusterStagedUpdateRunSucceed(updateRunNames[1], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecute)
+			createClusterStagedUpdateRunSucceed(updateRunNames[1], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecuted)
 		})
 
 		It("Should still have resources on member-cluster-1 and member-cluster-2 only and completes stage canary", func() {
@@ -648,7 +648,7 @@ var _ = Describe("test CRP rollout with staged update run", func() {
 		})
 
 		It("Should create a cluster staged update run successfully", func() {
-			createClusterStagedUpdateRunSucceed(updateRunNames[2], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecute)
+			createClusterStagedUpdateRunSucceed(updateRunNames[2], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecuted)
 		})
 
 		It("Should still have resources on all member clusters and complete stage canary", func() {
@@ -743,7 +743,7 @@ var _ = Describe("test CRP rollout with staged update run", func() {
 		})
 
 		It("Should create a cluster staged update run successfully", func() {
-			createClusterStagedUpdateRunSucceed(updateRunNames[0], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecute)
+			createClusterStagedUpdateRunSucceed(updateRunNames[0], crpName, resourceSnapshotIndex1st, strategyName, placementv1beta1.StateExecuted)
 		})
 
 		It("Should not rollout any resources to member clusters and complete stage canary", func() {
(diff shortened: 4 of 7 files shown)


pkg/controllers/updaterun/controller.go
```diff
@@ -104,11 +104,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
+	// Early check for abandoned state - this is a terminal state, no initialization needed.
+	state := updateRun.GetUpdateRunSpec().State
+	if state == placementv1beta1.StateAbandoned {
+		klog.V(2).InfoS("The updateRun is abandoned, terminating", "state", state, "updateRun", runObjRef)
+		return runtime.Result{}, r.recordUpdateRunAbandoned(ctx, updateRun)
+	} else if state == placementv1beta1.StateStopped { // Early check for stopped state - pause the update run if needed.
+		klog.V(2).InfoS("The updateRun is stopped, waiting to resume", "state", state, "updateRun", runObjRef)
+		return runtime.Result{}, r.recordUpdateRunPaused(ctx, updateRun)
+	}
+
```
ryanzhang-oss: nit: use switch to replace if/else
```suggestion
	switch state {
	case placementv1beta1.StateAbandoned:
		klog.V(2).InfoS("The updateRun is abandoned, terminating", "state", state, "updateRun", runObjRef)
		return runtime.Result{}, r.recordUpdateRunAbandoned(ctx, updateRun)
	case placementv1beta1.StateStopped: // Early check for stopped state - pause the update run if needed.
		klog.V(2).InfoS("The updateRun is stopped, waiting to resume", "state", state, "updateRun", runObjRef)
		return runtime.Result{}, r.recordUpdateRunPaused(ctx, updateRun)
	}
```

pkg/controllers/updaterun/controller.go
```diff
@@ -265,6 +285,50 @@ func (r *Reconciler) recordUpdateRunFailed(ctx context.Context, updateRun placem
...
+		Message:            "The update run has been abandoned",
+	})
+
+	if updateErr := r.Client.Status().Update(ctx, updateRun); updateErr != nil {
+		klog.ErrorS(updateErr, "Failed to update the updateRun status as failed", "updateRun", klog.KObj(updateRun))
+		// updateErr can be retried.
+		return controller.NewUpdateIgnoreConflictError(updateErr)
+	}
+	return nil
+}
```
ryanzhang-oss: We need to update the status of each cluster as those could be in the middle of upgrading.

pkg/controllers/updaterun/controller.go
```diff
@@ -151,28 +167,32 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
+		// The execution is not finished yet or it encounters a retriable error.
+		// We need to record the status and requeue.
+		if updateErr := r.recordUpdateRunStatus(ctx, updateRun); updateErr != nil {
+			return runtime.Result{}, updateErr
+		}
+		klog.V(2).InfoS("The updateRun is not finished yet", "requeueWaitTime", waitTime, "execErr", execErr, "updateRun", runObjRef)
+		if execErr != nil {
+			return runtime.Result{}, execErr
+		}
+		return runtime.Result{Requeue: true, RequeueAfter: waitTime}, nil
```
ryanzhang-oss, line 184: Requeue is deprecated

pkg/controllers/updaterun/execution.go
```diff
@@ -241,29 +186,128 @@ func (r *Reconciler) executeUpdatingStage(
...
+	clusterStatus *placementv1beta1.ClusterUpdatingStatus,
+	resourceSnapshotName string,
+	updateRun placementv1beta1.UpdateRunObj,
+	updatingStageStatus *placementv1beta1.StageUpdatingStatus,
+) error {
+	updateRunRef := klog.KObj(updateRun)
+	updateRunStatus := updateRun.GetUpdateRunStatus()
+
+	if !isBindingSyncedWithClusterStatus(resourceSnapshotName, updateRun, binding, clusterStatus) {
+		klog.V(2).InfoS("Found the first cluster that needs to be updated", "cluster", clusterStatus.ClusterName, "stage", updatingStageStatus.StageName, "updateRun", updateRunRef)
```
ryanzhang-oss: why is this. the first cluster?

test/e2e/cluster_staged_updaterun_test.go
```diff
@@ -147,10 +148,6 @@ var _ = Describe("test CRP rollout with staged update run", func() {
 			validateAndApproveClusterApprovalRequests(updateRunNames[0], envCanary)
 		})
 
-		It("Should rollout resources to member-cluster-1 first because of its name", func() {
-			checkIfPlacedWorkResourcesOnMemberClustersInUpdateRun([]*framework.Cluster{allMemberClusters[0]})
-		})
```
ryanzhang-oss, line 152: why removing this?

britaniar: This is removed in my other PR as well. Wantong said we don't need anymore.

pkg/controllers/updaterun/controller.go
britaniar: Will address stopped and Abandoned in separate PRs

pkg/controllers/updaterun/controller.go
```diff
@@ -104,15 +104,29 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
 		if condition.IsConditionStatusFalse(initCond, updateRun.GetGeneration()) {
 			klog.V(2).InfoS("The updateRun has failed to initialize", "errorMsg", initCond.Message, "updateRun", runObjRef)
 			return runtime.Result{}, nil
 		}
+
+		if initCond == nil {
+			// Update the status to indicate that the updateRun is initializing.
+			// Requeue immediately to continue with initialization.
+			klog.V(2).InfoS("The updateRun is initializing", "state", state, "updateRun", runObjRef)
+			return runtime.Result{RequeueAfter: utils.DefaultRequeueAfterDuration}, r.recordUpdateRunInitializing(ctx, updateRun)
```
Arvindthiru: Why do we return here ?, if recordUpdateRunInitializing succeeds we can proceed

pkg/controllers/updaterun/execution_integration_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	io_prometheus_client "github.com/prometheus/client_model/go"
```
Arvindthiru:
```suggestion
	promclient "github.com/prometheus/client_model/go"
```

pkg/controllers/updaterun/execution_integration_test.go
```diff
@@ -1304,6 +1305,122 @@ var _ = Describe("UpdateRun execution tests - single stage", func() {
 			}, timeout, interval).Should(BeTrue(), "failed to ensure the approvalRequest is not recreated")
 		})
 	})
+
+	Context("Cluster staged update run should update clusters one by one - different states (Initialized -> Execute)", Ordered, func() {
+		var wantMetrics []*io_prometheus_client.Metric
+		BeforeAll(func() {
+			By("Creating a new clusterStagedUpdateRun")
+			updateRun.Spec.State = placementv1beta1.StateNotStarted
```
Arvindthiru: Could we also update the var names to match initialize, execute as part of this PR ?

test/e2e/actuals_test.go
```diff
@@ -2126,6 +2141,112 @@ func clusterStagedUpdateRunStatusSucceededActual(
...
+				}
+			}
+		}
+		wantStatus.StagesStatus = stagesStatus
+		wantStatus.DeletionStageStatus = buildDeletionStatusWithoutConditions(wantUnscheduledClusters, updateRun)
+		wantStatus.Conditions = updateRunInitializedConditions(updateRun.Generation)
+		if diff := cmp.Diff(updateRun.Status, wantStatus, updateRunStatusCmpOption...); diff != "" {
+			return fmt.Errorf("UpdateRun status diff (-got, +want): %s", diff)
+		}
+		return nil
```
Arvindthiru: This part can be moved to a common func for reuse

pkg/controllers/updaterun/controller.go
ryanzhang-oss: I am not following. Why do we need this extra step instead of just calling initialize?

pkg/controllers/updaterun/execution.go
```diff
@@ -418,7 +419,7 @@ func (r *Reconciler) checkAfterStageTasksStatus(ctx context.Context, updatingSta
 		}
 	}
 	if passed {
-		afterStageWaitTime = 0
+		afterStageWaitTime = utils.DefaultRequeueAfterDuration
```
ryanzhang-oss: the original 0 means the controller runtime will do some backoff retry, I am not sure what's the current behavior when we put a requeue time

pkg/controllers/updaterun/controller.go
britaniar: I was thinking we update the statues to show the intermediate state (-ing), because if we go straight to initialize then we will go straight to failed or initialed (-ed)

pkg/controllers/updaterun/execution.go
britaniar: image
Currently, even without "Requeue" it still behaves how it previously did (shown above).

pkg/controllers/updaterun/controller.go
britaniar: Removed per discussion offline that this brings no benefit with how fast the initialization is.

pkg/controllers/updaterun/controller.go
```diff
@@ -151,28 +159,32 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
+		// The execution is not finished yet or it encounters a retriable error.
+		// We need to record the status and requeue.
+		if updateErr := r.recordUpdateRunStatus(ctx, updateRun); updateErr != nil {
+			return runtime.Result{}, updateErr
+		}
+		klog.V(2).InfoS("The updateRun is not finished yet", "requeueWaitTime", waitTime, "execErr", execErr, "updateRun", runObjRef)
+		if execErr != nil {
+			return runtime.Result{}, execErr
+		}
+		return runtime.Result{Requeue: true, RequeueAfter: waitTime}, nil
```
Arvindthiru, line 184: If RequeueAfter is specified, Requeue doesn't matter

britaniar: This is the original code. It is fine if we have both no need to separate. If waitTime is >0, requeue will be ignored. If waitTime is 0 then it will take a look at the requeue to do rate limiting.

pkg/controllers/updaterun/controller.go
```diff
@@ -104,15 +104,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
 	var toBeUpdatedBindings, toBeDeletedBindings []placementv1beta1.BindingObj
 	updateRunStatus := updateRun.GetUpdateRunStatus()
 	initCond := meta.FindStatusCondition(updateRunStatus.Conditions, string(placementv1beta1.StagedUpdateRunConditionInitialized))
-	if !condition.IsConditionStatusTrue(initCond, updateRun.GetGeneration()) {
+	// Check if initialized regardless of generation.
+	// The updateRun spec fields are immutable except for the state field. When the state changes,
+	// the update run generation increments, but we don't need to reinitialize since initialization is a one-time setup.
+	if !(initCond != nil && initCond.Status == metav1.ConditionTrue) {
+		// Check if initialization failed for the current generation.
 		if condition.IsConditionStatusFalse(initCond, updateRun.GetGeneration()) {
```
Arvindthiru: If initialize fails once and then if user sets state to execute, We would re-initialize since IsConditionStatusFalse returns false based on the new generation ?

I think we need to ignore generation here if initCondition was set to false

pkg/controllers/updaterun/controller.go
```diff
@@ -151,28 +159,32 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
+			return runtime.Result{}, updateErr
+		}
+		klog.V(2).InfoS("The updateRun is not finished yet", "requeueWaitTime", waitTime, "execErr", execErr, "updateRun", runObjRef)
+		if execErr != nil {
+			return runtime.Result{}, execErr
+		}
+		return runtime.Result{Requeue: true, RequeueAfter: waitTime}, nil
 	}
-	return runtime.Result{Requeue: true, RequeueAfter: waitTime}, nil
+	klog.V(2).InfoS("The updateRun is not started, waiting to be started", "state", state, "updateRun", runObjRef)
```
Arvindthiru:
```suggestion
	klog.V(2).InfoS("The updateRun is initialized but not executed, waiting to execute", "state", state, "updateRun", runObjRef)
```

pkg/utils/condition/reason.go
```diff
@@ -152,6 +152,9 @@ const (
 
 // A group of condition reason string which is used to populate the ClusterStagedUpdateRun condition.
 const (
+	// UpdateRunInitializingReason is the reason string of condition if the update run is currently initializing.
+	UpdateRunInitializingReason = "UpdateRunInitializing"
```
Arvindthiru: do we need this ?

pkg/utils/common.go
```diff
@@ -54,6 +54,15 @@ import (
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/informer"
 )
 
+const (
+	// DefaultRequeueAfterDuration is the default duration after which to requeue a reconcile request.
+	// This is used when a controller wants to requeue immediately. A small duration is used to mimic immediate requeue
+	// that Controller-Runtimes deprecated Requeue=true previously provided.
+	// For more details, see: https://github.com/kubernetes-sigs/controller-runtime/pull/3107
+	// The value needs to be small enough to avoid noticeable delay, but greater than 0 as RequeueAfter=0 is treated as no requeue.
+	DefaultRequeueAfterDuration = time.Microsecond * 1
```
Arvindthiru: Please remove

test/e2e/actuals_test.go
```diff
@@ -2161,6 +2248,40 @@ func stagedUpdateRunStatusSucceededActual(
 	}
 }
 
+func buildStageUpdatingStatusesForInitialized(
```
Arvindthiru, line 2193: buildStageUpdatingStatusesForInitialized, buildStageUpdatingStatuses are very similar only diff is addition of conditions where relevant. can we pass a bool like execute set to true/false to buildStageUpdatingStatuses to reuse code further?

We can do the same for buildDeletionStageStatus

test/e2e/cluster_staged_updaterun_test.go
```diff
@@ -2079,13 +2186,22 @@ func createClusterStagedUpdateRunSucceedWithNoResourceSnapshotIndex(updateRunNam
...
 		Spec: placementv1beta1.UpdateRunSpec{
+			State:                    placementv1beta1.StateExecute,
 			PlacementName:            crpName,
 			StagedUpdateStrategyName: strategyName,
 		},
 	}
 	Expect(hubClient.Create(ctx, updateRun)).To(Succeed(), "Failed to create ClusterStagedUpdateRun %s", updateRunName)
 }
 
+func updateClusterStagedUpdateRunState(updateRunName string, state placementv1beta1.State) {
```
Arvindthiru, line 2197: Please do the get, update inside an Eventually block. Please make the same changes for updateStagedUpdateRunState as well

test/e2e/staged_updaterun_test.go
```diff
@@ -1643,13 +1747,22 @@ func createStagedUpdateRunSucceedWithNoResourceSnapshotIndex(updateRunName, name
 			Namespace: namespace,
 		},
 		Spec: placementv1beta1.UpdateRunSpec{
+			State:                    placementv1beta1.StateExecute,
```
Arvindthiru: nit: why not pass the state as an parameter ?

britaniar: It is a case that is only used once.

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -152,13 +152,14 @@ func (c *ClusterStagedUpdateRun) SetUpdateRunStatus(status UpdateRunStatus) {
 type State string
 
 const (
-	// StateNotStarted describes user intent to initialize but not execute the update run.
+	// StateInitialize describes user intent to initialize but not execute the update run.
```
Arvindthiru: StateInitialized, StateExecuted

PR #347, opened 2025-11-25
feat: make hub statefulset work by stripping some properties from generated PVCs

weng271190436:

hub stateful set doesn't work when propagated to members because some properties were not stripped from generated PVCs

I have:

### DIFF
--- a/pkg/controllers/placement/resource_selector.go
+++ b/pkg/controllers/placement/resource_selector.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
-	pvutil "k8s.io/component-helpers/storage/volume"
 	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/util/deployment"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -445,18 +444,6 @@ func generateRawContent(object *unstructured.Unstructured) ([]byte, error) {
 		delete(annots, corev1.LastAppliedConfigAnnotation)
 		// Remove the revision annotation set by deployment controller.
 		delete(annots, deployment.RevisionAnnotation)
-		// Remove node-specific and provisioning-related annotations from PVCs that would break when propagated to member clusters
-		// These annotations reference specific nodes/provisioners from the hub cluster which don't exist on member clusters
-		// The member cluster's storage provisioner will set appropriate values for its own environment
-		// All annotations below are listed in well-known labels, annotations and taints document:
-		// https://kubernetes.io/docs/reference/labels-annotations-taints/
-		delete(annots, pvutil.AnnSelectedNode)           // Node selected for volume binding
-		delete(annots, pvutil.AnnBindCompleted)          // Binding completion status
-		delete(annots, pvutil.AnnBoundByController)      // Controller binding status
-		delete(annots, pvutil.AnnStorageProvisioner)     // Storage provisioner annotation
-		delete(annots, pvutil.AnnBetaStorageProvisioner) // Beta storage provisioner annotation
-		delete(annots, pvutil.AnnDynamicallyProvisioned) // Dynamically provisioned by annotation
-		delete(annots, pvutil.AnnMigratedTo)             // CSI migration annotation
 		if len(annots) == 0 {
 			object.SetAnnotations(nil)
 		} else {
@@ -504,10 +491,6 @@ func generateRawContent(object *unstructured.Unstructured) ([]byte, error) {
 			unstructured.RemoveNestedField(object.Object, "spec", "template", "metadata", "labels", "controller-uid")
 			unstructured.RemoveNestedField(object.Object, "spec", "template", "metadata", "labels", "batch.kubernetes.io/controller-uid")
 		}
-	} else if object.GetKind() == "PersistentVolumeClaim" && object.GetAPIVersion() == "v1" {
-		// Remove volumeName which references a specific PV from the hub cluster that won't exist on member clusters.
-		// The member cluster's storage provisioner will create and bind a new PV.
-		unstructured.RemoveNestedField(object.Object, "spec", "volumeName")
 	}
 
 	rawContent, err := object.MarshalJSON()
--- a/pkg/controllers/placement/resource_selector_test.go
+++ b/pkg/controllers/placement/resource_selector_test.go
@@ -27,14 +27,12 @@ import (
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
-	pvutil "k8s.io/component-helpers/storage/volume"
 	"k8s.io/kubectl/pkg/util/deployment"
 	"k8s.io/utils/ptr"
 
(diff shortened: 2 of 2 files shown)


pkg/controllers/placement/resource_selector.go
```diff
@@ -444,6 +445,15 @@ func generateRawContent(object *unstructured.Unstructured) ([]byte, error) {
...
 		delete(annots, deployment.RevisionAnnotation)
+		// Remove node-specific annotations from PVCs that would break when propagated to member clusters
+		// These annotations reference specific nodes from the hub cluster which don't exist on member clusters
+		// The member cluster's storage provisioner will set appropriate values for its own nodes
+		// All annotations below are listed in well-known labels, annotations and taints document:
+		// https://kubernetes.io/docs/reference/labels-annotations-taints/
+		delete(annots, pvutil.AnnSelectedNode)           // Node selected for volume binding
+		delete(annots, pvutil.AnnBindCompleted)          // Binding completion status
+		delete(annots, pvutil.AnnBoundByController)      // Controller binding status
+		delete(annots, pvutil.AnnBetaStorageProvisioner) // Beta storage provisioner annotation
```
ryanzhang-oss: there are more annotations like
```
AnnDynamicallyProvisioned
AnnMigratedTo
AnnSelectedNode
```
why not remove them all?

test/e2e/resource_placement_hub_workload_test.go
```diff
@@ -105,9 +109,30 @@ var _ = Describe("placing workloads using a CRP with PickAll policy", Label("res
...
+			{
+				Version:   "v1",
+				Kind:      "PersistentVolumeClaim",
+				Name:      fmt.Sprintf("%s-%s-%d", testStatefulSet.Spec.VolumeClaimTemplates[0].Name, testStatefulSet.Name, 1),
+				Namespace: workNamespace.Name,
+			},
 		}
 		// Use customizedPlacementStatusUpdatedActual with resourceIsTrackable=false
-		// because Jobs don't have availability tracking like Deployments/DaemonSets do
+		// because Jobs and PVCs don't have availability tracking like Deployments/DaemonSets do
```
ryanzhang-oss: PVC has conditions, I wonder why it's not trackable? However, do we really want to place PVC if they are the by-product of a statefulSet?

test/e2e/resource_placement_hub_workload_test.go
```diff
@@ -232,6 +282,46 @@ var _ = Describe("placing workloads using a CRP with PickAll policy", Label("res
...
+		if statefulSet.Status.UpdatedReplicas != requiredReplicas {
+			return fmt.Errorf("statefulset not updated: %d/%d updated replicas",
+				statefulSet.Status.UpdatedReplicas, requiredReplicas)
+		}
+
+		if statefulSet.Status.CurrentReplicas != statefulSet.Status.UpdatedReplicas {
+			return fmt.Errorf("statefulset replicas not synchronized: %d current != %d updated",
+				statefulSet.Status.CurrentReplicas, statefulSet.Status.UpdatedReplicas)
+		}
+
```
ryanzhang-oss, line 303: is there a way to use cmp.Diff instead of comparing them one by one?

test/e2e/resource_placement_hub_workload_test.go
weng271190436: I added logic to track availability of PVC in availability tracker
I don't think we should place PVC if they are the by-product of a statefulset

I originally wanted to not propagate PVC but realized that stateful set controller doesn't set owner reference on the PVC so cannot reliably tell whether a PVC is created because of a statefulset (unless I infer from the name)

The hub PVCs look like this during the hub workload e2e test

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
    volume.beta.kubernetes.io/storage-provisioner: rancher.io/local-path
    volume.kubernetes.io/selected-node: hub-control-plane
    volume.kubernetes.io/storage-provisioner: rancher.io/local-path
  creationTimestamp: "2025-11-26T17:02:01Z"
  finalizers:
  - kubernetes.io/pvc-protection
  labels:
    app: test-ss
  name: test-ss-pvc-test-ss-0
  namespace: application-1
  resourceVersion: "2328"
  uid: 1bab1a0e-d24d-44c6-bbbb-8a1f64a70e5c
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 100Mi
  storageClassName: standard
  volumeMode: Filesystem
  volumeName: pvc-1bab1a0e-d24d-44c6-bbbb-8a1f64a70e5c
status:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 100Mi
  phase: Bound
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
    volume.beta.kubernetes.io/storage-provisioner: rancher.io/local-path
    volume.kubernetes.io/selected-node: hub-control-plane
    volume.kubernetes.io/storage-provisioner: rancher.io/local-path
  creationTimestamp: "2025-11-26T17:02:08Z"
  finalizers:
  - kubernetes.io/pvc-protection
  labels:
    app: test-ss
  name: test-ss-pvc-test-ss-1
  namespace: application-1
  resourceVersion: "2424"
  uid: 6a89a3cd-7a26-47eb-82bf-af929a1d2f8b
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 100Mi
  storageClassName: standard
  volumeMode: Filesystem
  volumeName: pvc-6a89a3cd-7a26-47eb-82bf-af929a1d2f8b
status:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 100Mi
  phase: Bound
```

pkg/controllers/placement/resource_selector.go
weng271190436: removed as well

pkg/controllers/workapplier/availability_tracker.go
```diff
@@ -269,6 +271,28 @@ func trackPDBAvailability(curObj *unstructured.Unstructured) (ManifestProcessing
...
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(inMemberClusterObj.Object, &pvc); err != nil {
+		wrappedErr := fmt.Errorf("failed to convert the unstructured object to a persistent volume claim: %w", err)
+		_ = controller.NewUnexpectedBehaviorError(wrappedErr)
+		return AvailabilityResultTypeFailed, wrappedErr
+	}
+
+	// Check if the PVC is bound.
+	// A PVC is considered available when it's in the Bound phase, meaning it has been
+	// successfully bound to a PersistentVolume and is ready to be used by pods.
+	if pvc.Status.Phase == corev1.ClaimBound {
```
ryanzhang-oss: > If the StorageClass uses the WaitForFirstConsumer volume binding mode. The PVC will bind only when a Pod attempts to use it. 

this seems to be a bit tricky

test/e2e/resource_placement_hub_workload_test.go
```diff
@@ -105,9 +110,30 @@ var _ = Describe("placing workloads using a CRP with PickAll policy", Label("res
...
+				Kind:      "PersistentVolumeClaim",
+				Name:      fmt.Sprintf("%s-%s-%d", testStatefulSet.Spec.VolumeClaimTemplates[0].Name, testStatefulSet.Name, 0),
+				Namespace: workNamespace.Name,
+			},
+			{
+				Version:   "v1",
+				Kind:      "PersistentVolumeClaim",
+				Name:      fmt.Sprintf("%s-%s-%d", testStatefulSet.Spec.VolumeClaimTemplates[0].Name, testStatefulSet.Name, 1),
+				Namespace: workNamespace.Name,
+			},
```
ryanzhang-oss: What happened to the PVs created by the PVC? Do we ever place PV?

ryanzhang-oss: the controller on the member cluster will create PVCs for the statefulSet that we place on the member cluster too. From the look of it, their name would be the same (deterministic) but this can lead to problems.

weng271190436: the PVs are cluster scoped resources so they are not selected by the CRP

pkg/controllers/workapplier/availability_tracker.go
weng271190436: One way is to check the storage class using a client as well to tell the difference between VolumeBindingWaitForFirstConsumer vs others and if VolumeBindingWaitForFirstConsumer, treat pending as available as well

But since we are okay with not even propagating PVCs. That is a simpler option

ryanzhang-oss (approved):

PR #348, opened 2025-11-26
feat: rename update run state

ryanzhang-oss:
Rename the update run state so that they 
1. reflect that they are desired state instead of status
2. better convey the idea
3. keep the golang varible so it won't break existing PR

Fixes #

I have:


britaniar (approved):

britaniar:
Looks like the API validation tests need to be updated.

britaniar (approved):

PR #355, opened 2025-12-02
chore: fix the collect-logs.sh

zhiying-lin:

Hub agent logs are missing using kubectl logs command when running e2e tests.

The kubectl logs command retrieves container logs from files stored locally on the Kubernetes node by the kubelet agent, but it only accesses a limited, rotated portion of the log history by default.

Copy all the pod logs from the nodes directly.

Fixes #

I have:

### DIFF
--- a/test/e2e/collect-logs.sh
+++ b/test/e2e/collect-logs.sh
@@ -82,27 +82,27 @@ collect_agent_logs_from_node() {
     local agent_type=$4  # "hub-agent" or "member-agent"
 
     echo "    -> Collecting ${agent_type} logs from node filesystem"
-    echo "    -> Found log paths: $(docker exec "${node}" find /var/log/pods -path "*/fleet-system_*${agent_type}*")"
+    echo "    -> Found log paths: $(docker exec "${node}" find /var/log/pods -path "*/${NAMESPACE}_*${agent_type}*")"
 
     # First check if any agent logs exist on this node (including .log, .log.*, and .gz files)
     local log_files
-    log_files=$(docker exec "${node}" find /var/log/pods -path "*/fleet-system_*${agent_type}*" -type f \( -name "*.log" -o -name "*.log.*" -o -name "*.gz" \) 2>/dev/null || echo "")
+    log_files=$(docker exec "${node}" find /var/log/pods -path "*/${NAMESPACE}_*${agent_type}*" -type f \( -name "*.log" -o -name "*.log.*" -o -name "*.gz" \) 2>/dev/null || echo "")
 
     if [ -n "$log_files" ]; then
 
-        # Process each log file separately
-        echo "$log_files" | while read -r logfile; do
+        # Process each log file separately using process substitution to avoid subshell
+        while read -r logfile; do
             if [ -n "$logfile" ]; then
 
                 # Extract a meaningful filename from the log path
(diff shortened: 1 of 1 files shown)


test/e2e/collect-logs.sh
```diff
@@ -34,41 +34,124 @@ echo -e "${GREEN}Starting log collection at ${TIMESTAMP}${NC}"
...
+    local cluster_name=$2
+    local node_log_dir=$3
+    local agent_type=$4  # "hub-agent" or "member-agent"
+
+    echo "    -> Collecting ${agent_type} logs from node filesystem"
+    echo "    -> Found log paths: $(docker exec "${node}" find /var/log/pods -path "*/fleet-system_*${agent_type}*")"
+
+    # First check if any agent logs exist on this node (including .log, .log.*, and .gz files)
+    local log_files
+    log_files=$(docker exec "${node}" find /var/log/pods -path "*/fleet-system_*${agent_type}*" -type f \( -name "*.log" -o -name "*.log.*" -o -name "*.gz" \) 2>/dev/null || echo "")
```
weng271190436: can fleet-system be an argument or some constant on top? The fleet-system namespace might not change any time soon but this will help future reader realize that this script depends on fleet-system being the namespace of hub agent and member agent

test/e2e/collect-logs.sh
```diff
@@ -34,41 +34,124 @@ echo -e "${GREEN}Starting log collection at ${TIMESTAMP}${NC}"
...
+            if [ -n "$logfile" ]; then
+
+                # Extract a meaningful filename from the log path
+                local base_path=$(basename "$(dirname "$logfile")")
+                local original_filename=$(basename "$logfile")
+                local sanitized_filename="${base_path}_${original_filename}"
+
+                # Remove .gz extension for the output filename if present
+                local output_filename="${sanitized_filename%.gz}"
+                # Ensure output filename ends with .log
```
weng271190436, line 104: If there are only 2 cases, i.e. .gz and .log, maybe just handle the .gz case explicitly rather than appending .log for all cases where it doesn't end with .log? In case the file suffix is unexpected, we should log a warning because it might not even be a text file

zhiying-lin: it could be  .log, .log.*, and .gz files.

added the debug info in line 85 just in case there will be other formats i'm not aware of (but low possibility).

line 89 we only copy these three kinds of files.

test/e2e/collect-logs.sh
```diff
@@ -34,41 +34,124 @@ echo -e "${GREEN}Starting log collection at ${TIMESTAMP}${NC}"
...
+#   node_log_dir: The directory to save the collected logs
+#   agent_type: The type of agent ("hub-agent" or "member-agent")
+collect_agent_logs_from_node() {
+    local node=$1
+    local cluster_name=$2
+    local node_log_dir=$3
+    local agent_type=$4  # "hub-agent" or "member-agent"
+
+    echo "    -> Collecting ${agent_type} logs from node filesystem"
+    echo "    -> Found log paths: $(docker exec "${node}" find /var/log/pods -path "*/fleet-system_*${agent_type}*")"
```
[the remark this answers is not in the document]
zhiying-lin: would like to fail the script

weng271190436 (approved):

PR #357, opened 2025-12-02
feat: trim work object status if the object exceeds size limit

michaelawyu:

This PR sets the work applier to trim the work object status if the object to update has exceeded the size limit.

I have:

- [x] Unit tests

To control the PR size (and to validate the approach first), integration tests and E2E tests will be submitted in separate PRs.

### DIFF
--- a/pkg/controllers/workapplier/status.go
+++ b/pkg/controllers/workapplier/status.go
@@ -32,8 +32,8 @@ import (
 )
 
 const (
-	WorkStatusTrimmedDueToOversizedStatusReason  = "StatusOversized"
-	WorkStatusTrimmedDueToOversizedStatusMsgTmpl = "The status data in the Work object has been trimmed due to size constraints (%d bytes over limit %d)"
+	WorkStatusTrimmedDueToOversizedStatusReason  = "Oversized"
+	WorkStatusTrimmedDueToOversizedStatusMsgTmpl = "The status data (drift/diff details and back-reported status) has been trimmed due to size constraints (%d bytes over limit %d)"
 )
 
 // refreshWorkStatus refreshes the status of a Work object based on the processing results of its manifests.
--- a/pkg/utils/resource/resource_test.go
+++ b/pkg/utils/resource/resource_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package resource
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -70,22 +71,24 @@ func TestCalculateSizeDeltaOverLimitFor(t *testing.T) {
 			"key": "value",
 		},
 	}
+	cmBytes, err := json.Marshal(cm)
+	if err != nil {
+		t.Fatalf("Failed to marshal configMap")
+	}
+	cmSizeBytes := len(cmBytes)
 
 	testCases := []struct {
-		name               string
-		sizeLimitBytes     int
-		wantErred          bool
-		wantSizeDeltaBytes int
+		name           string
+		sizeLimitBytes int
+		wantErred      bool
 	}{
 		{
-			name:               "under size limit (negative delta)",
-			sizeLimitBytes:     10000,
-			wantSizeDeltaBytes: -9866,
+			name:           "under size limit (negative delta)",
+			sizeLimitBytes: 10000,
 		},
 		{
-			name:               "over size limit (positive delta)",
-			sizeLimitBytes:     1,
-			wantSizeDeltaBytes: 133,
+			name:           "over size limit (positive delta)",
+			sizeLimitBytes: 1,
 		},
 		{
 			name: "invalid size limit (negative size limit)",
@@ -105,8 +108,11 @@ func TestCalculateSizeDeltaOverLimitFor(t *testing.T) {
 				}
 				return
 			}
-			if !cmp.Equal(sizeDeltaBytes, tc.wantSizeDeltaBytes) {
-				t.Errorf("CalculateSizeDeltaOverLimitFor() = %d, want %d", sizeDeltaBytes, tc.wantSizeDeltaBytes)
+			// Note: this test spec uses runtime calculation rather than static values for expected
+			// size delta comparison as different platforms have slight differences in the serialization process,
+			// which may produce different sizing results.
+			if !cmp.Equal(sizeDeltaBytes, cmSizeBytes-tc.sizeLimitBytes) {
+				t.Errorf("CalculateSizeDeltaOverLimitFor() = %d, want %d", sizeDeltaBytes, cmSizeBytes-tc.sizeLimitBytes)
 			}
 		})
 	}


pkg/controllers/workapplier/status.go
```diff
@@ -643,3 +689,78 @@ func prepareRebuiltManifestCondQIdx(bundles []*manifestProcessingBundle) map[str
...
+			manifestCond.DiffDetails.ObservedDiffs = []fleetv1beta1.PatchDetail{
+				{
+					Path:          pathSummary,
+					ValueInMember: "(omitted)",
+					ValueInHub:    "(omitted)",
+				},
+			}
+		}
+
+		manifestCond.BackReportedStatus = nil
```
weng271190436, line 742: How would the user know that back reporting is not working because of oversized  work object status?

I am thinking if it is better to put something in the back reported status to let them know back reported status is nil because of oversized object not because of some misconfiguration on their side or a bug on our side.

pkg/utils/resource/resource.go
```diff
@@ -31,3 +39,20 @@ func HashOf(resource any) (string, error) {
...
 	return fmt.Sprintf("%x", sha256.Sum256(jsonBytes)), nil
 }
+
+// IsObjOversized checks if the given object exceeds the specified size limit in bytes.
+// It returns the number of bytes the object is over the limit (positive value) or
+// the number of bytes remaining before reaching the limit (negative value).
+//
+// This utility is useful in cases where KubeFleet needs to check if it can create/update
+// an object with additional information.
+func IsObjOversized(obj runtime.Object, sizeLimitBytes int) (int, error) {
```
weng271190436: the function name suggests that the return type is a boolean. Can you call this `CalculateBytesOverSizeLimit `?

pkg/utils/resource/resource_test.go
```diff
@@ -51,3 +54,60 @@ func TestHashOf(t *testing.T) {
...
+	}
+
+	testCases := []struct {
+		name               string
+		sizeLimitBytes     int
+		wantErred          bool
+		wantSizeDeltaBytes int
+	}{
+		{
+			name:               "positize size delta",
```
weng271190436: can you call this "under size limit"?

This test name is "positive size delta" but the wantSizeDeltaBytes is a negative number

pkg/utils/resource/resource_test.go
```diff
@@ -51,3 +54,60 @@ func TestHashOf(t *testing.T) {
...
+			sizeLimitBytes:     10000,
+			wantSizeDeltaBytes: -9866,
+		},
+		{
+			name:               "negative size delta",
```
weng271190436: can you call this "over size limit"?

It is confusing because it says negative but the wantSizeDeltaBytes is a positive value

Also I think positve/negative is not as descriptive as over/under size limit

pkg/controllers/workapplier/status.go
```diff
@@ -184,6 +192,44 @@ func (r *Reconciler) refreshWorkStatus(
...
+		// Normally this should never occur.
+		klog.ErrorS(err, "Failed to check Work object size before status update", "work", klog.KObj(work))
+		wrappedErr := fmt.Errorf("failed to check work object size before status update: %w", err)
+		return controller.NewUnexpectedBehaviorError(wrappedErr)
+	}
+	if sizeDeltaBytes > 0 {
+		klog.V(2).InfoS("Must trim status data as the work object has grown over its size limit",
+			"work", klog.KObj(work),
+			"sizeDeltaBytes", sizeDeltaBytes, "sizeLimitBytes", resource.DefaultObjSizeLimitWithPaddingBytes)
+		trimWorkStatusDataWhenOversized(work)
```
weng271190436, line 232: can the work object still be oversized after trimming?

pkg/controllers/workapplier/status.go
michaelawyu: Hi Wei! This PR introduces a new work object condition: `StatusTrimmed`, which notifies users that back-reported status (in addition to drift/diff details) are not be relayed back. A concern we are facing right now is that the way some of the controllers handle conditions (expecting a specific sequence of conditions) makes it a bit difficult to expose this condition on the CRP\RP API side... But at least we can show it on the work object for now.

michaelawyu: TBH my first instinct was to add a message in the back-reported status data as well 😂 The setback I encountered was that the field is designed to be a `runtime.RawExtension`, which has to wrap a valid K8s object (otherwise serialization might fail). It is a bit complicated to inject an error message this way -> we can definitely do that, but it either requires us to invent a field (which must be dropped when doing the deserialization, i.e., an implict coupling), or re-use the metadata fields (labels/annotations, they are currently always left empty and added only for structual completeness reasons), which feels just a bit weird.

michaelawyu: It will look something like this:

```yaml
backReportedStatus:
  observedStatus:
     apiVersion: apps/v1
     kind: Deployment
     metadata:
       annotations:
       - kubernetes-fleet.io/trimmed: true
     status:
```

or this:

```yaml
backReportedStatus:
  observedStatus:
     apiVersion: apps/v1
     kind: Deployment
     metadata:
     status:
     # This must be dropped when doing the de-serialization, as this field is invented and is not
     # part of the scheme for Deployments.
     trimmed: true
```

michaelawyu: I don't really have a preference on this though; if you feel that this looks better (or we should use both the `Trimmed` condition and this), please let me know.

pkg/utils/resource/resource.go
michaelawyu: Sure, let me make the edit 👌

michaelawyu: Sure, let me make the edit 👌

pkg/utils/resource/resource_test.go
michaelawyu: Will do 👌

pkg/utils/resource/resource_test.go
michaelawyu: Will do 👌

pkg/controllers/workapplier/status.go
michaelawyu: Hi Wei! Work objects are always subject to risks of getting oversized; an edge case would be, for example, the user attempts to place a single object of a size that is very close to 1.5 MiB (the system does have a 800 KiB size limit imposed, but this only applies to cases where there are multiple objects), and once the object is embedded in a Work object, the conditions we add alone might push the size over the 1.5 MiB limit and consequently all attempts to reconcile such Work objects will fail.

(Note also that the 1.5 MiB limit is configurable on the etcd side; it can be lower in upstream deployements)

The trimming can help reduce the risk as, generally speaking, conditions take very little space. But I wouldn't say it can eliminate the risk.

michaelawyu: With these being said, there are cases where we can do even better -> options include limiting the # of manifests in a resource snapshot, allow dynamic size limits, enabling compressed manifests/sharded manifests, etc. But those seem a bit like an overkill, as we do not see many cases where the limits are actually breached at this moment. Let me add a TO-DO comment to track our discussion.

weng271190436: yeah it might be an edge case of an edge case for it to be oversized after trimming. So probably fine for now

pkg/controllers/workapplier/status.go
```diff
@@ -28,9 +28,17 @@ import (
 	fleetv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/controller"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/resource"
+)
+
+const (
+	WorkStatusTrimmedDueToOversizedStatusReason  = "StatusOversized"
+	WorkStatusTrimmedDueToOversizedStatusMsgTmpl = "The status data in the Work object has been trimmed due to size constraints (%d bytes over limit %d)"
```
weng271190436: Maybe we can call out in the msg tmpl that back reported status is getting trimmed?

apis/placement/v1beta1/work_types.go
```diff
@@ -54,6 +54,10 @@ const (
 	// WorkConditionTypeDiffReported reports whether Fleet has successfully reported the
 	// configuration difference between the states in the hub cluster and a member cluster.
 	WorkConditionTypeDiffReported = "DiffReported"
+
+	// WorkConditionTypeStatusTrimmed reports whether the member agent has to trim
+	// the status data in the Work object due to size constraints.
```
weng271190436, line 59: maybe we can call out in the doc that back reported status is getting trimmed as well given that there is no "(omitted)" hint on the backReportedStatus field?

pkg/controllers/workapplier/status.go
weng271190436: This way every time users do kubectl get -oyaml or describe they should be able to see that back reported status is getting trimmed as well?

pkg/controllers/workapplier/status.go
weng271190436: Understand it is difficult to do this. And the annotation might be an overkill for now. Also it is not much better than calling out in the status message

pkg/controllers/workapplier/status.go
michaelawyu: Sure, let me add this. Thanks 🙏

apis/placement/v1beta1/work_types.go
michaelawyu: Yeah, I will refresh the docs. (Actually we haven't mentioned anything about back-reported status in the docs yet 😂 -> I'll remember to do this)

pkg/controllers/workapplier/status.go
```diff
@@ -643,3 +692,78 @@ func prepareRebuiltManifestCondQIdx(bundles []*manifestProcessingBundle) map[str
 	}
 	return rebuiltManifestCondQIdx
 }
+
+// trimWorkStatusDataWhenOversized trims some data from the Work object status when the object
+// reaches its size limit.
+func trimWorkStatusDataWhenOversized(work *fleetv1beta1.Work) {
```
ryanzhang-oss, line 698: is this deterministic? Work status change will trigger placement reconcile

PR #360, opened 2025-12-02
feat: update run abandon implementation

britaniar:

I have:
- Update the abandon state variable name to match description.
- Update the controller to take into account when to initialize, execute, abandon.
- Created a new file to include all abandon related functions in one place.
- Updated integration tests.

- Integration tests. 
Will add e2e tests later.

### DIFF
--- a/pkg/controllers/updaterun/controller_integration_test.go
+++ b/pkg/controllers/updaterun/controller_integration_test.go
@@ -779,8 +779,6 @@ func generateTrueCondition(obj client.Object, condType any) metav1.Condition {
 		typeStr = string(cond)
 	case placementv1beta1.ResourceBindingConditionType:
 		switch cond {
-		case placementv1beta1.ResourceBindingRolloutStarted:
-			reason = condition.RolloutStartedReason
 		case placementv1beta1.ResourceBindingAvailable:
 			reason = condition.AvailableReason
 		case placementv1beta1.ResourceBindingDiffReported:
--- a/pkg/controllers/updaterun/execution.go
+++ b/pkg/controllers/updaterun/execution.go
@@ -443,8 +443,8 @@ func (r *Reconciler) handleStageApprovalTask(
 ) (bool, error) {
 	updateRunRef := klog.KObj(updateRun)
 
-	approvedRequestCond := meta.FindStatusCondition(stageTaskStatus.Conditions, string(placementv1beta1.StageTaskConditionApprovalRequestApproved))
-	if approvedRequestCond != nil && approvedRequestCond.Status == metav1.ConditionTrue {
+	stageTaskApproved := condition.IsConditionStatusTrue(meta.FindStatusCondition(stageTaskStatus.Conditions, string(placementv1beta1.StageTaskConditionApprovalRequestApproved)), updateRun.GetGeneration())
+	if stageTaskApproved {
 		// The stageTask has been approved.
 		return true, nil
 	}


pkg/controllers/updaterun/abandon.go
```diff
@@ -0,0 +1,226 @@
...
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
+
+	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/controller"
+)
+
+// abandon handles the abandoning of the update run.
+func (r *Reconciler) abandon(
```
britaniar: taking suggestions on naming this function and the following function as it can used by abandon and pause. Same goes for the file name.

pkg/controllers/updaterun/halt.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/controller"
+)
+
+// abandon handles the abandoning of the update run.
+func (r *Reconciler) abandon(
+	updateRun placementv1beta1.UpdateRunObj,
+	updatingStageIndex int,
+	toBeUpdatedBindings, toBeDeletedBindings []placementv1beta1.BindingObj,
+) (bool, time.Duration, error) {
```
weng271190436: can you document the meaning of each return value in the comment above?

pkg/controllers/updaterun/abandon.go
britaniar: Thinking pause or halt would be best?

pkg/controllers/updaterun/controller.go
```diff
@@ -277,6 +310,32 @@ func (r *Reconciler) recordUpdateRunFailed(ctx context.Context, updateRun placem
...
+func (r *Reconciler) recordUpdateRunAbandoned(ctx context.Context, updateRun placementv1beta1.UpdateRunObj) error {
+	updateRunStatus := updateRun.GetUpdateRunStatus()
+	meta.SetStatusCondition(&updateRunStatus.Conditions, metav1.Condition{
+		Type:               string(placementv1beta1.StagedUpdateRunConditionProgressing),
+		Status:             metav1.ConditionFalse,
+		ObservedGeneration: updateRun.GetGeneration(),
+		Reason:             condition.UpdateRunAbandonedReason,
+		Message:            "The update run has been abandoned",
+	})
+	meta.SetStatusCondition(&updateRunStatus.Conditions, metav1.Condition{
```
Arvindthiru, line 323: What happens if an updateRun is already Succeeded, and then user tries to abandon the updateRun ?

pkg/controllers/updaterun/halt.go
```diff
@@ -0,0 +1,240 @@
...
+		bindingSpec := binding.GetBindingSpec()
+		toBeUpdatedBindingsMap[bindingSpec.TargetCluster] = binding
+	}
+
+	clusterUpdatingCount := 0
+	clusterUpdated := false
+	var stuckClusterNames []string
+	var clusterUpdateErrors []error
+	// Go through each cluster in the stage and check if it's updating/succeeded/failed/not started.
+	for i := 0; i < len(updatingStageStatus.Clusters) && clusterUpdatingCount < maxConcurrency; i++ {
```
Arvindthiru, line 89: We don't need to care about maxConcurrency here, since we are not updating any cluster just waiting for them to complete

PR #361, opened 2025-12-03
fix: always sort bindings in the scheduler to ensure deterministic decision list output

michaelawyu:

This PR sets the scheduler to always sort bindings (in list ops, and in the scheduling process), so that we always update the cluster decision list in the status deterministically.

I have:

- [x] Unit tests

### DIFF
--- a/cmd/hubagent/main.go
+++ b/cmd/hubagent/main.go
@@ -156,7 +156,8 @@ func main() {
 
 	if opts.EnableWebhook {
 		whiteListedUsers := strings.Split(opts.WhiteListedUsers, ",")
-		if err := SetupWebhook(mgr, options.WebhookClientConnectionType(opts.WebhookClientConnectionType), opts.WebhookServiceName, whiteListedUsers, opts.EnableGuardRail, opts.EnableV1Beta1APIs, opts.DenyModifyMemberClusterLabels, opts.EnableWorkload); err != nil {
+		if err := SetupWebhook(mgr, options.WebhookClientConnectionType(opts.WebhookClientConnectionType), opts.WebhookServiceName, whiteListedUsers,
+			opts.EnableGuardRail, opts.EnableV1Beta1APIs, opts.DenyModifyMemberClusterLabels, opts.EnableWorkload, opts.NetworkingAgentsEnabled); err != nil {
 			klog.ErrorS(err, "unable to set up webhook")
 			exitWithErrorFunc()
 		}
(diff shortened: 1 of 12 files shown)


pkg/utils/controller/binding_resolver.go
```diff
@@ -67,7 +68,19 @@ func ListBindingsFromKey(ctx context.Context, c client.Reader, placementKey type
 		return nil, NewAPIServerError(false, err)
 	}
 
-	return bindingList.GetBindingObjs(), nil
+	bindingObjs := bindingList.GetBindingObjs()
+
+	// Sort the list of bindings.
+	//
+	// This is needed to ensure deterministic decision output from the scheduler.
+	sort.Slice(bindingObjs, func(i, j int) bool {
```
ryanzhang-oss, line 76: just curious, why doesn't the client-go or api-server not sorting the list of bindings when they return them?

pkg/scheduler/framework/framework.go
```diff
@@ -455,6 +455,14 @@ func (f *framework) runSchedulingCycleForPickAllPlacementType(
 	// their names to achieve deterministic behaviors.
 	sort.Sort(scored)
 
+	// Sort all filtered clusters.
+	//
+	// This step is needed to produce deterministic decision outputs. If there are enough slots,
+	// the scheduler will explain why some clusters are filtered out in the decision list; to ensure
+	// that the list will not change across scheduling cycles without actual scheduling policy
+	// refreshes, the filtered clusters need to be sorted.
+	sort.Sort(filteredClusterWithStatusList(filtered))
```
ryanzhang-oss: why sort it here? I think it's actualy used in the updatePolicySnapshotStatusFromBindings function?

pkg/scheduler/framework/framework.go
```diff
@@ -944,6 +961,16 @@ func (f *framework) runSchedulingCycleForPickNPlacementType(
...
 
+	// Sort all filtered clusters.
+	//
+	// This step is needed to produce deterministic decision outputs. If there are enough slots,
+	// the scheduler will explain why some clusters are filtered out in the decision list; to ensure
+	// that the list will not change across scheduling cycles without actual scheduling policy
+	// refreshes, the filtered clusters need to be sorted.
+	//
+	// Note (chenyu1): the list of scored clusters (picked and notPicked) is already sorted.
+	sort.Sort(filteredClusterWithStatusList(filtered))
```
ryanzhang-oss: same here

pkg/utils/controller/binding_resolver.go
michaelawyu: Hi Ryan! The API side does not have a sorting related option for LIST ops IIRC (for performance reasons maybe?); `kubectl` does have a `--sort-by` param, but it's doing the sorting locally: [Link](https://github.com/kubernetes/kubectl/blob/9978cf7c5f8319b783ba044c2f226753f9c30fef/pkg/cmd/get/sorter.go#L70).

pkg/scheduler/framework/framework.go
michaelawyu: Hi Ryan! Yeah, this is used to populate the status (earlier we had a feedback that we should explain why some clusters are not picked in the status). If we do not sort this output, the list of decisions might be constantly changing, esp. when there are >20 unselected member clusters.

ryanzhang-oss: I understand why we need to sort. My question was we now have two places that we sort the list but the actual function used the list is in `updatePolicySnapshotStatusFromBindings`. Why not sort the list there so it's sorted at one place and close to where it's actually used.

michaelawyu: Hi Ryan! The `scored` list (i.e., the picked clusters) has to be sorted right after `runAllPluginsForPickAllPlacementType` because the scheduler is set to prefer clusters whose names rank higher alphabetically (deterministic decision making); that's why the `filtered` list (i.e., filtered clusters) are put at the same place. I don't have a preference on this though, if you'd like to keep the sorting behavior for the `filtered` list in `updatePolicySnapshotStatusFromBindings`, I could make the edit.

michaelawyu: Done ✅

pkg/scheduler/framework/framework.go
michaelawyu: Done ✅

PR #335, opened 2025-11-18
feat: only trigger change on generation change

weng271190436:
### Description of your changes

Should only trigger placement reconciliation if generation changes. Resource version can change when status was updated. Generation change indicates a change in desired state (spec + part of meta such as annotation and label)

I have:

- [x] Run `make reviewable` to ensure this PR is ready for review.

PR #362, opened 2025-12-04
fix: correctly set up priority queue in work applier

michaelawyu:

This PR fixes a variety of issues in the current option that enables usage of priority queue in the work applier, specifically:

* priority queue is not registered with the controller
* double enqueueing
* priority settings are not applied for enqueued items
* missing support for reportDiff mode
* incorrent priority assigment

I have:

- [x] Unit tests

N/A


pkg/controllers/workapplier/pq.go
```diff
@@ -0,0 +1,229 @@
...
+	// * with highest priority (-2): all Create/Delete events, and all Update events
+	//   that concern recently created Work objects or Work objects that are in a failed/undeterminted
+	//   state (apply op/availability check failure, or diff reporting failure).
+	// * with medium priority (-1): all other Update events.
+	// * with default priority (0): all requeues (with or with errors), and all Generic events.
+	//
+	// Note that requests with the same priority level will be processed in the FIFO order.
+	//
+	// TO-DO (chenyu1): evaluate if/how we need to/should prioritize requeues properly.
+	highPriorityLevel    = 2
```
weng271190436: the comment above says -2. Does smaller number mean higher priority or is it the other way around?

pkg/controllers/workapplier/pq.go
```diff
@@ -0,0 +1,229 @@
...
+	// Do a sanity check.
+	if h.qm.PriorityQueue() == nil {
+		wrappedErr := fmt.Errorf("received an Update event, but the priority queue is not initialized")
+		_ = controller.NewUnexpectedBehaviorError(wrappedErr)
+		klog.ErrorS(wrappedErr, "Failed to process Update event")
+		return
+	}
+
+	// Ignore status only updates.
+	if updateEvent.ObjectOld.GetGeneration() == updateEvent.ObjectNew.GetGeneration() {
```
weng271190436, line 166: I tried to do the same thing but realized that generation doesn't change when deployment label changes. There are probably other examples. We might still want to trigger a reconciliation if generation doesn't change

https://github.com/kubefleet-dev/kubefleet/pull/335

pkg/controllers/workapplier/pq.go
```diff
@@ -0,0 +1,229 @@
...
+			Name:      workObjName,
+		},
+	}
+	h.qm.PriorityQueue().AddWithOpts(opts, req)
+}
+
+// Update implements the TypedEventHandler interface.
+func (h *priorityBasedWorkObjEventHandler) Update(_ context.Context, updateEvent event.TypedUpdateEvent[client.Object], _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	// Do a sanity check.
+	if h.qm.PriorityQueue() == nil {
```
weng271190436, line 158: if priority queue is nil, it means somehow the controller is not even doing reconciling? I am wondering if we should just crash the controller given that its primary functionality is compromised

pkg/controllers/workapplier/pq.go
```diff
@@ -0,0 +1,229 @@
...
+type priorityBasedWorkObjEventHandler struct {
+	qm                                 CustomPriorityQueueManager
+	workObjAgeForPrioritizedProcessing time.Duration
+}
+
+// Create implements the TypedEventHandler interface.
+func (h *priorityBasedWorkObjEventHandler) Create(_ context.Context, createEvent event.TypedCreateEvent[client.Object], _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	// Do a sanity check.
+	//
+	// Normally when this method is called, the priority queue has been initialized.
```
weng271190436: Why would the priority queue not be initialized when this method is called?

pkg/controllers/workapplier/pq.go
```diff
@@ -0,0 +1,229 @@
...
+		NamespacedName: types.NamespacedName{
+			Namespace: workObjNS,
+			Name:      workObjName,
+		},
+	}
+	h.qm.PriorityQueue().AddWithOpts(opts, req)
+}
+
+func (h *priorityBasedWorkObjEventHandler) determineUpdateEventPriority(oldWorkObj, newWorkObj *fleetv1beta1.Work) int {
+	// If the work object is recently created (its age is within the given threshold),
```
weng271190436, line 221: Can you comment on why we would want to process a new work object before an old work object?

pkg/controllers/workapplier/pq_test.go
```diff
@@ -0,0 +1,604 @@
+/*
```
weng271190436, line 1: how does this PR fix double enqueueing?

Is it possible to test that by enqueueing 2 identical events in the pq_test?

pkg/controllers/workapplier/controller.go
```diff
@@ -252,23 +138,37 @@ func NewReconciler(
 		parallelizer = parallelizerutil.NewParallelizer(1)
 	}
 
+	woAgeForPrioritizedProcessing := workObjAgeForPrioritizedProcessing
+	if usePriorityQueue && woAgeForPrioritizedProcessing < minWorkObjAgeForPrioritizedQueueing {
+		klog.V(2).InfoS("Work object age for prioritized processing is too short; set to the longer default", "workObjAgeForPrioritizedProcessing", woAgeForPrioritizedProcessing)
```
weng271190436: Can you comment on why 30-minute is a recommended minimum?

Also suggest just crash the controller if the user input is invalid. Users might not read logs and expect whatever they specify as workObjAgeForPrioritizedProcessing will be respected by the controller

weng271190436: and this is during startup, which is a good time to fail fast

pkg/controllers/workapplier/pq.go
michaelawyu: Hi Wei! The current values are correct (as verified by the UTs); let me fix the comment.

pkg/controllers/workapplier/pq.go
michaelawyu: Hi Wei! The work applier only processes work objects, which are CRs; and for CRs the generation always changes when their specs (not metadata) change. This does not necesssarily apply to Kubernetes built-in APIs.

michaelawyu: This might be of some relevance: [link](https://github.com/kubernetes-sigs/controller-runtime/blob/7a1b16d039d204e7e4c5c8df57124d4e060a55f0/pkg/predicate/predicate.go#L172)

```golang
// GenerationChangedPredicate implements a default update predicate function on Generation change.
//
// This predicate will skip update events that have no change in the object's metadata.generation field.
// The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object.
// This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.
//
// For CustomResource objects the Generation is incremented when spec is changed, or status changed and status not modeled as subresource.
// subresource status update will not increase Generation.
//
// Caveats:
//
// * The assumption that the Generation is incremented only on writing to the spec does not hold for all APIs.
// E.g For Deployment objects the Generation is also incremented on writes to the metadata.annotations field.
// For object types other than CustomResources be sure to verify which fields will trigger a Generation increment when they are written to.
//
// * With this predicate, any update events with writes only to the status field will not be reconciled.
// So in the event that the status block is overwritten or wiped by someone else the controller will not self-correct to restore the correct status.
type GenerationChangedPredicate = TypedGenerationChangedPredicate[client.Object]
```

pkg/controllers/workapplier/pq.go
michaelawyu: Hi Wei! The use of priority queues is optional; if not enabled the system will use a default, FIFO work queue. For this specific part, we only set up the `priorityBasedWorkObjEventHandler` when the use of priority queues is enabled, and since all of such setup are private variables, we do not do any additional checkup/validation.

michaelawyu: The sanity check is added because:

* The `Create`, `Update`, `Delete`, and `Generic` event handlers are invoked by the controller-runtime package when the controller runs (i.e., it's something out of our control);
* The priority queue is only initialized after the controller has been set up with the controller manager; if the event handlers are called before such setup is done, the queue variable would be nil.

Of course, logically speaking the event handlers can only run after the informer gets ready so the controller must have already been set up with the manager; so this is only added as sanity check. Normally the branch will never run.

pkg/controllers/workapplier/pq.go
michaelawyu: Hi Wei! This is also connected to the reply above; if there's any concern, please let me know.

pkg/controllers/workapplier/pq.go
michaelawyu: Hi Wei! A bit background on this PR:

The use of priority queues is first proposed by one of our upstream users, whose major concern/complaint was that, when there are a lot of work objects (i.e., a lot of CRPs/RPs), the work applier will spend a lot of cycles re-processing existing work objects as we periodically requeues objects, and as a result it might take a long while for a newly created work object to be processed. This is esp. true when our agents restart; the user finds that sometimes  it may take 30 minutes or more before the system becomes responsive to newly created CRPs/RPs again. 

For obvious reasons most of the requeues for older work objects do not actually make any difference; we have to do them just to ensure correctness. The use of priority queues will allow the work applier for prioritize new objects first and only return to the mundance processing for requeues when there's no more work to do.

pkg/controllers/workapplier/pq_test.go
michaelawyu: Hi Wei! A bit of clarification:

In the current implementation (main branch), when the use of priority queues is enabled, the system is actually set up with two queues (a normal work queue and a priority queue). When a Create/Update/Delete/Generic event arrives, it is enqueued to both queues, but only the default queue is actually working (i.e., the system only processes events from the default queue). This is why I mentioned about double enqueueing in the PR description. It does not imply that the same key is enqueued twice to the same queue. Sorry if there has been any confusion.

michaelawyu: And all built-in K8s workqueues (priority based or not) already handle enqueueing of identical keys gracefully (via dirty writes); it is a core/foundational feature of informers so we do not typically verify it.

pkg/controllers/workapplier/controller.go
michaelawyu: Ah, the current value in use in the main branch is 1-hour, which was set by the user who originally proposed priority queues, as aforementioned (this PR does not change this default value). 

The value has to strike a balance between responsiveness and fairness; if set too high, all requests will be assigned high priority, which defeats the purpose of having priority queues (this mostly applies to those users that use KubeFleet to run batch work), and if set too low new work objects might not get its fair chance to become ready.

The proper value is kind of user-dependent, which is why in this PR I am only blocking this value from being set too low.

michaelawyu: With this being said, admitted we do not have enough empirical data on what a good default value really is at the moment.

And for the check part, I do not have a strong preference on the subject (and I 100% agree that many do not actually read logs 😂). This is currently following the pattern with the work applier backoff rate limiter (which, too, will overwrite invalid values with defaults and emit an error log).

pkg/controllers/workapplier/pq.go
weng271190436:
```
Normally the branch will never run.
```

If something not normal happens, shouldn't we crash rather than not process events at all?

pkg/controllers/workapplier/pq.go
weng271190436: I think it depends on what "normal" means here:
1. if "normal" means that if this happens, there is a bug, then I think we should crash
2. if "normal" means that even if there is no bug, this can still happen from time to time (e.g. there is a race inherent to how controller runtime is implemented), then I think we should comment on under what situation this can happen and say this is the best we can do given how controller runtime is implemented

pkg/controllers/workapplier/pq.go
weng271190436: Got it! I think this is good context worth a comment in code because a future contributor wouldn't understand why this threshold exists in the first place just by reading code

pkg/controllers/workapplier/pq.go
michaelawyu: Yeah, it's true... Though for this case, if the event handler does get called before the controller finishes setup, there's no side effect and the situation can still recover (as soon as the controller gets back). 

To me this is a little bit like the interface casting checks we do in event handlers (e.g., the one in the scheduler member cluster watcher, or the rollout controller), which are almost guaranteed to work, we just keep the checks there for completeness reasons 😂

pkg/controllers/workapplier/pq_test.go
weng271190436: Got it! Thanks for explaining!

pkg/controllers/workapplier/controller.go
weng271190436: Got it. I do find it strange that if a user provides an invalid input, we just set it to a closest valid value rather than tell the user that your input is invalid please try again with a valid value because this gives the user the impression that their input is perfectly fine.

But I wouldn't block the PR if the team agree that this is best practice

pkg/controllers/workapplier/pq.go
weng271190436: yeah but I guess the "out of our control" argument is valid. Maybe just comment on exactly why you need the sanity check - because it is out of our control while admitting the risk of silently dropping events?

pkg/controllers/workapplier/controller.go
ryanzhang-oss: Wei has a point, we normally should not silently change a user input.

pkg/controllers/workapplier/suite_test.go
```diff
@@ -370,9 +370,9 @@ var _ = BeforeSuite(func() {
 		maxConcurrentReconciles,
 		pWithDelay,
 		30*time.Second,
-		true,
-		60,
-		nil, // Use the default backoff rate limiter.
+		nil,   // Use the default backoff rate limiter.
+		false, // Disable priority queueing.
+		0,
 	)
```
ryanzhang-oss, line 404: it seems that there are no tests other than UT in this PR? We should add IT/e2e tests.

pkg/controllers/workapplier/pq.go
```diff
@@ -0,0 +1,229 @@
...
+		// Use medium priority for the Update event.
+		return mediumPriorityLevel
+	case isReportDiffModeEnabled:
+		// The ReportDiff mode is enabled, but the diff reporting has not been completed yet or has failed.
+		// Use high priority for the Update event.
+		return highPriorityLevel
+	case condition.IsConditionStatusTrue(appliedCond, oldWorkObj.Generation) && condition.IsConditionStatusTrue(availableCond, oldWorkObj.Generation):
+		// The apply strategy is set to the CSA/SSA mode and the work object is applied and available.
+		// Use medium priority for the Update event.
+		return mediumPriorityLevel
```
ryanzhang-oss: should this be default?

pkg/controllers/workapplier/pq.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func (h *priorityBasedWorkObjEventHandler) determineUpdateEventPriority(oldWorkObj, newWorkObj *fleetv1beta1.Work) int {
+	// If the work object is recently created (its age is within the given threshold),
+	// process its Update event with high priority.
+
+	// The age is expected to be the same for both old and new work objects, as the field
+	// is immutable and not user configurable.
+	workObjAge := time.Since(newWorkObj.CreationTimestamp.Time)
+	if workObjAge <= h.workObjAgeForPrioritizedProcessing {
```
ryanzhang-oss: using a single time based cut off makes this logic pretty arbituary, I wonder if we should have multiple slot?

ryanzhang-oss (commented):
We need at least integation test

pkg/controllers/workapplier/suite_test.go
michaelawyu: Hi Ryan! For this one I'd like to introduce a test combo to run the ITs on both setup (w/ simple queue and w/ priority queue) rather than just adding one test for the priority queue. The implementation is not enabled yet; if OK I'd like to send a separate PR to enable the tests for both setup all together.

pkg/controllers/workapplier/pq.go
michaelawyu: Hi Ryan! Sure, I could rename it as default level and update the other two names accordingly.

pkg/controllers/workapplier/pq.go
michaelawyu: Hi Ryan! Do you mean a hierarchy, e.g., <5 minutes -> Pri. 5, 5-10 minutes -> Pri 4, 10-20 minutes -> Pri 3, etc. ?

ryanzhang-oss: yeah, something like that

pkg/controllers/workapplier/suite_test.go
ryanzhang-oss: I wonder why we need to do that? each Queue should have their own UT and the IT should be testing against the common interface. We do need to run perf test though

pkg/controllers/workapplier/pq.go
michaelawyu: I've added a comment.

pkg/controllers/workapplier/pq.go
michaelawyu: I've added a comment, PTAL.

pkg/controllers/workapplier/controller.go
michaelawyu: For this PR I've added the validation in the entrypoint part. Will submit subsequent PRs to keep things consistent.

pkg/controllers/workapplier/suite_test.go
michaelawyu: Hi Ryan! For UTs things are separate in nature. For ITs, as you mentioned, we are testing against a standard interface (with different underly queue implementations); to make sure that the behavior is consistent, the best way to achieve optimal coverage would be to run the same tests against both implementations I assume. Adding one specific IT for priority queues that covers only the basic scenario seems a bit inadequate I fear.

pkg/controllers/workapplier/pq.go
michaelawyu: The names have been updated.

pkg/controllers/workapplier/pq.go
michaelawyu: Hi Ryan! Per our earlier offline discussion I've switched to a linear equation with user-configurable parameters.

pkg/controllers/workapplier/controller.go
```diff
@@ -251,24 +137,39 @@ func NewReconciler(
 		klog.V(2).InfoS("parallelizer is not set; using the default parallelizer with a worker count of 1")
 		parallelizer = parallelizerutil.NewParallelizer(1)
 	}
+	if priorityLinearEquationCoeffA == nil || priorityLinearEquationCoeffB == nil {
```
ryanzhang-oss, line 146: how can these be nil? The option has default value.

ryanzhang-oss: are those just for test?

pkg/controllers/workapplier/pq.go
```diff
@@ -0,0 +1,260 @@
...
+		// Use high priority for the Update event.
+		return highestPriorityLevel
+	case condition.IsConditionStatusTrue(appliedCond, oldWorkObj.Generation) && condition.IsConditionStatusTrue(availableCond, oldWorkObj.Generation):
+		// The apply strategy is set to the CSA/SSA mode and the work object is applied and available.
+		// Determine the priority level based on the age of the Work object.
+		return h.calcArgBasedPriWithLinearEquation(workObjAgeMinutes)
+	default:
+		// The apply strategy is set to the CSA/SSA mode and the work object is in a failed/undetermined state.
+		// Use high priority for the Update event.
+		return highestPriorityLevel
```
ryanzhang-oss, line 260: IIRC, we will reconcile a failed work aggressively (this logic applies to requeued message too)? Is that what we want?

pkg/controllers/workapplier/controller.go
michaelawyu: Hi Ryan! I just added the check for completeness reasons. It's true that normally such parameters will not be nil.

pkg/controllers/workapplier/pq.go
michaelawyu: Hi Ryan! I've made some changes: originally the controller runtime will requeue with the default priority; I saw that with the latest version the behavior has changed to requeue with a request's original priority, which does not work well in our case (a priorized event will remain prioritized). I have made some changes so that requeues are processed manually when priority queue is in use.

michaelawyu: To summarize, in the current setup:

* only Create/Delete/Update events (i.e., spec changes) are prioritized.
* regular requeues always have the lowest priority (but might have different backoff periods). This is consistent with the current behavior.
* requeues with errors are processed with its original priority.

In other words, failed (user errors, e.g., availability errors) works will always be requeued with the lowest priority and the delay given by the exponential backoff mechanism; however, if the user updates the spec of such works, the updates will be handled with the highest priority. If work is successful and has a spec change, the request is prioritized based on its creation timestamp.

michaelawyu: If there's any concern, please let me know.

pkg/controllers/workapplier/pq.go
```diff
@@ -0,0 +1,294 @@
...
+		NamespacedName: types.NamespacedName{
+			Namespace: workObjNS,
+			Name:      workObjName,
+		},
+	}
+	h.qm.PriorityQueue().AddWithOpts(opts, req)
+}
+
+// Update implements the TypedEventHandler interface.
+func (h *priorityBasedWorkObjEventHandler) Update(_ context.Context, updateEvent event.TypedUpdateEvent[client.Object], _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
```
ryanzhang-oss, line 156: so we don't have rate limit for the priority queue? Does it have dedup?

michaelawyu: Hi Ryan! The priority queue we use does have a rate limiter:

```go
newPQ := func(controllerName string, rateLimiter workqueue.TypedRateLimiter[reconcile.Request]) workqueue.TypedRateLimitingInterface[reconcile.Request] {
			withRateLimiterOpt := func(opts *priorityqueue.Opts[reconcile.Request]) {
				opts.RateLimiter = rateLimiter
			}
			r.pqSetupOnce.Do(func() {
				r.pq = priorityqueue.New(controllerName, withRateLimiterOpt)
			})
			return r.pq
		}
```

The rate limiting applies when we call the `AddRateLimited` method, which would apply to all requeue (erred or not) requests. The Create/Update/Delete event handlers call the `AddWithOpts` methods with the rate limiting option set to false, this is consistent with other non-prioritized event handlers in our code (which just call the `Add` method -> the rate limiting will also be bypassed).

michaelawyu: The priority queue has de-dup (the built-in logic in controller runtime).

michaelawyu:
Merged to unblock progress; if there is any concern, please let me know.

PR #366, opened 2025-12-05
feat: enable HA hub agents by optionally depending on cert manager

weng271190436:

Currently the in memory 10-year self-signed cert cannot be shared among replicas of hub agents so only the leader has a valid cert registered with API server

Using cert manger decouples cert management and hub agent core functionality. It also does cert rotation.

Trying to partially address https://github.com/Azure/fleet/issues/1224

Fixes #

I have:

### DIFF
--- a/pkg/resourcewatcher/informer_populator.go
+++ b/pkg/resourcewatcher/informer_populator.go
@@ -37,7 +37,8 @@ const (
 
 // make sure that our InformerPopulator implements controller runtime interfaces
 var (
-	_ manager.Runnable = &InformerPopulator{}
+	_ manager.Runnable               = &InformerPopulator{}
+	_ manager.LeaderElectionRunnable = &InformerPopulator{}
 )
 
 // InformerPopulator discovers API resources and creates informers for them WITHOUT adding event handlers.
@@ -92,4 +93,10 @@ func (p *InformerPopulator) discoverAndCreateInformers() {
 	p.InformerManager.Start()
 
 	klog.V(2).InfoS("Informer populator: discovered resources", "count", len(resourcesToWatch))
+}
+
+// NeedLeaderElection implements LeaderElectionRunnable interface.
+// Returns false so this runs on ALL pods (leader and followers).
+func (p *InformerPopulator) NeedLeaderElection() bool {
+	return false
 }
--- a/pkg/resourcewatcher/informer_populator_test.go
+++ b/pkg/resourcewatcher/informer_populator_test.go
@@ -39,6 +39,15 @@ const (
 	testSleep = 150 * time.Millisecond
 )
 
+func TestInformerPopulator_NeedLeaderElection(t *testing.T) {
+	populator := &InformerPopulator{}
+
+	// InformerPopulator should NOT need leader election so it runs on all pods
+	if populator.NeedLeaderElection() {
+		t.Error("InformerPopulator should not need leader election")
+	}
+}
+
 func TestInformerPopulator_discoverAndCreateInformers(t *testing.T) {
 	tests := []struct {
 		name                    string


cmd/hubagent/main.go
```diff
@@ -156,18 +157,27 @@ func main() {
...
 	if err := workload.SetupControllers(ctx, &wg, mgr, config, opts); err != nil {
-		klog.ErrorS(err, "unable to set up ready check")
+		klog.ErrorS(err, "unable to set up controllers")
 		exitWithErrorFunc()
 	}
 
+	// Add webhook readiness check AFTER controllers are set up (when ResourceInformer is initialized)
+	// This prevents webhook from accepting requests before discovery cache is populated
+	if opts.EnableWebhook {
+		if err := mgr.AddReadyzCheck("webhook-cache", webhook.ResourceInformerReadinessChecker(validator.ResourceInformer)); err != nil {
```
weng271190436: This PR stacks on top of an informer readiness check change because now that I have multiple replicas of webhook servers, it becomes likely that some webhook servers might start serving requests before the cache is synced

pkg/webhook/webhook.go
```diff
@@ -174,17 +180,34 @@ func NewWebhookConfig(mgr manager.Manager, webhookServiceName string, port int32
 		serviceNamespace:              namespace,
 		serviceName:                   webhookServiceName,
 		serviceURL:                    fmt.Sprintf("https://%s.%s.svc.cluster.local:%d", webhookServiceName, namespace, port),
+		certDir:                       certDir,
```
ryanzhang-oss: where is this used?

weng271190436: Removed. Thanks for spotting

cmd/hubagent/main.go
weng271190436: Still discussing the other PR https://github.com/kubefleet-dev/kubefleet/pull/367

If the other PR doesn't merge first e2e tests won't pass for HA hub agent setup

pkg/resourcewatcher/informer_populator.go
```diff
@@ -0,0 +1,102 @@
...
+
+const (
+	// informerPopulatorDiscoveryPeriod is how often the InformerPopulator rediscovers API resources
+	informerPopulatorDiscoveryPeriod = 30 * time.Second
+)
+
+// make sure that our InformerPopulator implements controller runtime interfaces
+var (
+	_ manager.Runnable               = &InformerPopulator{}
+	_ manager.LeaderElectionRunnable = &InformerPopulator{}
```
ryanzhang-oss, line 41: IIRC, you just don't need to implement this interface and it will always run

cmd/hubagent/main.go
```diff
@@ -157,12 +156,31 @@ func main() {
 	}
 
 	if opts.EnableWebhook {
-		whiteListedUsers := strings.Split(opts.WhiteListedUsers, ",")
-		if err := SetupWebhook(mgr, options.WebhookClientConnectionType(opts.WebhookClientConnectionType), opts.WebhookServiceName, whiteListedUsers,
```
weng271190436, line 161: refactored this a bit to avoid having SetupWebhook with long list arguments since I am adding some more

charts/hub-agent/values.yaml
```diff
@@ -17,13 +17,26 @@ webhookServiceName: fleetwebhook
...
+useCertManager: false
+# webhookCertDir is the directory where webhook certificates are mounted
+# This is configurable via --webhook-cert-dir flag and must match the volumeMount path in deployment
+webhookCertDir: /tmp/k8s-webhook-server/serving-certs
+# webhookCertName is the name of the Certificate resource created by cert-manager
+# This is referenced in the cert-manager.io/inject-ca-from annotation
+webhookCertName: fleet-webhook-certificate
+# webhookCertSecretName is the name of the Secret where cert-manager stores the certificate
+# This must match the secretName in the Certificate resource and the volume mount in deployment
+webhookCertSecretName: fleet-webhook-server-cert
```
ryanzhang-oss: why set those values when useCertManager: false?

charts/hub-agent/README.md
```diff
@@ -2,11 +2,33 @@
...
+When using cert-manager for certificate management, install cert-manager as a prerequisite first:
+
+```console
+# Install cert-manager
+helm repo add jetstack https://charts.jetstack.io
+helm repo update
+helm install cert-manager jetstack/cert-manager \
+  --namespace cert-manager \
+  --create-namespace \
+  --version v1.16.2 \
```
ryanzhang-oss: what happens when this version does not exist anymore?

charts/hub-agent/README.md
```diff
@@ -41,4 +71,67 @@ _See [helm install](https://helm.sh/docs/helm/helm_install/) for command documen
...
\ No newline at end of file
+| `enableWorkload`                          | Enable kubernetes builtin workload to run in hub cluster.                                  | `false`                                          |
+
+## Certificate Management
+
+The hub-agent supports two modes for webhook certificate management:
+
+### Automatic Certificate Generation (Default)
+
+By default, the hub-agent generates certificates automatically at startup. This mode:
```
ryanzhang-oss, line 80: but this mode only allows one replica for the hub-agent deployment?

cmd/hubagent/options/options.go
```diff
@@ -110,6 +110,15 @@ type Options struct {
 	// EnableWorkload enables workload resources (pods and replicasets) to be created in the hub cluster.
 	// When set to true, the pod and replicaset validating webhooks are disabled.
 	EnableWorkload bool
+	// UseCertManager indicates whether to use cert-manager for webhook certificate management.
+	// When enabled, webhook certificates are managed by cert-manager instead of self-signed generation.
+	UseCertManager bool
+	// WebhookCertDir is the directory where webhook certificates are stored/mounted.
+	// This must match the mountPath in the Helm chart deployment when using cert-manager.
+	WebhookCertDir string
```
ryanzhang-oss: why do we want to expose this to user?

cmd/hubagent/options/options.go
```diff
@@ -110,6 +110,15 @@ type Options struct {
...
+	// WebhookCertName is the name of the Certificate resource created by cert-manager.
+	// This is referenced in the cert-manager.io/inject-ca-from annotation.
+	WebhookCertName string
```
ryanzhang-oss: same

pkg/resourcewatcher/informer_populator.go
weng271190436: I tried but didn't work. The reason is that NeedLeaderElection is by default true, according to https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/config

```
         // NeedLeaderElection indicates whether the controller needs to use leader election.
	// Defaults to true, which means the controller will use leader election.
	NeedLeaderElection *[bool](https://pkg.go.dev/builtin#bool)
```

cmd/hubagent/options/options.go
weng271190436: You are right that this is not something we want the user to configure
I made this an option because the value in deployment.yaml and in hub agent code must match.

I removed the option and left comments in webhook.go constants and yamls saying that the values must match instead.

Did the same for WebhookCertName

charts/hub-agent/README.md
weng271190436: do not specify --version in README now but call out that CERT_MANAGER_VERSION in .github/workflows/ci.yml has the version of cert-manager which is validated by e2e tests

charts/hub-agent/README.md
weng271190436: call out the limitationi that this only supports single replica deployment in a bullet point below

PR #368, opened 2025-12-05
fix: continue to propagate PVC when workload is not enabled

ryanzhang-oss:
We should continue to propagate PVC when workload is not enabled

 
Fixes #

I have:

### DIFF
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -506,7 +506,7 @@ func CheckCRDInstalled(discoveryClient discovery.DiscoveryInterface, gvk schema.
 }
 
 // ShouldPropagateObj decides if one should propagate the object.
-// If enableWorkload is true, PVCs are allowed to be propagated.
+// PVCs are only propagated when enableWorkload is false (workloads not allowed on hub).
 func ShouldPropagateObj(informerManager informer.Manager, uObj *unstructured.Unstructured, enableWorkload bool) (bool, error) {
 	// TODO:  add more special handling for different resource kind
 	switch uObj.GroupVersionKind() {


pkg/utils/common.go
```diff
@@ -505,29 +505,30 @@ func CheckCRDInstalled(discoveryClient discovery.DiscoveryInterface, gvk schema.
 	return err
 }
 
-// ShouldPropagateObj decides if one should propagate the object
-func ShouldPropagateObj(informerManager informer.Manager, uObj *unstructured.Unstructured) (bool, error) {
+// ShouldPropagateObj decides if one should propagate the object.
+// If enableWorkload is true, PVCs are allowed to be propagated.
```
weng271190436: if enableworkload is false?

pkg/utils/common_test.go
```diff
@@ -1209,6 +1210,21 @@ func TestShouldPropagateObj_PodAndReplicaSet(t *testing.T) {
 				},
 			},
 			ownerReferences: nil,
+			enableWorkload:  true,
+			want:            true,
+		},
+		{
+			name: "standalone replicaset without ownerReferences should propagate if workload is disabled",
```
weng271190436, line 1217: replicaset propagation has nothing to do with workload being disabled or not?

pkg/utils/common_test.go
```diff
@@ -1328,8 +1364,30 @@ func TestShouldPropagateObj_PodAndReplicaSet(t *testing.T) {
 				},
 			},
 			ownerReferences: nil,
+			enableWorkload:  true,
 			want:            false,
 		},
+		{
+			name: "PVC with ownerReferences should NOT propagate when workload is enabled",
```
weng271190436, line 1371: Don't think we use owner reference to determine PVC propagation

pkg/utils/common_test.go
ryanzhang-oss: yes, you are right. This test is to make sure that is the case. We test on behaviors not code.

pkg/utils/common_test.go
ryanzhang-oss: yes, you are right. This is to test behavior instead of implementation.

PR #369, opened 2025-12-05
test: fix TestPickBindingsToRoll unit test

weng271190436:

fix TestPickBindingsToRoll unit test

The existing now() is defined at package level. Some time might have elapsed between that and when the test starts 

I have:


pkg/controllers/rollout/controller_test.go
```diff
@@ -2501,6 +2508,15 @@ func TestPickBindingsToRoll(t *testing.T) {
 	}
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
+			now := time.Now() // Capture fresh timestamp for this specific test case
```
ryanzhang-oss, line 2511: just use this "now" should suffice? We don't need the now to be embeded?

PR #373, opened 2025-12-09
feat: add implementation for scheduler work queues with batch processing support

michaelawyu:

This PR adds an implementation for scheduler work queues that allows certain change events (e.g., member cluster CR refreshes) to be processed in batch periodically rather than immediately, to save some unnecessary processing cycles.

I have:

- [x] Unit tests

### DIFF
--- a/pkg/scheduler/queue/batched.go
+++ b/pkg/scheduler/queue/batched.go
@@ -68,7 +68,8 @@ func (bq *batchedProcessingPlacementSchedulingQueue) Close() {
 	// Signal the mover goroutine to exit.
 	//
 	// Note that this will trigger the mover goroutine to attempt another key move, but the
-	// active queue will not be able to accept the key any more.
+	// active queue might not be able to accept the key any more (which is OK and does not
+	// result in an error).
 	close(bq.moveNow)
 
 	bq.batched.ShutDown()
@@ -82,9 +83,10 @@ func (bq *batchedProcessingPlacementSchedulingQueue) CloseWithDrain() {
 	// Signal that all items in the batched queue should be moved to the active queue right away.
 	close(bq.moveNow)
 
-	// Wait until all the items in the batched queue have been processed..
+	// Wait until all the items in the moving process from the batched queue to the active queue have completed
+	// their moves.
 	bq.batched.ShutDownWithDrain()
-	// Wait until all the items in the active queue have been processed.
+	// Wait until all the items that are currently being processed by the scheduler to finish.
 	bq.active.ShutDownWithDrain()
 }
 
@@ -102,9 +104,7 @@ func (bq *batchedProcessingPlacementSchedulingQueue) NextPlacementKey() (key Pla
 // Done marks a PlacementKey as done.
 func (bq *batchedProcessingPlacementSchedulingQueue) Done(placementKey PlacementKey) {
 	bq.active.Done(placementKey)
-	// It is OK for Done to be called on the batched queue even if the item has never been
-	// added to the batched queue before. In this case the call is simply a no-op.
-	bq.batched.Done(placementKey)
+	// The keys in the batched queue are marked as done as soon as they are moved to the active queue.
 }
 
 // Add adds a PlacementKey to the work queue for immediate processing.
@@ -130,9 +130,7 @@ func (bq *batchedProcessingPlacementSchedulingQueue) AddRateLimited(placementKey
 // Forget untracks a PlacementKey from rate limiter(s) (if any) set up with the queue.
 func (bq *batchedProcessingPlacementSchedulingQueue) Forget(placementKey PlacementKey) {
 	bq.active.Forget(placementKey)
-	// It is OK for Forget to be called on the batched queue even if the item has never been
-	// added to the batched queue before. In this case the call is simply a no-op.
-	bq.batched.Forget(placementKey)
+	// The keys in the batched queue are forgotten as soon as they are moved to the active queue.
 }
 
 // AddBatched tracks a PlacementKey and adds such keys in batch later to the work queue when appropriate.
@@ -188,7 +186,7 @@ func (bq *batchedProcessingPlacementSchedulingQueue) moveAllBatchedItemsToActive
 			// The keys popped from the batched queue are not yet added to the active queue, in other words,
 			// they are not yet marked as done; the batched queue will still track them and adding them
 			// to the batched queue again at this moment will not trigger the batched queue to yield the same
-			// keys again. This implies that the at maximum we will be movine a number of keys equal to
+			// keys again. This implies that the at maximum we will be moving a number of keys equal to
 			// the number of placement objects in the system at a time, which should be a finite number.
 			// Still, to be on the safer side here KubeFleet sets a cap the number of keys to move per go.
 			break
@@ -196,12 +194,14 @@ func (bq *batchedProcessingPlacementSchedulingQueue) moveAllBatchedItemsToActive
 	}
 
 	for _, key := range keysToMove {
-		// Add the keys to the active queue in batch. Here the implementation does not move keys one by one
-		// right after they are popped as this pattern risks synchronized processing (i.e., a key is popped
-		// from the batched queue, immeidiately added to the active queue and gets processed, then added
-		// back to the batched queue again by one of the watchers before the key moving attempt is finished,
-		// which results in perpetual key moving).
+		// Mark the keys as done in the batched queue and add the keys to the active queue in batch. Here the
+		// implementation keeps the keys in memory first and does not move keys right after they are popped as
+		// this pattern risks synchronized processing (i.e., a key is popped from the batched queue, immeidiately added to the
+		// active queue and gets marked as done by the scheduler, then added back to the batched queue again by
+		// one of the watchers before the key moving attempt is finished, which results in perpetual key moving).
 		bq.active.Add(key)
+		bq.batched.Done(key)
+		bq.batched.Forget(key)
 	}
 }
 
--- a/pkg/scheduler/queue/batched_test.go
+++ b/pkg/scheduler/queue/batched_test.go
@@ -55,7 +55,8 @@ func TestBatchedProcessingPlacementSchedulingQueue_BasicOps(t *testing.T) {
 // TestBatchedProcessingPlacementSchedulingQueue_BatchedOps tests the batched ops
 // (AddBatched) of a batchedProcessingPlacementSchedulingQueue.
 func TestBatchedProcessingPlacementSchedulingQueue_BatchedOps(t *testing.T) {
-	bq := NewBatchedProcessingPlacementSchedulingQueue("TestOnly", nil, nil, 5)
+	movePeriodSeconds := int32(5) // 5 seconds
+	bq := NewBatchedProcessingPlacementSchedulingQueue("TestOnly", nil, nil, movePeriodSeconds)
 	bq.Run()
 
 	addedTimestamp := time.Now()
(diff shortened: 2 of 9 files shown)


pkg/scheduler/queue/batched.go
```diff
@@ -0,0 +1,233 @@
...
+		if shutdown {
+			break
+		}
+		keysToMove = append(keysToMove, placementKey.(PlacementKey))
+
+		if len(keysToMove) > maxNumberOfKeysToMoveFromBatchedToActiveQueuePerGo {
+			// The keys popped from the batched queue are not yet added to the active queue, in other words,
+			// they are not yet marked as done; the batched queue will still track them and adding them
+			// to the batched queue again at this moment will not trigger the batched queue to yield the same
+			// keys again. This implies that the at maximum we will be movine a number of keys equal to
```
weng271190436: moving

pkg/scheduler/queue/batched.go
```diff
@@ -0,0 +1,233 @@
...
+			// Still, to be on the safer side here KubeFleet sets a cap the number of keys to move per go.
+			break
+		}
+	}
+
+	for _, key := range keysToMove {
+		// Add the keys to the active queue in batch. Here the implementation does not move keys one by one
+		// right after they are popped as this pattern risks synchronized processing (i.e., a key is popped
+		// from the batched queue, immeidiately added to the active queue and gets processed, then added
+		// back to the batched queue again by one of the watchers before the key moving attempt is finished,
```
weng271190436: aren't you still moving them one by one here because in between 2 calls of `bq.active.Add(key)`, some watcher goroutine can still run and add the key back? Are you arguing that because this tight loop is happening very fast so it is highly unlikely?

pkg/scheduler/queue/batched.go
```diff
@@ -0,0 +1,233 @@
...
+		// which results in perpetual key moving).
+		bq.active.Add(key)
+	}
+}
+
+// NewBatchedProcessingPlacementSchedulingQueue returns a batchedProcessingPlacementSchedulingQueue.
+func NewBatchedProcessingPlacementSchedulingQueue(name string, activeQRateLimiter, batchedQRateLimiter workqueue.TypedRateLimiter[any], movePeriodSeconds int32) PlacementSchedulingQueue {
```
weng271190436, line 209: can you pass in the option object instead of 5 variables?

I am imagining how this would be used:
```go
NewBatchedProcessingPlacementSchedulingQueue("", nil, nil, nil, 0)
```

which is difficult to read because it is not obvious what "", nil, nil, 0 are supposed to mean

pkg/scheduler/queue/batched.go
```diff
@@ -0,0 +1,233 @@
...
+	// Wait until all the items in the batched queue have been processed..
+	bq.batched.ShutDownWithDrain()
+	// Wait until all the items in the active queue have been processed.
+	bq.active.ShutDownWithDrain()
+}
+
+// NextPlacementKey returns the next PlacementKey (either clusterResourcePlacementKey or resourcePlacementKey)
+// in the work queue for the scheduler to process.
+func (bq *batchedProcessingPlacementSchedulingQueue) NextPlacementKey() (key PlacementKey, closed bool) {
+	// This will block on a condition variable if the queue is empty.
```
weng271190436, line 96: this means that the caller will wait up to timer interval to get the next key? Isn't that 300 seconds?

pkg/scheduler/queue/batched_test.go
```diff
@@ -0,0 +1,183 @@
...
+	for _, key := range keysToAddBatched {
+		bq.AddBatched(key)
+	}
+
+	// Send a move now signal.
+	bqStruct, ok := bq.(*batchedProcessingPlacementSchedulingQueue)
+	if !ok {
+		t.Fatalf("Failed to cast to batchedProcessingPlacementSchedulingQueue")
+	}
+	bqStruct.moveNow <- struct{}{}
```
weng271190436, line 105: Is this intended to be a public method that the scheduler will call? If so it should  be on the public interface?

pkg/scheduler/queue/batched_test.go
```diff
@@ -0,0 +1,183 @@
...
+		keysRecved = append(keysRecved, key)
+		bq.Done(key)
+		bq.Forget(key)
+	}
+
+	if !cmp.Equal(keysToAddBatched, keysRecved) {
+		t.Fatalf("Received keys %v, want %v", keysRecved, keysToAddBatched)
+	}
+	// Allow some buffer time (1 second).
+	if timeSpent := time.Since(moveNowTriggeredTimestamp); timeSpent > 2*time.Second {
```
weng271190436: why is the comment saying 1 second but the code is 2 seconds?

pkg/scheduler/queue/batched_test.go
```diff
@@ -0,0 +1,183 @@
...
+	bq.CloseWithDrain()
+
+	wantKeys := make([]PlacementKey, 0, len(keysToAdd)+len(keysToAddBatched))
+	wantKeys = append(wantKeys, keysToAdd...)
+	wantKeys = append(wantKeys, keysToAddBatched...)
+	if !cmp.Equal(wantKeys, keysRecved) {
+		t.Fatalf("Received keys %v, want %v", keysRecved, wantKeys)
+	}
+	// Allow some buffer time (1 second).
+	if timeSpent := time.Since(closeWithDrainTimestamp); timeSpent > 6*time.Second {
```
weng271190436: Is this 6 second calculated by 5 seconds + 1 second? If so extract the 5 seconds into a variable and instead of saying 6 seconds here say `variable + 1` seconds here?

pkg/scheduler/queue/batched.go
```diff
@@ -0,0 +1,233 @@
...
+
+				// The batched queue might still be running; move all items and re-enter the loop.
+				bq.moveAllBatchedItemsToActiveQueue()
+			case <-timer.C:
+				// The timer has fired; move all items.
+				bq.moveAllBatchedItemsToActiveQueue()
+			}
+
+			// Reset the timer for the next round.
+			timer.Reset(time.Duration(bq.movePeriodSeconds) * time.Second)
```
weng271190436, line 165: why do we reset the timer for `case _, closed := <-bq.moveNow:` ?

pkg/scheduler/queue/batched.go
michaelawyu: Fixed. Thanks 🙏

pkg/scheduler/queue/batched.go
michaelawyu: Hi Wei! It is totally OK for a watcher to add a key back when we are moving it. K8s queues are dirty-writing queues, which means that if we `Get()` a key from the queue first, then `Add()` the same key without marking the key as done, the queue will hold the key (mark it as dirty) instead of pushing it to the queue. The dirty key will only be added when the key has been marked as done.

michaelawyu: For this part of the code, the comment is implying that:

a) if we do something like this:

```go
for ... {
  k := batched.Get()
  active.Add(k)
}
```

there exists a corner case where the consumer (the scheduler) can get really fast and mark the key as done before the moving process completes; if the key has been marked as dirty, it will be pushed to the batched queue again, and the same key will be moved again as the moving process is still on.

b) instead, we keep the keys in memory first, and mark them as done only after the moving process itself has concluded. This guarantees that each key is moved only once.

michaelawyu: I've tweaked the comment/code to be more clear; there's also a sameness situation that has been addressed.

pkg/scheduler/queue/batched.go
michaelawyu: Yeah, let me address this.

pkg/scheduler/queue/batched.go
michaelawyu: Hi Wei! The scheduler queue consists of two separate queues:

* The `active` queue handles changes that need immediate processing
* The `batched` queue handles changes that can be batched and process periodically; we periodically move keys in the `batched` queue to the `active` queue.

It is up to the caller (watchers, etc.) to decide which queue to use. At this moment only member cluster changes are sent to the `batched` queue (`AddBatched()`), other watchers use the `active` queue (`Add()`).

michaelawyu: NextPlacementKey() will only block when the `active` queue is empty.

pkg/scheduler/queue/batched_test.go
michaelawyu: Hi Wei! At this moment I couldn't think of a case where the caller (the scheduler) needs to move all the keys immediately so I haven't exposed the channel. It is currently only used to signal queue shutdowns (and for testing purposes).

pkg/scheduler/queue/batched_test.go
michaelawyu: Fixed. Thanks 🙏

pkg/scheduler/queue/batched_test.go
michaelawyu: I've made the edits to be clearer. Thanks 🙏

pkg/scheduler/queue/batched.go
michaelawyu: Hi Wei! The moving process is set to wait for 300 seconds after each moving attempt; I thought about using a tick-based mechanism (instead of waiting for 300 seconds after each move, do a move every 300 seconds regardless) but it has its own concerns.

michaelawyu: (missing ticks, two consecutive moves if we are unlucky, etc.)

pkg/scheduler/queue/simple.go
```diff
@@ -0,0 +1,133 @@
...
+	sq.active.Add(placementKey)
+}
+
+// Forget untracks a PlacementKey from rate limiter(s) (if any) set up with the queue.
+func (sq *simplePlacementSchedulingQueue) Forget(placementKey PlacementKey) {
+	sq.active.Forget(placementKey)
+}
+
+// NewSimplePlacementSchedulingQueue returns a simplePlacementSchedulingQueue.
+func NewSimplePlacementSchedulingQueue(name string, rateLimiter workqueue.TypedRateLimiter[any]) PlacementSchedulingQueue {
```
ryanzhang-oss, line 120: why change this function signature? AFAIK, this is not part of the new common interface

pkg/scheduler/queue/batched.go
```diff
@@ -0,0 +1,233 @@
...
+// batchedProcessingPlacementSchedulingQueue implements the PlacementSchedulingQueue
+// interface.
+//
+// It consists of two work queues to allow processing for both immediate and batched
+// processing for scheduling related events (changes) of different responsiveness levels.
+type batchedProcessingPlacementSchedulingQueue struct {
+	active  workqueue.TypedRateLimitingInterface[any]
+	batched workqueue.TypedRateLimitingInterface[any]
+
+	moveNow           chan struct{}
```
ryanzhang-oss, line 39: this channel is only closed and never used to pass anything, I wonder why it's called moveNow?

pkg/scheduler/queue/simple.go
michaelawyu: Hi Ryan! Before we were using function based initialization, e.g., `WithName`, `WithRateLimiter`, but now we have two queue implementations (in the same package) this pattern is a bit difficult to follow. I could move them to separate child packages if you prefer continuing using this pattern

pkg/scheduler/queue/batched.go
michaelawyu: Hi Ryan! For testing we did send items to the channel to verify the behavior. Would you prefer a name like `done`, `quit`, or `stop`?

PR #374, opened 2025-12-10
feat: add stopping stage update run implementation

britaniar:

I have:
- Added a check to make sure no cluster start updating and let currently updating cluster finish within a stage before marking the stage and update run as stopped.
- Update the stage condition status as stopping or stopped.
- Update integration tests and added UTs.

- Integration Test
- Unit Test

While waiting for cluster to finish updating, stage and update run will have a progressing unknown condition with the reason as stopping.

### DIFF
--- a/pkg/controllers/updaterun/stop.go
+++ b/pkg/controllers/updaterun/stop.go
@@ -187,7 +187,6 @@ func (r *Reconciler) stopDeleteStage(
 	klog.V(2).InfoS("The delete stage is stopping", "numberOfDeletingClusters", len(toBeDeletedBindings), "updateRun", updateRunRef)
 	allDeletingClustersDeleted := true
 	for _, clusterStatus := range updateRunStatus.DeletionStageStatus.Clusters {
-		klog.InfoS("Checking deletion status of cluster in delete stage", "cluster", clusterStatus.ClusterName, "conditions", clusterStatus.Conditions, "updateRun", updateRunRef)
 		if condition.IsConditionStatusTrue(meta.FindStatusCondition(clusterStatus.Conditions,
 			string(placementv1beta1.ClusterUpdatingConditionStarted)), updateRun.GetGeneration()) && !condition.IsConditionStatusTrue(
 			meta.FindStatusCondition(clusterStatus.Conditions, string(placementv1beta1.ClusterUpdatingConditionSucceeded)),
@@ -197,7 +196,7 @@ func (r *Reconciler) stopDeleteStage(
 		}
 	}
 
-	if allDeletingClustersDeleted || len(toBeDeletedBindings) == 0 {
+	if allDeletingClustersDeleted {
 		markStageUpdatingStopped(updateRunStatus.DeletionStageStatus, updateRun.GetGeneration())
 	}
 	return len(toBeDeletedBindings) == 0, nil
--- a/pkg/controllers/updaterun/stop_test.go
+++ b/pkg/controllers/updaterun/stop_test.go
@@ -26,7 +26,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
@@ -80,14 +79,6 @@ func TestStopUpdatingStage(t *testing.T) {
 							},
 						},
 					},
-					UpdateStrategySnapshot: &placementv1beta1.UpdateStrategySpec{
-						Stages: []placementv1beta1.StageConfig{
-							{
-								Name:           "test-stage",
-								MaxConcurrency: &intstr.IntOrString{Type: intstr.Int, IntVal: 1},
-							},
-						},
-					},
 				},
 			},
 			bindings:     nil,
@@ -129,14 +120,6 @@ func TestStopUpdatingStage(t *testing.T) {
 										},
 									},
 								},
-							},
-						},
-					},
-					UpdateStrategySnapshot: &placementv1beta1.UpdateStrategySpec{
-						Stages: []placementv1beta1.StageConfig{
-							{
-								Name:           "test-stage",
-								MaxConcurrency: &intstr.IntOrString{Type: intstr.Int, IntVal: 1},
 							},
 						},
 					},
@@ -226,12 +209,8 @@ func TestStopUpdatingStage(t *testing.T) {
 				tt.updateRun.Status.StagesStatus[0].Conditions,
 				string(placementv1beta1.StageUpdatingConditionProgressing),
 			)
-			if progressingCond == nil {
-				t.Errorf("stopUpdatingStage() missing progressing condition")
-			} else {
-				if diff := cmp.Diff(tt.wantProgressCond, *progressingCond, cmpOptions...); diff != "" {
-					t.Errorf("stopUpdatingStage() status mismatch: (-want +got):\n%s", diff)
-				}
+			if diff := cmp.Diff(tt.wantProgressCond, *progressingCond, cmpOptions...); diff != "" {
+				t.Errorf("stopUpdatingStage() status mismatch: (-want +got):\n%s", diff)
 			}
 		})
 	}
@@ -246,8 +225,7 @@ func TestStopDeleteStage(t *testing.T) {
 		updateRun           *placementv1beta1.ClusterStagedUpdateRun
 		toBeDeletedBindings []placementv1beta1.BindingObj
 		wantFinished        bool
-		wantError           bool
-		wantAbortError      bool
+		wantError           error
 		wantProgressCond    metav1.Condition
 	}{
 		{
@@ -266,7 +244,7 @@ func TestStopDeleteStage(t *testing.T) {
 			},
 			toBeDeletedBindings: []placementv1beta1.BindingObj{},
 			wantFinished:        true,
-			wantError:           false,
+			wantError:           nil,
 			wantProgressCond: metav1.Condition{
 				Type:               string(placementv1beta1.StageUpdatingConditionProgressing),
 				Status:             metav1.ConditionFalse,
@@ -312,7 +290,7 @@ func TestStopDeleteStage(t *testing.T) {
 				},
 			},
 			wantFinished: false,
-			wantError:    false,
+			wantError:    nil,
 			wantProgressCond: metav1.Condition{
 				Type:               string(placementv1beta1.StageUpdatingConditionProgressing),
 				Status:             metav1.ConditionUnknown,
@@ -357,9 +335,8 @@ func TestStopDeleteStage(t *testing.T) {
 					},
 				},
 			},
-			wantFinished:   false,
-			wantError:      true,
-			wantAbortError: true,
+			wantFinished: false,
+			wantError:    errors.New("the cluster `cluster-1` in the deleting stage is marked as deleting but its corresponding binding is not deleting"),
 			wantProgressCond: metav1.Condition{
 				Type:               string(placementv1beta1.StageUpdatingConditionProgressing),
 				Status:             metav1.ConditionUnknown,
@@ -395,9 +372,8 @@ func TestStopDeleteStage(t *testing.T) {
 					},
 				},
 			},
-			wantFinished:   false,
-			wantError:      false,
-			wantAbortError: false,
+			wantFinished: false,
+			wantError:    nil,
 			wantProgressCond: metav1.Condition{
 				Type:               string(placementv1beta1.StageUpdatingConditionProgressing),
 				Status:             metav1.ConditionFalse,
@@ -462,7 +438,7 @@ func TestStopDeleteStage(t *testing.T) {
 				},
 			},
 			wantFinished: false,
-			wantError:    false,
+			wantError:    nil,
 			wantProgressCond: metav1.Condition{
 				Type:               string(placementv1beta1.StageUpdatingConditionProgressing),
 				Status:             metav1.ConditionUnknown,
@@ -483,15 +459,16 @@ func TestStopDeleteStage(t *testing.T) {
 				t.Errorf("stopDeleteStage() finished = %v, want %v", gotFinished, tt.wantFinished)
 			}
 
-			// Check error expectations.
-			if tt.wantError {
-				if gotErr == nil {
-					t.Errorf("stopDeleteStage() error = nil, want error")
-				} else if tt.wantAbortError && !errors.Is(gotErr, errStagedUpdatedAborted) {
-					t.Errorf("stopDeleteStage() error = %v, want errStagedUpdatedAborted", gotErr)
+			// Verify error expectation.
+			if (tt.wantError != nil) != (gotErr != nil) {
+				t.Fatalf("stopUpdatingStage() want error: %v, got error: %v", tt.wantError, gotErr)
+			}
+
+			// Verify error message contains expected substring.
+			if tt.wantError != nil && gotErr != nil {
+				if !strings.Contains(gotErr.Error(), tt.wantError.Error()) {
+					t.Fatalf("stopUpdatingStage() want error: %v, got error: %v", tt.wantError, gotErr)
 				}
-			} else if gotErr != nil {
-				t.Errorf("stopDeleteStage() error = %v, want nil", gotErr)
 			}
 
 			// Check stage status condition.
@@ -499,12 +476,8 @@ func TestStopDeleteStage(t *testing.T) {
 				tt.updateRun.Status.DeletionStageStatus.Conditions,
 				string(placementv1beta1.StageUpdatingConditionProgressing),
 			)
-			if progressingCond == nil {
-				t.Errorf("stopDeleteStage() missing progressing condition")
-			} else {
-				if diff := cmp.Diff(tt.wantProgressCond, *progressingCond, cmpOptions...); diff != "" {
-					t.Errorf("stopDeleteStage() status mismatch: (-want +got):\n%s", diff)
-				}
+			if diff := cmp.Diff(tt.wantProgressCond, *progressingCond, cmpOptions...); diff != "" {
+				t.Errorf("stopDeleteStage() status mismatch: (-want +got):\n%s", diff)
 			}
 		})
 	}


apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -427,7 +427,7 @@ const (
 	// Its condition status can be one of the following:
 	// - "True": The staged update run is making progress.
 	// - "False": The staged update run is waiting/paused/abandoned.
-	// - "Unknown" means it is unknown.
+	// - "Unknown": The staged update run is in the process of stopping.
```
jwtty: We have the `Reason` field to tell it's in the process of stopping. Here I'd prefer to say it's in a transitioning state.

pkg/controllers/updaterun/controller.go
```diff
@@ -202,6 +202,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	return runtime.Result{}, nil
 }
 
+func (r *Reconciler) handleIncompleteUpdateRun(ctx context.Context, updateRun placementv1beta1.UpdateRunObj, waitTime time.Duration, err error, state placementv1beta1.State, runObjRef klog.ObjectRef) (runtime.Result, error) {
```
Arvindthiru, line 205: Lets rename it to handleIncompleteUpdateRun, I misunderstood the flow. handleIncompleteUpdateRun makes more sense

pkg/controllers/updaterun/controller.go
Arvindthiru, line 172:
```suggestion
if execErr != nil {
  if errors.Is(execErr, errStagedUpdatedAborted) {
    // errStagedUpdatedAborted cannot be retried.
    return runtime.Result{}, r.recordUpdateRunFailed(ctx, updateRun, execErr.Error())
  }
  return runtime.Result{}, err
}
```

pkg/controllers/updaterun/execution.go
```diff
@@ -232,9 +224,7 @@ func (r *Reconciler) executeUpdatingStage(
 				}
 			}
 			markClusterUpdatingStarted(clusterStatus, updateRun.GetGeneration())
-			if finishedClusterCount == 0 {
-				markStageUpdatingStarted(updatingStageStatus, updateRun.GetGeneration())
-			}
+			markStageUpdatingStarted(updatingStageStatus, updateRun.GetGeneration())
```
Arvindthiru: Why are we always setting markStageUpdatingStarted now?

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,226 @@
...
+		clusterStatus := &updatingStageStatus.Clusters[i]
+		clusterStartedCond := meta.FindStatusCondition(clusterStatus.Conditions, string(placementv1beta1.ClusterUpdatingConditionStarted))
+		if clusterStartedCond == nil {
+			// Cluster has not started updating therefore no need to do anything.
+			continue
+		}
+
+		clusterUpdateSucceededCond := meta.FindStatusCondition(clusterStatus.Conditions, string(placementv1beta1.ClusterUpdatingConditionSucceeded))
+		if condition.IsConditionStatusFalse(clusterUpdateSucceededCond, updateRun.GetGeneration()) || condition.IsConditionStatusTrue(clusterUpdateSucceededCond, updateRun.GetGeneration()) {
+			// The cluster has already been updated.
```
Arvindthiru:
```suggestion
			// The cluster has already been updated or failed to update.
```

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,226 @@
...
+		Reason:             condition.StageUpdatingStoppedReason,
+		Message:            "All the updating clusters have finished updating, the stage is now stopped, waiting to be resumed",
+	})
+}
+
+func checkIfErrorStagedUpdateAborted(err error, updateRun placementv1beta1.UpdateRunObj, updatingStageStatus *placementv1beta1.StageUpdatingStatus) {
+	updateRunStatus := updateRun.GetUpdateRunStatus()
+	if errors.Is(err, errStagedUpdatedAborted) {
+		if updatingStageStatus != nil {
+			klog.InfoS("The update run is aborted due to unexpected behavior in updating stage, marking the stage as failed", "stage", updatingStageStatus.StageName, "updateRun", klog.KObj(updateRun))
```
jwtty: The error may not be unexpected, should be `unrecoverable`.

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	clusterUpdatingCount := 0
+	var stuckClusterNames []string
+	var clusterUpdateErrors []error
+	// Go through each cluster in the stage and check if it's updating/succeeded/failed/not started.
+	for i := 0; i < len(updatingStageStatus.Clusters); i++ {
+		clusterStatus := &updatingStageStatus.Clusters[i]
+		clusterStartedCond := meta.FindStatusCondition(clusterStatus.Conditions, string(placementv1beta1.ClusterUpdatingConditionStarted))
+		if clusterStartedCond == nil {
```
Arvindthiru: Can we do a sanity check to ensure clusterStartedCond is set to true ?

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,226 @@
...
+// markUpdateRunStopping marks the update run as stopping in memory.
+func markUpdateRunStopping(updateRun placementv1beta1.UpdateRunObj) {
+	klog.V(2).InfoS("Marking the update run as stopping", "updateRun", klog.KObj(updateRun))
+	updateRunStatus := updateRun.GetUpdateRunStatus()
+	meta.SetStatusCondition(&updateRunStatus.Conditions, metav1.Condition{
+		Type:               string(placementv1beta1.StagedUpdateRunConditionProgressing),
+		Status:             metav1.ConditionUnknown,
+		ObservedGeneration: updateRun.GetGeneration(),
+		Reason:             condition.UpdateRunStoppingReason,
+		Message:            "The update run is the process of stopping, waiting for all the updating clusters to finish updating before completing the stop process",
```
jwtty:
```suggestion
		Message:            "The update run is the process of stopping, waiting for all the updating/deleting clusters to finish before completing the stop process",
```

pkg/controllers/updaterun/execution.go
britaniar: If it was previous false it will update to true (progressing), but it is already true then it won't do anything. But this makes sure it has been marked progressing true when we find any updating cluster.

pkg/controllers/updaterun/execution_integration_test.go
```diff
@@ -785,47 +785,71 @@ var _ = Describe("UpdateRun execution tests - double stages", func() {
 			validateUpdateRunMetricsEmitted(wantMetrics...)
 		})
 
-		It("Should stop the in middle of cluster updating when update run state is Stop", func() {
+		It("Should be stopping the in middle of cluster updating when update run state is Stop", func() {
```
jwtty: Please move stop related tests to a new file `stop_integration_test.go`, thanks!

pkg/controllers/updaterun/stop.go
jwtty: or startedCond is false.

pkg/controllers/updaterun/execution.go
```diff
@@ -564,6 +554,9 @@ func calculateMaxConcurrencyValue(status *placementv1beta1.UpdateRunStatus, stag
 func aggregateUpdateRunStatus(updateRun placementv1beta1.UpdateRunObj, stageName string, stuckClusterNames []string) {
 	if len(stuckClusterNames) > 0 {
 		markUpdateRunStuck(updateRun, stageName, strings.Join(stuckClusterNames, ", "))
+	} else if updateRun.GetUpdateRunSpec().State == placementv1beta1.StateStop {
```
Arvindthiru: This check shouldn't be inside aggregateUpdateRunStatus, it's not aggregating here

britaniar: We call this in the stopUpdatingStage as well in case we need to mark the update run as stuck. What it originally did was if there were no stuck clusters then it would mark it prepressing, but if we are in stop state we do not want to mark as progressing since we are stopping... what do you suggest we do then?

pkg/controllers/updaterun/execution.go
britaniar: I talked about this with Wantong and this was what she had suggested doing. If we keep the previous check then when we restart and we have clusters that have succeed then the finished cluster count is not equal to 0, but we need to mark it as starting since the stage is resuming (progressing).

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,226 @@
...
+		existingDeleteStageClusterMap[existingDeleteStageStatus.Clusters[i].ClusterName] = &existingDeleteStageStatus.Clusters[i]
+	}
+	// Mark the delete stage as stopping in case it's not.
+	markStageUpdatingStopping(existingDeleteStageStatus, updateRun.GetGeneration())
+
+	for _, binding := range toBeDeletedBindings {
+		bindingSpec := binding.GetBindingSpec()
+		curCluster, exist := existingDeleteStageClusterMap[bindingSpec.TargetCluster]
+		if !exist {
+			// The cluster is not in the delete stage. This happens when the update run is pauseed as delete stage starts.
```
jwtty: I don't get this comment. In executeDeleteStage, this is an unexpected err.

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,226 @@
...
+
+	if clusterUpdatingCount == 0 {
+		// All the clusters in the stage have finished updating or not started.
+		markStageUpdatingStopped(updatingStageStatus, updateRun.GetGeneration())
+		klog.InfoS("The stage has finished all clusters updating", "stage", updatingStageStatus.StageName, "updateRun", updateRunRef)
+		return true, 0, nil
+	}
+	// Some clusters are still updating.
+	klog.InfoS("The updating stage is waiting for updating clusters to finish before completely stopping", "numberOfUpdatingClusters", clusterUpdatingCount, "stage", updatingStageStatus.StageName, "updateRun", updateRunRef)
+	markStageUpdatingStopping(updatingStageStatus, updateRun.GetGeneration())
```
Arvindthiru: Should we move this to the top this function ?

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,226 @@
...
+				klog.ErrorS(unexpectedErr, "The binding should be deleting before we mark a cluster deleting", "clusterStatus", curCluster, "updateRun", updateRunRef)
+				return false, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
+			}
+			return false, nil
+		}
+	}
+
+	klog.V(2).InfoS("The delete stage is stopping", "numberOfDeletingClusters", len(toBeDeletedBindings), "updateRun", updateRunRef)
+	if len(toBeDeletedBindings) == 0 {
+		markStageUpdatingStopped(updateRunStatus.DeletionStageStatus, updateRun.GetGeneration())
```
jwtty, line 200: Is this part missed? https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/controllers/updaterun/execution.go#L378C2-L384C3 You need to mark deleted clusters to be succeeded.

pkg/controllers/updaterun/execution.go
Arvindthiru: Then this check can be made after we call aggregateUpdateRunStatus in the stopUpdatingStage ?, we can update aggregateUpdateRunStatus to only mark UpdateRun as progressing when state is Execute

pkg/controllers/updaterun/execution.go
Arvindthiru: Do we set Started condition to false for a stage ?

britaniar: While we are trying to stop ("stopping") the progressing condition is unknown, and when it is done stopping ("stopped") the progressing condition is false for the stage and the update run. Only the stage that in the middle of being updated when stopped is updated with these conditions.

pkg/controllers/updaterun/controller.go
britaniar: updateUpdateRunStatus checks if its not nil and returns the error already. Plus I think we need to update the status anyways if we get an error

pkg/controllers/updaterun/stop.go
britaniar: why are we checking if it's true when the this is checking if it is nil? I think you mean false.

pkg/controllers/updaterun/stop.go
britaniar: I'll move the marking func to the top but I think the log should stay here because we should let the user know if there clusters updating after we are done going through the clusters not before.

pkg/controllers/updaterun/stop.go
britaniar: Oh must have missed that. Thanks for catching that.

pkg/controllers/updaterun/execution.go
Arvindthiru: Lets rename markStageUpdatingStarted to markStageUpdatingProgressStarted, I got confused by the name and thought stage has a Started condition but in this case we set Progressing condition to true with Started reason

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,237 @@
...
+	meta.SetStatusCondition(&stageUpdatingStatus.Conditions, metav1.Condition{
+		Type:               string(placementv1beta1.StageUpdatingConditionProgressing),
+		Status:             metav1.ConditionFalse,
+		ObservedGeneration: generation,
+		Reason:             condition.StageUpdatingStoppedReason,
+		Message:            "All the updating clusters have finished updating, the stage is now stopped, waiting to be resumed",
+	})
+}
+
+func checkIfErrorStagedUpdateAborted(err error, updateRun placementv1beta1.UpdateRunObj, updatingStageStatus *placementv1beta1.StageUpdatingStatus) {
```
Arvindthiru, line 240: nit: we should move this to a common util file since it's used across execution, stop states

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,237 @@
...
+	// Mark the stage as stopping in case it's not.
+	markStageUpdatingStopping(updatingStageStatus, updateRun.GetGeneration())
+	clusterUpdatingCount := 0
+	var stuckClusterNames []string
+	var clusterUpdateErrors []error
+	// Go through each cluster in the stage and check if it's updating/succeeded/failed/not started.
+	for i := 0; i < len(updatingStageStatus.Clusters); i++ {
+		clusterStatus := &updatingStageStatus.Clusters[i]
+		clusterStartedCond := meta.FindStatusCondition(clusterStatus.Conditions, string(placementv1beta1.ClusterUpdatingConditionStarted))
+		if clusterStartedCond == nil || condition.IsConditionStatusFalse(clusterStartedCond, updateRun.GetGeneration()) {
```
Arvindthiru: We never seem to mark ClusterUpdatingConditionStarted as false, https://github.com/Arvindthiru/kubefleet/blob/9961b09f4f716c97d02619bed8f6f071e39b0e08/apis/placement/v1beta1/stageupdate_types.go#L473-L476

This is what I got confused with https://github.com/kubefleet-dev/kubefleet/pull/374/changes#r2615781769

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+		if condition.IsConditionStatusTrue(meta.FindStatusCondition(curCluster.Conditions, string(placementv1beta1.ClusterUpdatingConditionStarted)), updateRun.GetGeneration()) {
+			// The cluster status is marked as being deleted.
+			if binding.GetDeletionTimestamp().IsZero() {
+				// The cluster is marked as deleting but the binding is not deleting.
+				unexpectedErr := controller.NewUnexpectedBehaviorError(fmt.Errorf("the cluster `%s` in the deleting stage is marked as deleting but its corresponding binding is not deleting", curCluster.ClusterName))
+				klog.ErrorS(unexpectedErr, "The binding should be deleting before we mark a cluster deleting", "clusterStatus", curCluster, "updateRun", updateRunRef)
+				return false, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
+			}
+			return false, nil
```
Arvindthiru: Why do we return here ?, we just know that this binding is deleting at the moment why not check other toBeDeletedBindings ?

pkg/controllers/updaterun/stop.go
britaniar: Wantong wanted me to add this check. It is like the sanity check you wanted here https://github.com/kubefleet-dev/kubefleet/pull/374#discussion_r2612447168. I'll change it to !true

pkg/controllers/updaterun/stop.go
britaniar: I think this should be fine. there are some func used across initialization.go and validation.go and they are in the initialization file. Plus, this uses `errStagedUpdatedAborted` which would have to be exported to use elsewhere when we only use it in update run.
Moving it would make more sense if it was across directories, but it is used only in updaterun.

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,237 @@
...
+		Type:               string(placementv1beta1.StageUpdatingConditionProgressing),
+		Status:             metav1.ConditionFalse,
+		ObservedGeneration: generation,
+		Reason:             condition.StageUpdatingStoppedReason,
+		Message:            "All the updating clusters have finished updating, the stage is now stopped, waiting to be resumed",
+	})
+}
+
+func checkIfErrorStagedUpdateAborted(err error, updateRun placementv1beta1.UpdateRunObj, updatingStageStatus *placementv1beta1.StageUpdatingStatus) {
+	updateRunStatus := updateRun.GetUpdateRunStatus()
```
Arvindthiru: nit: This line can be moved to 233

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,237 @@
...
+	for i := 0; i < len(updatingStageStatus.Clusters); i++ {
+		clusterStatus := &updatingStageStatus.Clusters[i]
+		clusterStartedCond := meta.FindStatusCondition(clusterStatus.Conditions, string(placementv1beta1.ClusterUpdatingConditionStarted))
+		if !condition.IsConditionStatusTrue(clusterStartedCond, updateRun.GetGeneration()) {
+			// Cluster has not started updating therefore no need to do anything.
+			continue
+		}
+
+		clusterUpdateSucceededCond := meta.FindStatusCondition(clusterStatus.Conditions, string(placementv1beta1.ClusterUpdatingConditionSucceeded))
+		if condition.IsConditionStatusFalse(clusterUpdateSucceededCond, updateRun.GetGeneration()) || condition.IsConditionStatusTrue(clusterUpdateSucceededCond, updateRun.GetGeneration()) {
```
Arvindthiru, line 88: If an update to a cluster fails, should we abort here ?, If not can you add a comment explaining why

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,237 @@
...
+		if !exist {
+			// This is unexpected because we already checked in validation.
+			missingErr := controller.NewUnexpectedBehaviorError(fmt.Errorf("the to be deleted cluster `%s` is not in the deleting stage during stopping", bindingSpec.TargetCluster))
+			klog.ErrorS(missingErr, "The cluster in the deleting stage does not include all the to be deleted binding", "updateRun", updateRunRef)
+			return false, fmt.Errorf("%w: %s", errStagedUpdatedAborted, missingErr.Error())
+		}
+		// In validation, we already check the binding must exist in the status.
+		delete(existingDeleteStageClusterMap, bindingSpec.TargetCluster)
+		if condition.IsConditionStatusTrue(meta.FindStatusCondition(curCluster.Conditions, string(placementv1beta1.ClusterUpdatingConditionSucceeded)), updateRun.GetGeneration()) {
+			// The cluster status is marked as deleted.
```
Arvindthiru, line 161: We should be aborting here as per deleteStage logic in execute. Since this case shouldn't happen as we only markClusterUpdatingSucceeded for a cluster if it's not present in toBeDeletedBindings in line 181

pkg/controllers/updaterun/stop_test.go
```diff
@@ -0,0 +1,263 @@
...
+			} else if gotErr != nil {
+				t.Errorf("stopDeleteStage() error = %v, want nil", gotErr)
+			}
+
+			// Check stage status condition.
+			progressingCond := meta.FindStatusCondition(
+				tt.updateRun.Status.DeletionStageStatus.Conditions,
+				string(placementv1beta1.StageUpdatingConditionProgressing),
+			)
+			if progressingCond == nil {
```
Arvindthiru: Lets construct wantProgressingCond and use cmp.Diff to compare got, want

pkg/controllers/updaterun/stop_integration_test.go
```diff
@@ -0,0 +1,852 @@
...
+
+			By("Validating the 1st stage has startTime set")
+			Expect(updateRun.Status.StagesStatus[0].StartTime).ShouldNot(BeNil())
+
+			By("Checking update run status metrics are emitted")
+			wantMetrics = append(wantMetrics, generateProgressingMetric(updateRun))
+			validateUpdateRunMetricsEmitted(wantMetrics...)
+		})
+
+		It("Should be stopping the in middle of cluster updating when update run state is Stop", func() {
```
Arvindthiru:
```suggestion
		It("Should be stopping in the middle of cluster updating when update run state is Stop", func() {
```

pkg/controllers/updaterun/stop_integration_test.go
```diff
@@ -0,0 +1,852 @@
...
+				}
+				return condition.IsConditionStatusTrue(meta.FindStatusCondition(approvalRequest.Status.Conditions, string(placementv1beta1.ApprovalRequestConditionApprovalAccepted)), approvalRequest.Generation), nil
+			}, timeout, interval).Should(BeTrue(), "failed to validate the approvalRequest approval accepted")
+
+			By("Checking update run status metrics are emitted")
+			wantMetrics = append(wantMetrics, generateWaitingMetric(updateRun), generateProgressingMetric(updateRun))
+			validateUpdateRunMetricsEmitted(wantMetrics...)
+		})
+
+		It("Should delete all the clusterResourceBindings in the delete stage and complete the update run", func() {
```
Arvindthiru, line 803: Can we add a case where we stop in between the deletion stage ?

pkg/controllers/updaterun/stop_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
+)
+
+func TestStopDeleteStage(t *testing.T) {
```
Arvindthiru, line 219: Can we add UTs for stopUpdatingStage as well ?, if not necessary please explain why ?

britaniar: stopUpdatingStage is covered by the integration test. I added this UT to help with code coverage.

pkg/controllers/updaterun/stop_integration_test.go
britaniar: I tried previously and we have no control over the clusters in the stage so the stop timing here is difficult.

pkg/controllers/updaterun/stop.go
jwtty: We can continue if a cluster update fails, like our release pipeline, though we currently does not have a very confident way to say cluster update has failed. We currently only fail when the cluster is preempted by another updateRun but that would abort the updateRun during execution.

jwtty: basically during stop I think we don't need to worry about a cluster that is already in a terminal state.

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,237 @@
...
+			markClusterUpdatingStarted(clusterStatus, updateRun.GetGeneration())
+		}
+		markClusterUpdatingSucceeded(clusterStatus, updateRun.GetGeneration())
+	}
+
+	klog.V(2).InfoS("The delete stage is stopping", "numberOfDeletingClusters", len(toBeDeletedBindings), "updateRun", updateRunRef)
+	if len(toBeDeletedBindings) == 0 {
+		markStageUpdatingStopped(updateRunStatus.DeletionStageStatus, updateRun.GetGeneration())
+	}
+	return len(toBeDeletedBindings) == 0, nil
```
jwtty, line 202: what if some binding is never deleted at all? Stopping will never finish.

pkg/controllers/updaterun/stop_test.go
```diff
@@ -0,0 +1,263 @@
...
+				},
+			},
+			wantFinished:    false,
+			wantError:       true,
+			wantAbortError:  true,
+			wantStageStatus: metav1.ConditionUnknown,
+			wantReason:      condition.StageUpdatingStoppingReason,
+		},
+		{
+			name: "multiple clusters with mixed states",
```
jwtty, line 385: What about clustr not marked and binding not deleting?

pkg/controllers/updaterun/stop_integration_test.go
```diff
@@ -0,0 +1,852 @@
...
+			meta.SetStatusCondition(&wantStatus.Conditions, generateProgressingUnknownConditionWithReason(updateRun, condition.UpdateRunStoppingReason))
+			validateClusterStagedUpdateRunStatus(ctx, updateRun, wantStatus, "")
+
+			By("Checking update run status metrics are emitted")
+			wantMetrics = append(wantMetrics, generateStoppingMetric(updateRun))
+			validateUpdateRunMetricsEmitted(wantMetrics...)
+		})
+
+		It("Should wait for cluster to finish updating before update run is completely stopped", func() {
+			By("Validating the 2nd clusterResourceBinding is  updated to Bound")
```
jwtty:
```suggestion
			By("Validating the 2nd clusterResourceBinding is updated to Bound")
```

pkg/controllers/updaterun/stop_integration_test.go
```diff
@@ -0,0 +1,852 @@
...
+			// Mark updateRun progressing condition as unknown with stopping reason.
+			meta.SetStatusCondition(&wantStatus.Conditions, generateProgressingUnknownConditionWithReason(updateRun, condition.UpdateRunStoppingReason))
+			validateClusterStagedUpdateRunStatus(ctx, updateRun, wantStatus, "")
+
+			By("Checking update run status metrics are emitted")
+			wantMetrics = append(wantMetrics, generateStoppingMetric(updateRun))
+			validateUpdateRunMetricsEmitted(wantMetrics...)
+		})
+
+		It("Should wait for cluster to finish updating before update run is completely stopped", func() {
```
jwtty, line 253: This "It" section is redundant unless you want to check updateRun is in stopping state CONSISTENTLY.

pkg/controllers/updaterun/stop_integration_test.go
```diff
@@ -0,0 +1,852 @@
...
+
+		It("Should not continue to delete stage after approval when still stopped", func() {
+			By("Approving the approvalRequest")
+			approveClusterApprovalRequest(ctx, wantApprovalRequest.Name)
+
+			By("Validating the to-be-deleted bindings are NOT deleted")
+			Eventually(func() error {
+				for i := numTargetClusters; i < numTargetClusters+numUnscheduledClusters; i++ {
+					binding := &placementv1beta1.ClusterResourceBinding{}
+					if err := k8sClient.Get(ctx, types.NamespacedName{Name: resourceBindings[i].Name}, binding); err != nil {
```
jwtty, line 741: Also verify the bindings do not have DeletionTimestamp?

pkg/controllers/updaterun/stop_integration_test.go
jwtty: We can add a finalizer on the bindings so that they are not deleted.

pkg/controllers/updaterun/stop_integration_test.go
```diff
@@ -0,0 +1,852 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	clusterv1beta1 "github.com/kubefleet-dev/kubefleet/apis/cluster/v1beta1"
+	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
+)
+
+var _ = Describe("UpdateRun stop tests", func() {
```
jwtty, line 37: Not a blocker to this PR, but this test is too complicated, including many unnecessary tests for stopping. If we can create a small number of updating/deleting clusters and a simpler strategy to just test the stop cases, that's better. We actually have one https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/controllers/updaterun/controller_integration_test.go#L473C1-L490C2 This one does not have any to-be-deleted bindings though. Please feel free to add if you want to update.

pkg/controllers/updaterun/stop_test.go
britaniar: I added UT for the parts that are not covered.

pkg/controllers/updaterun/stop.go
```diff
@@ -0,0 +1,252 @@
...
+		if condition.IsConditionStatusTrue(meta.FindStatusCondition(clusterStatus.Conditions,
+			string(placementv1beta1.ClusterUpdatingConditionStarted)), updateRun.GetGeneration()) && !condition.IsConditionStatusTrue(
+			meta.FindStatusCondition(clusterStatus.Conditions, string(placementv1beta1.ClusterUpdatingConditionSucceeded)),
+			updateRun.GetGeneration()) {
+			allDeletingClustersDeleted = false
+			break
+		}
+	}
+
+	if allDeletingClustersDeleted || len(toBeDeletedBindings) == 0 {
```
Arvindthiru: Do we need to check if len(toBeDeletedBindings), if len(toBeDeletedBindings) == 0 wouldn't allDeletingClustersDeleted be true ?

pkg/controllers/updaterun/stop_test.go
```diff
@@ -18,16 +18,225 @@ package updaterun
...
+			if waitTime != tt.wantWaitTime {
+				t.Fatalf("stopUpdatingStage() want waitTime: %v, got waitTime: %v", tt.wantWaitTime, waitTime)
+			}
+
+			// Verify progressing condition.
+			progressingCond := meta.FindStatusCondition(
+				tt.updateRun.Status.StagesStatus[0].Conditions,
+				string(placementv1beta1.StageUpdatingConditionProgressing),
+			)
+			if progressingCond == nil {
```
Arvindthiru: cmp.Diff can handle nil values passed to it

pkg/controllers/updaterun/stop_test.go
```diff
@@ -0,0 +1,511 @@
...
+											Status:             metav1.ConditionFalse,
+											ObservedGeneration: 1,
+											Reason:             condition.ClusterUpdatingFailedReason,
+										},
+									},
+								},
+							},
+						},
+					},
+					UpdateStrategySnapshot: &placementv1beta1.UpdateStrategySpec{
```
Arvindthiru: nit: do we need UpdateStrategySnapshot for this UT ?

pkg/controllers/updaterun/stop_test.go
```diff
@@ -0,0 +1,511 @@
...
+
+			gotFinished, gotErr := r.stopDeleteStage(tt.updateRun, tt.toBeDeletedBindings)
+
+			// Check finished result.
+			if gotFinished != tt.wantFinished {
+				t.Errorf("stopDeleteStage() finished = %v, want %v", gotFinished, tt.wantFinished)
+			}
+
+			// Check error expectations.
+			if tt.wantError {
```
Arvindthiru: Lets check the error similar to TestStopUpdatingStage, i.e. check the error msg instead of checking to see if error is errStagedUpdatedAborted, if we construct the wantErr correctly we don't need to use errors.Is ?

PR #378, opened 2025-12-15
fix: use the mc generation when copying the agentStatus

zhiying-lin:

The wrong observedGeneration generated wrong log info about healthy condition.

> 2025-12-10T10:10:23.355648245Z stderr F I1210 10:10:23.355609       1 v1beta1/membercluster_controller.go:542] "Update the memberCluster status" memberCluster="kind-cluster-2" joined=true healthy=false lastReceivedHeartbeat="2025-12-10T10:10:23Z"

https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/controllers/membercluster/v1beta1/membercluster_controller.go#L542

The bug won't cause the scheduling failure, as we did not check the generation when checking the cluster status in https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/scheduler/clustereligibilitychecker/checker.go#L127

Fixes #

I have:

Updated the unit tests and validated in the e2e tests

```
grep -rin "healthy=false" .
./hub-agent-hub-agent_0.log.20251215-065512.log:22435:2025-12-15T06:54:45.501260198Z stderr F I1215 06:54:45.501201       1 v1beta1/membercluster_controller.go:552] "Update the memberCluster status" memberCluster="mc-11" joined=false healthy=false lastReceivedHeartbeat="nil"
```
the healthy will be false when the agent has not joined yet.

### DIFF
--- a/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
+++ b/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
@@ -515,8 +515,18 @@ func (r *Reconciler) syncInternalMemberClusterStatus(imc *clusterv1beta1.Interna
 	}
 
 	// TODO: We didn't handle condition type: clusterv1beta1.ConditionTypeMemberClusterHealthy.
-	// Copy Agent status.
-	mc.Status.AgentStatus = imc.Status.AgentStatus
+	// Copy Agent status and set ObservedGeneration for agent conditions.
+	if len(imc.Status.AgentStatus) > 0 {
+		mc.Status.AgentStatus = make([]clusterv1beta1.AgentStatus, len(imc.Status.AgentStatus))
+	}
+	for i := range imc.Status.AgentStatus {
+		mc.Status.AgentStatus[i] = *imc.Status.AgentStatus[i].DeepCopy()
+		// Set ObservedGeneration for each agent condition.
+		for j := range mc.Status.AgentStatus[i].Conditions {
+			mc.Status.AgentStatus[i].Conditions[j].ObservedGeneration = mc.GetGeneration()
+		}
+	}
+
 	r.aggregateJoinedCondition(mc)
 	// Copy resource usages.
 	mc.Status.ResourceUsage = imc.Status.ResourceUsage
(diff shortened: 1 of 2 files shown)


pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -515,8 +515,18 @@ func (r *Reconciler) syncInternalMemberClusterStatus(imc *clusterv1beta1.Interna
...
-	mc.Status.AgentStatus = imc.Status.AgentStatus
+	// Copy Agent status and set ObservedGeneration for agent conditions.
+	if len(imc.Status.AgentStatus) > 0 {
+		mc.Status.AgentStatus = make([]clusterv1beta1.AgentStatus, len(imc.Status.AgentStatus))
+	}
+	for i := range imc.Status.AgentStatus {
+		mc.Status.AgentStatus[i] = *imc.Status.AgentStatus[i].DeepCopy()
+		// Set ObservedGeneration for each agent condition.
+		for j := range mc.Status.AgentStatus[i].Conditions {
+			mc.Status.AgentStatus[i].Conditions[j].ObservedGeneration = mc.GetGeneration()
```
weng271190436, line 526: Can you comment on why we have to deep copy the IMC's agent statuses but mutate the observed generation?

I think it is difficult for someone reading this piece of code to make the connection between this and the log line that indicates the healthiness of the member agent.

Even if someone reads line 546, the person would need to understand that `IsConditionStatusTrue` is not just about condition status being true, but also about observedGeneration matching latestGeneration.

zhiying-lin: no matter how we used it,  we should always use the generation of the mc.

It's defined in the ObservedGeneration of the condition.

```
// observedGeneration represents the .metadata.generation that the condition was set based upon.
	// For instance, if .metadata.generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date
	// with respect to the current state of the instance.
	// +optional
	// +kubebuilder:validation:Minimum=0
```
The bug was introduced when we just started writing the k8 code.

ryanzhang-oss (approved):

PR #380, opened 2025-12-16
feat: decouple informer cache population and event handling

weng271190436:

decouple informer cache population and event handling

This is a pre-requisite for https://github.com/kubefleet-dev/kubefleet/pull/366 to make validating webhook on follower hub agents work.

If I want to run a leader hub agent and multiple followers, the followers need to run webhook validation, which requires informer cache. However, the informer cache is populated by change detector, which only runs on leader replica.

Previously, change detector calls AddDynamicResources to add informer and attach event handler

Now, I am changing it so that a new controller informer populator starts the informer caches while change detector attaches event handler to the informer caches.

I have:

### DIFF
--- a/pkg/utils/informer/informermanager.go
+++ b/pkg/utils/informer/informermanager.go
@@ -205,25 +205,6 @@ func (s *informerManagerImpl) Stop() {
 	s.cancel()
 }
 
-// getOrCreateInformerWithTransform gets or creates an informer for the given resource and ensures
-// the ManagedFields transform is set. This is idempotent - if the informer exists, we get the same
-// instance.
-func (s *informerManagerImpl) getOrCreateInformerWithTransform(resource schema.GroupVersionResource) cache.SharedIndexInformer {
-	// Get or create the informer (this is idempotent - if it exists, we get the same instance)
-	// The idempotent behavior is important because this method may be called multiple times,
-	// potentially concurrently, and relies on the shared informer instance from the factory.
-	informer := s.informerFactory.ForResource(resource).Informer()
-
-	// Set the transform to strip ManagedFields. This is safe to call even if
-	// already set, since we get the same informer instance. If the informer has already
-	// started, this will fail silently (which is fine).
-	if err := informer.SetTransform(ctrlcache.TransformStripManagedFields()); err != nil {
-		klog.V(4).InfoS("Transform already set or informer started", "gvr", resource, "err", err)
-	}
-
-	return informer
-}
-
 // AddEventHandlerToInformer adds an event handler to an existing informer for the given resource.
 // If the informer doesn't exist, it will be created. This is used by the leader's ChangeDetector
 // to add event handlers to informers that were created by the InformerPopulator.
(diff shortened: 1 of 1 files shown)


pkg/utils/informer/informermanager.go
```diff
@@ -255,6 +205,50 @@ func (s *informerManagerImpl) Stop() {
 	s.cancel()
 }
 
+func (s *informerManagerImpl) AddEventHandlerToInformer(resource schema.GroupVersionResource, handler cache.ResourceEventHandler) {
+	s.resourcesLock.Lock()
```
ryanzhang-oss: why lock this when this function didn't touch s.apiResources. Is the informerFactory.ForResource() function thread safe by itself?

pkg/utils/informer/informermanager.go
```diff
@@ -255,6 +205,50 @@ func (s *informerManagerImpl) Stop() {
...
+	defer s.resourcesLock.Unlock()
+
+	// Get or create the informer (this is idempotent - if it exists, we get the same instance)
+	// The idempotent behavior is important because this method may be called multiple times,
+	// potentially concurrently, and relies on the shared informer instance from the factory.
+	informer := s.informerFactory.ForResource(resource).Informer()
+
+	// Add the event handler to the informer
+	_, _ = informer.AddEventHandler(handler)
```
ryanzhang-oss: what are the return?

pkg/utils/informer/informermanager.go
```diff
@@ -255,6 +205,50 @@ func (s *informerManagerImpl) Stop() {
...
 }
 
+func (s *informerManagerImpl) AddEventHandlerToInformer(resource schema.GroupVersionResource, handler cache.ResourceEventHandler) {
+	s.resourcesLock.Lock()
+	defer s.resourcesLock.Unlock()
+
+	// Get or create the informer (this is idempotent - if it exists, we get the same instance)
+	// The idempotent behavior is important because this method may be called multiple times,
+	// potentially concurrently, and relies on the shared informer instance from the factory.
+	informer := s.informerFactory.ForResource(resource).Informer()
```
ryanzhang-oss: we did a transform on the informer cache in the other function but not here. Is it enough to do it once?

pkg/utils/informer/informermanager.go
weng271190436: I originally didn't have a lock but copilot suggested me to lock it defensively. But I can remove it.

Quote:
AddEventHandlerToInformer should acquire the resourcesLock when creating an informer, since it modifies shared state (the informerFactory) and could race with CreateInformerForResource. While the informerFactory itself may be thread-safe for ForResource calls, without proper synchronization there's a potential race condition where both methods could be accessing/creating informers concurrently for the same resource.

weng271190436: removed

pkg/utils/informer/informermanager.go
weng271190436: I am handling the 2nd return value, which is an error, and document what is the first return value

pkg/utils/informer/informermanager.go
weng271190436: I am doing transform for both now because "AddEventHandlerToInformer" could happen before "CreateInformerForResource" if the leader election happens quickly

ryanzhang-oss (approved):

PR #382, opened 2025-12-17
fix: do not emit unexpectedError when getting objs with dynamic client

jwtty:
DynamicClient is a simply REST client without underlying cache: https://github.com/kubernetes/client-go/blob/v0.34.3/dynamic/simple.go#L230C1-L261C2. When reading obj with a dynamic client, it can report errors like apiserver not available. These are errors that should not be classified as unexpected.

Fixes #

I have:


michaelawyu:
LGTM ;) Thanks, Wantong 🙏

PR #383, opened 2025-12-17
fix: add CEL to block user from changing External rollout strategy type to others

britaniar:

I have: 
- Made placement CRD changes to block changing from external to other types of strategys..
- Add API validation test

- API Validation Integration test

This is related to resource snapshot changes where an external strategy type CRP will not create a resourceSnapshot. The update run controller makes the snapshot vs the placement controller. Otherwise, race conditions could occur when the user updates the spec and two controllers handle the resourceSnapshots concurrently.


approval.json
```diff
@@ -0,0 +1,12 @@
+"status": {
```
ryanzhang-oss: what is this for?

britaniar: Oh, that was me testing update run... let me remove.

ryanzhang-oss (approved):

PR #386, opened 2025-12-18
fix: check the caData field too

ryanzhang-oss:
Fix the issue that clusterProfile's CAData is not populated.  No test can be easily added as the member agent use incluster usually which does not have CAData populated. This only happens if somehow a kubeConfig env is set for a controller.

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterinventory/clusterprofile/controller.go
+++ b/pkg/controllers/clusterinventory/clusterprofile/controller.go
@@ -201,7 +201,6 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
 	} else {
 		// throw an alert
 		_ = controller.NewUnexpectedBehaviorError(fmt.Errorf("cluster certificate authority data not found in member cluster %s status", mc.Name))
-		cp.Status.AccessProviders[0].Cluster.InsecureSkipTLSVerify = true
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -201,7 +201,6 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
 	} else {
 		// throw an alert
 		_ = controller.NewUnexpectedBehaviorError(fmt.Errorf("cluster certificate authority data not found in member cluster %s status", mc.Name))
-		cp.Status.AccessProviders[0].Cluster.InsecureSkipTLSVerify = true
```
jwtty, line 204: Why do we want to remove this line?

ryanzhang-oss: oh, actually there was a comment from @michaelawyu  on the previous PR.  It's better to play safe and not assume that we should use insecure mode since this is unexpected.

pkg/propertyprovider/azure/provider.go
```diff
@@ -261,9 +261,13 @@ func (p *PropertyProvider) Start(ctx context.Context, config *rest.Config) error
 		}
 		p.clusterCertificateAuthority = cadata
 		p.clusterCertificateAuthorityObservedTime = time.Now()
+		klog.V(2).Info("Cached cluster certificate authority data from file")
```
zhiying-lin, line 264: did we remove the unexpected error when "cluster certificate authority data not found" ?

ryanzhang-oss: no, we didn't. That's still unexpected. right?

zhiying-lin: nvm, it's unexpected behavior in the upstream, but it could happen in azure, so we updated our alert.

ryanzhang-oss: it's unexpected in any k8s cluster as we assume it's a TLS connection. Did I miss something?

PR #389, opened 2025-12-20
fix: add state and remove generation labels from update run metrics

britaniar:

Fixes #

I have:
- Update metrics to remove generation from update run metrics to help with cardinality.
- Add state label to help with debugging.
- Update corresponding tests.

- Integration test

Having generation as a label could potentially cause cardinality concerns when we allow the spec to change to patch the state. When the state is changed, the generation also changes. We do not limit how many times a user changes the state, there could be multiple metrics emitted instead of a core few if we keep generation. At the moment, our alerts/dashboards do use generation as a deterministic field. The latest metrics can be identified by the value (timestamp) of the metric.

Metrics that can be emitted:
```bash
// When `state` is `Initialize`, the potential metrics could be:
fleet_workload_update_run_status_last_timestamp_seconds{condition="Initialized",name="example-run",reason="UpdateRunInitializedFailed",status="False"} 1.7452803474895709e+09

fleet_workload_update_run_status_last_timestamp_seconds{condition="Initialized",name="example-run",reason="UpdateRunInitialized",status="True"} 1.7452803474895709e+09

// Once the `state` is `Run`, the potential metrics could be:
fleet_workload_update_run_status_last_timestamp_seconds{condition="Initialized",name="example-run",reason="UpdateRunInitializedFailed",status="False"} 1.7452803474895709e+09 // (if `state` starts as `Run`)

fleet_workload_update_run_status_last_timestamp_seconds{condition="Initialized",name="example-run",reason="UpdateRunInitialized",status="True"} 1.7452803474895709e+09 // (if `state` starts as `Run`)

fleet_workload_update_run_status_last_timestamp_seconds{condition="Progressing",name="example-run",reason="UpdateRunProgressing",status="True"} 1.7452803474895709e+09

fleet_workload_update_run_status_last_timestamp_seconds{condition="Progressing",name="example-run",reason="UpdateRunWaiting",status="False"} 1.7452803474895709e+09

fleet_workload_update_run_status_last_timestamp_seconds{condition="Progressing",name="example-run",reason="UpdateRunStuck",status="False"} 1.7452803474895709e+09

fleet_workload_update_run_status_last_timestamp_seconds{condition="Succeeded",name="example-run",reason="UpdateRunFailed",status="False"} 1.7452803474895709e+09

fleet_workload_update_run_status_last_timestamp_seconds{condition="Succeeded",,name="example-run",reason="UpdateRunSucceeded",status="True"} 1.7452803474895709e+09

// When `state` is `Stop`, the potential metrics could be:
fleet_workload_update_run_status_last_timestamp_seconds{condition="Progressing",name="example-run",reason="UpdateRunStopping",status="False"} 1.7452803474895709e+09

fleet_workload_update_run_status_last_timestamp_seconds{condition="Progressing",name="example-run",reason="UpdateRunStopped",status="False"} 1.7452803474895709e+09

fleet_workload_update_run_status_last_timestamp_seconds{condition="Progressing",name="example-run",reason="UpdateRunStuck",status="False"} 1.7452803474895709e+09

fleet_workload_update_run_status_last_timestamp_seconds{condition="Succeeded",name="example-run",reason="UpdateRunFailed",status="False"} 1.7452803474895709e+09 
```

### DIFF
--- a/pkg/controllers/updaterun/controller_integration_test.go
+++ b/pkg/controllers/updaterun/controller_integration_test.go
@@ -258,6 +258,10 @@ func validateUpdateRunMetricsEmitted(wantMetrics ...*prometheusclientmodel.Metri
 	}, timeout, interval).Should(Succeed(), "failed to validate the update run status metrics")
 }
 
+// generateMetricsLabels generates the labels for the update run status metrics.
+// We pass the state explicitly instead of using updateRun.Spec.State because the metric
+// should reflect the state at the time the condition occurred, which may be different from
+// the current updateRun state if the updateRun has transitioned since then.
 func generateMetricsLabels(
 	updateRun *placementv1beta1.ClusterStagedUpdateRun,
 	state, condition, status, reason string,


pkg/controllers/updaterun/controller.go
```diff
@@ -489,25 +489,26 @@ func handleApprovalRequestDelete(obj client.Object, q workqueue.TypedRateLimitin
 func emitUpdateRunStatusMetric(updateRun placementv1beta1.UpdateRunObj) {
 	generation := updateRun.GetGeneration()
 	genStr := strconv.FormatInt(generation, 10)
+	state := string(updateRun.GetUpdateRunSpec().State)
```
ryanzhang-oss: is the status state in the metrics too? like when the spec state is stop, is the status state stopped or stopping?

We have 3 mutually exclusive metrics, I am not entirely sure if that covers all the cases.

britaniar: Status is in this metric as a label. 
The reason is also a label, which could be stopping or stopped and the status would be False when state is Stop

britaniar: This would help correlate to the generation change. Give us mor insight when investigating.

pkg/controllers/updaterun/controller.go
```diff
@@ -489,25 +489,26 @@ func handleApprovalRequestDelete(obj client.Object, q workqueue.TypedRateLimitin
...
 
 	updateRunStatus := updateRun.GetUpdateRunStatus()
 	succeedCond := meta.FindStatusCondition(updateRunStatus.Conditions, string(placementv1beta1.StagedUpdateRunConditionSucceeded))
 	if succeedCond != nil && succeedCond.ObservedGeneration == generation {
-		hubmetrics.FleetUpdateRunStatusLastTimestampSeconds.WithLabelValues(updateRun.GetNamespace(), updateRun.GetName(), genStr,
+		hubmetrics.FleetUpdateRunStatusLastTimestampSeconds.WithLabelValues(updateRun.GetNamespace(), updateRun.GetName(), state, genStr,
```
Arvindthiru: Is cardinality of the metric a concern since, each time state changes generation changes, and do we also need to emit observed generation as it's no longer guaranteed to remain the same for status ?

britaniar: I don't think cardinality should be a concern. CRP uses generation as well. We only emit if observed generation and generation are the same.

Arvindthiru: Could you calculate the maxCardinality for this metric for a given updateRun ?. When the status condition changes do we overwrite existing metric for an update run or log a new line ?

britaniar: Currently we will log a new line every time the state changes because the generation will change. If you take a look at the current integration you can see that. Max cardinality is:
update run * 1 status (false) * N generation (number of times the state changes) * [ 2 reasons (reasons for succeeded condition - Failed or aborted, stuck)  + 5 reasons (reasons for progressing condition - failed or aborted, stuck, waiting, stopping, stopped)]
1 * 1 * N * [ 2 + 5] -> 7*N

We would have to remove the generation from the metric if we want to overwrite the existing metric

Arvindthiru:
Should we still emit state, since it's bound to 3 values ?

pkg/controllers/updaterun/controller_integration_test.go
```diff
@@ -260,12 +260,12 @@ func validateUpdateRunMetricsEmitted(wantMetrics ...*prometheusclientmodel.Metri
 
 func generateMetricsLabels(
 	updateRun *placementv1beta1.ClusterStagedUpdateRun,
-	condition, status, reason string,
+	state, condition, status, reason string,
```
Arvindthiru, line 267: Why do we need to pass state here, can we get it from the spec instead ?

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -874,14 +875,13 @@ var _ = Describe("Updaterun initialization tests", func() {
...
 			By("Validating the clusterStagedUpdateRun initialized consistently")
-			validateClusterStagedUpdateRunStatusConsistently(ctx, updateRun, want, "")
+			validateClusterStagedUpdateRunStatusConsistently(ctx, updateRun, initialized, "")
 
 			By("Checking update run status metrics are emitted")
-			validateUpdateRunMetricsEmitted(generateWaitingMetric(updateRun))
+			validateUpdateRunMetricsEmitted(generateInitializationSucceededMetric(updateRun))
 		})
 
 		It("Should put related ClusterResourceOverrides in the status", func() {
```
Arvindthiru, line 888: This IT description doesn't make sense, it looks lke the only diff between this one and the one above is that we specify ResourceSnapshotIndex ?

pkg/controllers/updaterun/controller_integration_test.go
britaniar: No, some of the metrics are like stopping and stopped can still exist (while update run is in state `Run`) but would have the state as `Stop` as it is only emitted then.

pkg/controllers/updaterun/initialization_integration_test.go
britaniar: Added a check to make sure the resource snapshot index is correct.

pkg/controllers/updaterun/stop_integration_test.go
```diff
@@ -270,7 +265,7 @@ var _ = Describe("UpdateRun stop tests", func() {
 			Expect(updateRun.Status.StagesStatus[0].StartTime).ShouldNot(BeNil())
 
 			By("Checking update run status metrics are emitted")
-			validateUpdateRunMetricsEmitted(wantMetrics...)
+			validateUpdateRunMetricsEmitted(generateWaitingMetric(updateRun), generateStoppedMetric(updateRun), generateProgressingMetric(updateRun))
```
Arvindthiru: Can we desired state of the updateRun as an arguement for the metrirc from each generate func ?

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -874,14 +875,14 @@ var _ = Describe("Updaterun initialization tests", func() {
 
 			By("Validating the clusterStagedUpdateRun stats")
 			initialized := generateSucceededInitializationStatus(crp, updateRun, testResourceSnapshotIndex, policySnapshot, updateStrategy, clusterResourceOverride)
-			want := generateExecutionNotStartedStatus(updateRun, initialized)
```
Arvindthiru, line 877: How was this passing till now ?

PR #390, opened 2025-12-22
fix: use the correct client in work applier ITs and set a longer timeout

michaelawyu:

This PR addresses the following issues:

a) a utility used in the work applier ITs is set to use a client for a specific env statically;
b) the work applier ITs might time out given the current setting (20 min).

Will submit subsequent PRs to run ITs via `ginkgo` CLI instead of `go test` and enable parallelization properly.

I have:

N/A

N/A

### DIFF
--- a/pkg/controllers/workapplier/waves_integration_test.go
+++ b/pkg/controllers/workapplier/waves_integration_test.go
@@ -56,7 +56,7 @@ import (
 // to (relatively speaking) reliably verify the wave-based parallel processing
 // in the work applier.
 
-var _ = FDescribe("parallel processing with waves", func() {
+var _ = Describe("parallel processing with waves", func() {
 	Context("single wave", Ordered, func() {
 		workName := fmt.Sprintf(workNameTemplate, utils.RandStr())
 		// The environment prepared by the envtest package does not support namespace


Makefile
```diff
@@ -145,7 +145,7 @@ test: manifests generate fmt vet local-unit-test integration-test## Run tests.
 local-unit-test: $(ENVTEST) ## Run tests.
 	export CGO_ENABLED=1 && \
 	export KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" && \
-	go test `go list ./pkg/... ./cmd/...` -race -coverpkg=./...  -coverprofile=ut-coverage.xml -covermode=atomic -v -timeout=20m
+	go test `go list ./pkg/... ./cmd/...` -race -coverpkg=./...  -coverprofile=ut-coverage.xml -covermode=atomic -v -timeout=30m
```
zhiying-lin, line 148: > current work applier UTs/ITs will time out

which work applier UTs/ITs are you talking about? has it been merged into the master branch?

michaelawyu: Actually there was another issue about the situation; let me address that first.

michaelawyu: A bit background on this:

Earlier we have a few utilities that are set to use static clients -> there wasn't any error before as the impacted check will always pass **immediately** due to the fact that using the wrong client means we always get object not found/deleted successfully results;

a PR merged earlier fixed some of the erred utilities, but has a leftover, which is partly the reason why impacted checks begin to fail now.

during the checkup it also appears that sometimes the work applier ITs can run for more than 1200 secs, which already exceeds the current threshold (20 min * 60 secs = 1200 secs), hence the timeout value bump.

michaelawyu: It should get fixed now by this PR -> will wait for the results and verify. Sorry for the back and forth and the confusion 😞

zhiying-lin: do you mean it takes 1200s for work applier IT to finish?

zhiying-lin: i checked the master IT failure. https://github.com/kubefleet-dev/kubefleet/actions/runs/20425296409/job/58684353347

it seems not related to the total timeout of the tests

>  [FAILED] Timed out after 10.000s.
>   Failed to remove the Work object
>   Expected success, but got an error:
>       <*errors.errorString | 0xc000b10670>: 
>       work object is being deleted, but the finalizer is still present
>       {
>           s: "work object is being deleted, but the finalizer is still present",
>       }
>   In [AfterAll] at: /home/runner/work/kubefleet/kubefleet/pkg/controllers/workapplier/waves_integration_test.go:204 @ 12/22/25 07:52:55.663

michaelawyu: See the data for the current run ([link](https://github.com/kubefleet-dev/kubefleet/actions/runs/20426641569/job/58688239852?pr=390)):

```
2025-12-22T09:13:55.7917809Z coverage: 9.8% of statements in ./...
2025-12-22T09:13:55.7918254Z ok  	github.com/kubefleet-dev/kubefleet/pkg/controllers/workapplier
1193.146s
coverage: 9.8% of statements in ./...
```

For that one specific run it took 1193.146s for othe UTs and ITs to complete in the work applier, which is on the brink of the original limit (1200s). If there's any fluctuation, the test will time out.

The work applier ITs take time to complete as a) all of our ITs are executed via `go test`, and in this mode Ginkgo will disable parallelization, and b) we have tests for exponential backoff and waved execution, which needs to wait for a long while to verify the outcome

michaelawyu: And like I mentioned, 

```
a PR merged earlier fixed some of the erred utilities, but has a leftover, which is partly the reason why impacted checks begin to fail now.
```

PR #392, opened 2025-12-23
fix: use max. 5 second work applier requeue delay for the E2E environment

michaelawyu:

This PR sets the work applier in the E2E environment to requeue at max. every 5 seconds. This is to accommodate some of the timeout settings (e.g., 10 seconds `Eventually` duration) in our E2E specs, which was designed to align with the original, non-exponential-backoff setup.

I have:

N/A

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,7 @@ test: manifests generate fmt vet local-unit-test integration-test## Run tests.
 local-unit-test: $(ENVTEST) ## Run tests.
 	export CGO_ENABLED=1 && \
 	export KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" && \
-	go test `go list ./pkg/... ./cmd/...` -race -coverpkg=./...  -coverprofile=ut-coverage.xml -covermode=atomic -v -timeout=20m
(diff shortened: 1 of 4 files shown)


charts/member-agent/templates/deployment.yaml
```diff
@@ -36,6 +36,30 @@ spec:
             - --enable-pprof={{ .Values.enablePprof }}
             - --pprof-port={{ .Values.pprofPort }}
             - --hub-pprof-port={{ .Values.hubPprofPort }}
+            {{- if .Values.workApplierRequeueRateLimiterAttemptsWithFixedDelay }}
+            - --work-applier-requeue-rate-limiter-attempts-with-fixed-delay={{ .Values.workApplierRequeueRateLimiterAttemptsWithFixedDelay }}
```
jwtty, line 40: We don't have a value set. Do we want to leave it empty and apply the default value in this case?

michaelawyu: Hi jwtty! I didn't set the default values and used the skip-if-not-set-explicitly pattern here as we already have the defaults in the entrypoint and I assume that most users will just use the default settings. I don't really have a preference on this though, do you feel that it would be better if we add the values explicitly in `Values.yaml`?

PR #393, opened 2025-12-23
fix: return unexpected error if no. of updating cluster > maxConcurrency

Arvindthiru:

Fixes #

I have:


pkg/controllers/updaterun/validation.go
```diff
@@ -234,7 +239,22 @@ func validateClusterUpdatingStatus(
...
 		updatingStageIndex = curStage
-		// TODO(arvindth): add validation to ensure updating cluster count should not exceed maxConcurrency.
+		// Collect the updating clusters.
+		var updatingClusters []string
+		for j := range stageStatus.Clusters {
+			clusterStartedCond := meta.FindStatusCondition(stageStatus.Clusters[j].Conditions, string(placementv1beta1.ClusterUpdatingConditionStarted))
+			clusterFinishedCond := meta.FindStatusCondition(stageStatus.Clusters[j].Conditions, string(placementv1beta1.ClusterUpdatingConditionSucceeded))
+			// cluster is updating if it has started but not yet finished, we also consider failed clusters as updating clusters in execution.
+			if condition.IsConditionStatusTrue(clusterStartedCond, updateRun.GetGeneration()) &&
+				!(condition.IsConditionStatusTrue(clusterFinishedCond, updateRun.GetGeneration()) || condition.IsConditionStatusFalse(clusterFinishedCond, updateRun.GetGeneration())) {
```
britaniar: Why are we checking for !true and or false? Shouldn’t the !true cover it? 

If a cluster did not successfully update do we try again?

pkg/controllers/updaterun/validation.go
```diff
@@ -164,9 +165,12 @@ func validateUpdateStagesStatus(existingStageStatus []placementv1beta1.StageUpda
 				return -1, -1, fmt.Errorf("%w: %s", errStagedUpdatedAborted, mismatchErr.Error())
 			}
 		}
-
-		var err error
-		updatingStageIndex, lastFinishedStageIndex, err = validateClusterUpdatingStatus(curStage, updatingStageIndex, lastFinishedStageIndex, &existingStageStatus[curStage], updateRun)
+		// Calculate maxConcurrency for the current stage.
+		maxConcurrency, err := calculateMaxConcurrencyValue(updateRunStatus, curStage)
+		if err != nil {
+			return -1, -1, err
```
jwtty: Do we not abort here?

pkg/controllers/updaterun/validation.go
```diff
@@ -234,7 +239,21 @@ func validateClusterUpdatingStatus(
...
 		updatingStageIndex = curStage
-		// TODO(arvindth): add validation to ensure updating cluster count should not exceed maxConcurrency.
+		// Collect the updating clusters.
+		var updatingClusters []string
+		for j := range stageStatus.Clusters {
+			clusterStartedCond := meta.FindStatusCondition(stageStatus.Clusters[j].Conditions, string(placementv1beta1.ClusterUpdatingConditionStarted))
+			clusterFinishedCond := meta.FindStatusCondition(stageStatus.Clusters[j].Conditions, string(placementv1beta1.ClusterUpdatingConditionSucceeded))
+			// cluster is updating if it has started but not yet finished, we also consider failed clusters as updating clusters in execution.
+			if condition.IsConditionStatusTrue(clusterStartedCond, updateRun.GetGeneration()) && !(condition.IsConditionStatusTrue(clusterFinishedCond, updateRun.GetGeneration())) {
+				updatingClusters = append(updatingClusters, stageStatus.Clusters[j].ClusterName)
```
jwtty: Do we need an array? Or just an interger count is enough?

jwtty: Or if you want to print all the updating cluster names.

pkg/controllers/updaterun/validation_test.go
```diff
@@ -159,14 +160,67 @@ func TestValidateClusterUpdatingStatus(t *testing.T) {
...
 		},
+		{
+			name:                   "validateClusterUpdatingStatus should return error if number of updating clusters exceeds maxConcurrency",
+			curStage:               0,
+			updatingStageIndex:     -1,
+			lastFinishedStageIndex: -1,
+			stageStatus: &placementv1beta1.StageUpdatingStatus{
+				StageName:  "test-stage",
+				Conditions: []metav1.Condition{generateTrueCondition(updateRun, placementv1beta1.StageUpdatingConditionProgressing)},
+				Clusters: []placementv1beta1.ClusterUpdatingStatus{
```
jwtty, line 226: Add a case without Started condition at all?

pkg/controllers/updaterun/validation.go
```diff
@@ -234,7 +239,21 @@ func validateClusterUpdatingStatus(
 			return -1, -1, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
 		}
 		updatingStageIndex = curStage
-		// TODO(arvindth): add validation to ensure updating cluster count should not exceed maxConcurrency.
+		// Collect the updating clusters.
+		var updatingClusters []string
+		for j := range stageStatus.Clusters {
+			clusterStartedCond := meta.FindStatusCondition(stageStatus.Clusters[j].Conditions, string(placementv1beta1.ClusterUpdatingConditionStarted))
+			clusterFinishedCond := meta.FindStatusCondition(stageStatus.Clusters[j].Conditions, string(placementv1beta1.ClusterUpdatingConditionSucceeded))
```
ryanzhang-oss, line 246: how fresh is the stageStatus? can it be stale that the previous reconcile mark it as finished in-memory but failed to upgdate it before starting a new cluster resource update?

pkg/controllers/updaterun/validation.go
britaniar: ^ We need something like `return -1, -1, fmt.Errorf("%w: %s", errStagedUpdatedAborted, err.Error())`, or update within the function `calculateMaxConcurrencyValue`

pkg/controllers/updaterun/validation.go
Arvindthiru: it can be stale, but the unexpected error still holds true

On every execute call, we only start updates to clusters with limit set by maxConcurrency. And only on a successful reconcile loop do we update stage status on updateRun.

Even if we start a subsequent reconcile mark some clusters as finished and start updates to a new set of clusters (still bound by maxConcurrency) unless we update stage status successfully they are not persisted in the status and won't count towards this check

jwtty: It can be stale, but the stale updating count (previous state) should not exceed the maxConcurrency either.

pkg/controllers/updaterun/validation_test.go
```diff
@@ -144,6 +145,28 @@ func TestValidateClusterUpdatingStatus(t *testing.T) {
...
+					{
+						ClusterName: "cluster-1",
+					},
+					{
+						ClusterName: "cluster-2",
+					},
+				},
+			},
+			maxConcurrency:             2,
```
jwtty: There are 2 clusters and the maxConcurrency is set to 2. It will never fail the check. I think it's better to either add a cluster or set maxConcurrency to 1 to make sure these clusters are indeed not counted. Same as other tests.

PR #395, opened 2026-01-06
feat: add ns controller to collect ns for member clusters

zhiying-lin:

- use the property provider to collect ns periodically
- store the ns information in the status of imc, instead of labels (the api is cleaner compared with storing in labels)
- The pros of putting the ns info as the label in the mc is it's easier to deprecate and no api changes once we support placement affinity.
- if the ns is applied by the CRP, store the work name instead as we cannot get the CRP name directly from the resources.

Fixes #

I have:

ut + it

We probably need to create a default property provider, including these common property collectors.

### DIFF
--- a/pkg/propertyprovider/default/controllers/namespace.go
+++ b/pkg/propertyprovider/default/controllers/namespace.go
@@ -96,11 +96,17 @@ func (r *NamespaceReconciler) SetupWithManager(mgr ctrl.Manager, controllerName 
 			}
 
 			// Watch for namespace transitioning to terminating state
+			// because we want to stop tracking the namespace once it
+			// is marked for deletion, to prevent new workloads from being
+			// scheduled to the namespace.
 			if oldNs.DeletionTimestamp == nil && newNs.DeletionTimestamp != nil {
 				return true
 			}
 
-			// Watch for owner reference changes
+			// Watch for owner reference changes because the owner applied work name
+			// is part of the namespace information we track for namespace affinity.
+			// We want to keep the namespace information up to date in the tracker
+			// if the owner reference changes.
 			if !reflect.DeepEqual(oldNs.OwnerReferences, newNs.OwnerReferences) {
 				return true
 			}
--- a/apis/cluster/v1beta1/internalmembercluster_types.go
+++ b/apis/cluster/v1beta1/internalmembercluster_types.go
@@ -78,6 +78,7 @@ type InternalMemberClusterStatus struct {
 	// Namespaces is a map of namespace names to their associated work names for namespaces
 	// that are managed by Fleet (i.e., have AppliedWork owner references when created).
 	// The key is the namespace name and the value is the work name from the AppliedWork owner reference.
+	// If the namespace does not have an AppliedWork owner reference, the value will be an empty string.
 	// This field is populated by the property provider when namespace collection is enabled.
 	// +optional
 	Namespaces map[string]string `json:"namespaces,omitempty"`
--- a/config/crd/bases/cluster.kubernetes-fleet.io_internalmemberclusters.yaml
+++ b/config/crd/bases/cluster.kubernetes-fleet.io_internalmemberclusters.yaml
@@ -493,6 +493,7 @@ spec:
                   Namespaces is a map of namespace names to their associated work names for namespaces
                   that are managed by Fleet (i.e., have AppliedWork owner references when created).
                   The key is the namespace name and the value is the work name from the AppliedWork owner reference.
+                  If the namespace does not have an AppliedWork owner reference, the value will be an empty string.
                   This field is populated by the property provider when namespace collection is enabled.
                 type: object
               properties:


weng271190436 (dismissed):

weng271190436 (approved):

pkg/propertyprovider/default/controllers/namespace.go
```diff
@@ -0,0 +1,82 @@
...
+	klog.V(2).InfoS("Attempt to track the namespace", "namespace", namespaceRef)
+	r.NamespaceTracker.AddOrUpdate(namespace)
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *NamespaceReconciler) SetupWithManager(mgr ctrl.Manager, controllerName string) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(controllerName).
+		For(&corev1.Namespace{}).
```
ryanzhang-oss, line 126: do we care if there are just updates on the namespace?

pkg/propertyprovider/default/trackers/namespaces.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+}
+
+// AddOrUpdate starts tracking a namespace or updates the information about a namespace that has been
+// tracked.
+func (nt *NamespaceTracker) AddOrUpdate(namespace *corev1.Namespace) {
+	nt.mu.Lock()
+	defer nt.mu.Unlock()
+
+	nsKObj := klog.KObj(namespace)
```
ryanzhang-oss, line 73: why do we need to track a namespace that is not placed by the fleet?

pkg/propertyprovider/default/trackers/namespaces.go
```diff
@@ -0,0 +1,135 @@
...
+	result := make(map[string]string, len(nt.namespacesByName))
+	for name, namespace := range nt.namespacesByName {
+		workName := nt.extractWorkNameFromNamespace(namespace)
+		result[name] = workName
+	}
+	return result, nt.reachLimit
+}
+
+// extractWorkNameFromNamespace extracts the work name from a namespace by checking its owner references.
+func (nt *NamespaceTracker) extractWorkNameFromNamespace(namespace *corev1.Namespace) string {
```
ryanzhang-oss, line 127: what happens if someone removes the ownerRef?

pkg/propertyprovider/default/controllers/namespace.go
weng271190436: only watching create and delete now

pkg/propertyprovider/default/trackers/namespaces.go
weng271190436: not tracking those now. The check happens right before AddOrUpdate

pkg/propertyprovider/default/trackers/namespaces.go
weng271190436: To answer this question, we need to decide on which namespaces are in scope for namespace affinity:
1. namespaces managed by fleet
- If we consider only namespaces managed by fleet, once the onwer reference is removed, the namespace is no longer managed by fleet. It is okay not to consider the namespace for namespace affinity. I should also listen to update events where owner reference is changed if this is the case (not doing this yet because of another comment "do we care if there are just updates on the namespace?").

2. all namespaces
- If we consider all namespaces, whether the owner ref was removed is not important.

3. namespace managed by fleet at some point (created by fleet or taken over by fleet):
- It is difficult to determine whether a namespace was managed by fleet at some point once it is no longer managed by fleet.

pkg/propertyprovider/default/controllers/namespace.go
```diff
@@ -0,0 +1,107 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	// Track the namespace. If it has been tracked, update its information with the tracker.
+	klog.V(2).InfoS("Attempt to track the namespace", "namespace", namespaceRef)
+	r.NamespaceTracker.AddOrUpdate(namespace)
+	return ctrl.Result{}, nil
+}
+
+// hasAppliedWorkOwnerRef checks if a namespace has an AppliedWork owner reference.
```
ryanzhang-oss: is there any benefit to create this one line function that is only called in one place?

pkg/propertyprovider/default/controllers/namespace.go
```diff
@@ -0,0 +1,107 @@
...
+	startTime := time.Now()
+	klog.V(2).InfoS("Reconciliation starts for namespace objects in the property provider", "namespace", namespaceRef)
+	defer func() {
+		latency := time.Since(startTime).Milliseconds()
+		klog.V(2).InfoS("Reconciliation ends for namespace objects in the property provider", "namespace", namespaceRef, "latency", latency)
+	}()
+
+	namespace := &corev1.Namespace{}
+	if err := r.Client.Get(ctx, req.NamespacedName, namespace); err != nil {
+		if errors.IsNotFound(err) {
```
ryanzhang-oss, line 55: we don't watch upate event. For NS that has a finalizer, I am not sure if we will get a "delete" event. Maybe eventually but it might stuck for a while if the finalizer is not removed immedidately.

pkg/propertyprovider/default/trackers/namespaces.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		// New namespace being tracked
+		nt.namespaceCreationTimes[namespace.Name] = time.Now()
+		klog.V(2).InfoS("Started tracking namespace", "namespace", nsKObj)
+	} else {
+		klog.V(4).InfoS("Updated namespace information", "namespace", nsKObj)
+	}
+
+	// Store a copy of the namespace to avoid potential modifications
+	nt.namespacesByName[namespace.Name] = namespace.DeepCopy()
```
ryanzhang-oss: why not store the appliedWork here instead? The collect is called every few seconds and work name is smaller. We use more memory and compute by storing the entire ns here.

pkg/propertyprovider/default/controllers/namespace.go
weng271190436: yes eventually we get delete event

to handle terminating namespace we need to process some update events

pkg/propertyprovider/default/trackers/namespaces.go
weng271190436: after discussion, we are considering all namespaces so someone removing the owner reference is no longer an issue

pkg/propertyprovider/default/controllers/namespace.go
```diff
@@ -0,0 +1,129 @@
...
+			// scheduled to the namespace.
+			if oldNs.DeletionTimestamp == nil && newNs.DeletionTimestamp != nil {
+				return true
+			}
+
+			// Watch for owner reference changes because the owner applied work name
+			// is part of the namespace information we track for namespace affinity.
+			// We want to keep the namespace information up to date in the tracker
+			// if the owner reference changes.
+			if !reflect.DeepEqual(oldNs.OwnerReferences, newNs.OwnerReferences) {
```
weng271190436, line 110: One can argue that the applied work name is not even necessary because we are tracking all namespaces regardless of whether it is managed by fleet or not. Only keeping it here to give additional information on the origin of the namespace.

pkg/propertyprovider/default/controllers/namespace.go
weng271190436: I am listening to UPDATE events where namespace becomes terminating.

PR #398, opened 2026-01-07
fix: add cluster inventory group to CRD validation webhook

jwtty:
This PR adds clusterInventory CRD group to the validation webhook so that only master/admin user group can create/update/delete these CRDs, like "clusterProfile".

Fixes #

I have:

### DIFF
--- a/pkg/webhook/validation/uservalidation_test.go
+++ b/pkg/webhook/validation/uservalidation_test.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
 	admissionv1 "k8s.io/api/admission/v1"
 	authenticationv1 "k8s.io/api/authentication/v1"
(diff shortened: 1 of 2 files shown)


pkg/webhook/validation/uservalidation_test.go
```diff
@@ -437,3 +437,175 @@ func TestValidateFleetMemberClusterUpdate(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateUserForFleetCRD(t *testing.T) {
```
Arvindthiru, line 442: If we are adding more tests, can we add them here instead https://github.com/kubefleet-dev/kubefleet/blob/main/pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook_test.go#L35 ?

zhiying-lin: prefer to add new test cases for ValidateUserForFleetCRD func.

in general, any public func should have ut covered.

we need to make sure it should not break existing TestHandleCRD tests.

pkg/webhook/validation/uservalidation_test.go
```diff
@@ -437,3 +437,175 @@ func TestValidateFleetMemberClusterUpdate(t *testing.T) {
...
+			wantResponse: admission.Allowed(fmt.Sprintf(ResourceAllowedFormat, "regular-user", utils.GenerateGroupString([]string{"regular-group"}), admissionv1.Create,
+				&metav1.GroupVersionKind{Group: "apiextensions.k8s.io", Version: "v1", Kind: "CustomResourceDefinition"}, "",
+				types.NamespacedName{Name: "gateways.gateway.networking.k8s.io"})),
+		},
+	}
+
+	for testName, testCase := range testCases {
+		t.Run(testName, func(t *testing.T) {
+			gotResult := ValidateUserForFleetCRD(testCase.req, testCase.whiteListedUsers, testCase.group)
+			assert.Equal(t, testCase.wantResponse, gotResult, utils.TestCaseMsg, testName)
```
zhiying-lin: use the golang built-in package instead of assert.Equal

PR #408, opened 2026-01-08
fix: error messages used for updateRun

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/updaterun/execution.go
+++ b/pkg/controllers/updaterun/execution.go
@@ -690,11 +690,15 @@ func generateStuckClustersString(stuckClusterNames []string) string {
 	if len(stuckClusterNames) == 0 {
 		return ""
 	}
-	if len(stuckClusterNames) == 1 {
-		return stuckClusterNames[0]
+	if len(stuckClusterNames) <= 3 {
+		return strings.Join(stuckClusterNames, ", ")
 	}
-	// Multiple clusters: join all with commas and append ", ..."
-	return strings.Join(stuckClusterNames, ", ") + "..."
+	// Print first 3 if more than 3 stuck clusters.
+	clustersToShow := stuckClusterNames
+	if len(stuckClusterNames) > 3 {
+		clustersToShow = stuckClusterNames[:3]
+	}
+	return strings.Join(clustersToShow, ", ") + "..."
 }
 
 // markUpdateRunWaiting marks the updateRun as waiting in memory.
--- a/pkg/controllers/updaterun/execution_test.go
+++ b/pkg/controllers/updaterun/execution_test.go
@@ -1205,3 +1205,47 @@ func TestCheckBeforeStageTasksStatus_NegativeCases(t *testing.T) {
 		})
 	}
 }
+
+func TestGenerateStuckClustersString(t *testing.T) {
+	tests := []struct {
+		name              string
+		stuckClusterNames []string
+		wantClusterString string
+	}{
+		{
+			name:              "empty cluster list",
+			stuckClusterNames: []string{},
+			wantClusterString: "",
+		},
+		{
+			name:              "single cluster",
+			stuckClusterNames: []string{"cluster1"},
+			wantClusterString: "cluster1",
+		},
+		{
+			name:              "two clusters",
+			stuckClusterNames: []string{"cluster1", "cluster2"},
+			wantClusterString: "cluster1, cluster2",
+		},
+		{
+			name:              "three clusters",
+			stuckClusterNames: []string{"cluster1", "cluster2", "cluster3"},
+			wantClusterString: "cluster1, cluster2, cluster3",
+		},
+		{
+			name:              "five clusters - should only show first three",
+			stuckClusterNames: []string{"cluster1", "cluster2", "cluster3", "cluster4", "cluster5"},
+			wantClusterString: "cluster1, cluster2, cluster3...",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := generateStuckClustersString(tt.stuckClusterNames)
+
+			if got != tt.wantClusterString {
+				t.Fatalf("generateStuckClustersString() = %v, want %v", got, tt.wantClusterString)
+			}
+		})
+	}
+}


britaniar (approved):

pkg/controllers/updaterun/controller.go
```diff
@@ -51,9 +51,10 @@ import (
 var (
 	// errStagedUpdatedAborted is the error when the updateRun is aborted.
 	errStagedUpdatedAborted = fmt.Errorf("cannot continue the updateRun")
-	// errInitializedFailed is the error when the updateRun fails to initialize.
-	// It is a wrapped error of errStagedUpdatedAborted, because some initialization functions are reused in the validation step.
-	errInitializedFailed = fmt.Errorf("%w: failed to initialize the updateRun", errStagedUpdatedAborted)
```
Arvindthiru, line 56: We use this error during initalization and validation phases, and once initalized we use the same error to consturct messages for execution related condition as well. Since we perform validation in both cases changing it here. For more context: https://github.com/kubefleet-dev/website/pull/65#discussion_r2677456343

pkg/controllers/updaterun/controller.go
```diff
@@ -126,7 +127,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		if toBeUpdatedBindings, toBeDeletedBindings, initErr = r.initialize(ctx, updateRun); initErr != nil {
 			klog.ErrorS(initErr, "Failed to initialize the updateRun", "updateRun", runObjRef)
 			// errInitializedFailed cannot be retried.
```
jwtty: comment needs to be updated

pkg/controllers/updaterun/execution.go
```diff
@@ -675,14 +674,14 @@ func markUpdateRunProgressingIfNotWaitingOrStuck(updateRun placementv1beta1.Upda
...
-func markUpdateRunStuck(updateRun placementv1beta1.UpdateRunObj, stageName, clusterNames string) {
+func markUpdateRunStuck(updateRun placementv1beta1.UpdateRunObj, stageName string, clusterCount int) {
 	updateRunStatus := updateRun.GetUpdateRunStatus()
 	meta.SetStatusCondition(&updateRunStatus.Conditions, metav1.Condition{
 		Type:               string(placementv1beta1.StagedUpdateRunConditionProgressing),
 		Status:             metav1.ConditionFalse,
 		ObservedGeneration: updateRun.GetGeneration(),
 		Reason:             condition.UpdateRunStuckReason,
-		Message:            fmt.Sprintf("The updateRun is stuck waiting for cluster(s) %s in stage %s to finish updating, please check placement status for potential errors", clusterNames, stageName),
+		Message:            fmt.Sprintf("The updateRun is stuck waiting for %d cluster(s) in stage %s to finish updating, please check placement status for potential errors", clusterCount, stageName),
```
jwtty: It might still be good to mention a few, first 3 stuck clusters for example.

pkg/controllers/updaterun/controller.go
```diff
@@ -126,7 +127,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
-			if errors.Is(initErr, errInitializedFailed) {
+			if errors.Is(initErr, errValidationFailed) {
```
jwtty: We can use `errors.Is(initErr, errStagedUpdatedAborted)` too?

pkg/controllers/updaterun/execution.go
```diff
@@ -553,7 +553,7 @@ func calculateMaxConcurrencyValue(status *placementv1beta1.UpdateRunStatus, stag
 // It marks the update run as stuck if any clusters are stuck, or as progressing if some clusters have finished updating.
 func aggregateUpdateRunStatus(updateRun placementv1beta1.UpdateRunObj, stageName string, stuckClusterNames []string) {
 	if len(stuckClusterNames) > 0 {
-		markUpdateRunStuck(updateRun, stageName, strings.Join(stuckClusterNames, ", "))
+		markUpdateRunStuck(updateRun, stageName, stuckClusterNames)
```
ryanzhang-oss, line 556: why making this change?

pkg/controllers/updaterun/controller.go
ryanzhang-oss: should this become two errors? One for init error and one for validation error?

pkg/controllers/updaterun/controller.go
```diff
@@ -125,8 +126,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		var initErr error
 		if toBeUpdatedBindings, toBeDeletedBindings, initErr = r.initialize(ctx, updateRun); initErr != nil {
 			klog.ErrorS(initErr, "Failed to initialize the updateRun", "updateRun", runObjRef)
-			// errInitializedFailed cannot be retried.
-			if errors.Is(initErr, errInitializedFailed) {
+			// errStagedUpdatedAborted cannot be retried.
```
ryanzhang-oss, line 129: although it's still correct, I wonder why change this?

pkg/controllers/updaterun/controller.go
ryanzhang-oss: is that every init error is because of validation failure?

pkg/controllers/updaterun/execution.go
Arvindthiru: Adding a new util func generateStuckClustersString to create a bound message string based on stuck cluster names. If the number of stuck clusters is greater than 3 we trim the stuck cluster names to cluster1, cluster2, cluster3... this way the message string is bounded to a fixed length instead of being unbounded

pkg/controllers/updaterun/controller.go
Arvindthiru: Keeping things consistent cause we do similar err checks in the reconcile function

pkg/controllers/updaterun/controller.go
Arvindthiru: Yes, we emit API server errors and validation errors from initialize

jwtty: This was the design from the very beginning. The functions are shared during initialization and validation. The error should be a wrapper of the errAborted so that controller can avoid retrying an updateRun in face of unrecoverable errors.

ryanzhang-oss (approved):

PR #409, opened 2026-01-10
feat: skip update condition of clusterProfile

ryanzhang-oss:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterinventory/clusterprofile/controller.go
+++ b/pkg/controllers/clusterinventory/clusterprofile/controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -128,30 +129,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	createOrUpdateRes, err := controllerutil.CreateOrUpdate(ctx, r, cp, func() error {
-		if cp.CreationTimestamp.IsZero() {
-			// Only set the ClusterManager field if the object is being created; this field
-			// is immutable by definition.
-			cp.Spec = clusterinventory.ClusterProfileSpec{
-				ClusterManager: clusterinventory.ClusterManager{
-					Name: controller.ClusterManagerName,
-				},
+		if !cp.CreationTimestamp.IsZero() {
+			// log an unexpected error if the cluster profile content is modified behind our back
```
ytimocin, line 133: For external contributors to understand, should there be a comment on what would happen if the cluster profile content is modified? What would it break down the line?

pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -308,6 +310,15 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 					NamespacedName: types.NamespacedName{Name: e.Object.GetName()},
 				})
 			},
+			UpdateFunc: func(ctx context.Context, e event.UpdateEvent, q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+				if e.ObjectNew.GetGeneration() == e.ObjectOld.GetGeneration() {
+					return
+				}
+				klog.V(2).InfoS("Handling a clusterProfil spec update event", "clusterProfile", klog.KObj(e.ObjectOld))
+				q.Add(reconcile.Request{
+					NamespacedName: types.NamespacedName{Name: e.ObjectOld.GetName()},
```
ytimocin, line 319: May not be a very valid question, but, could the name have changed for **ObjectOld** and **ObjectNew**?

pkg/controllers/clusterinventory/clusterprofile/controller.go
ryanzhang-oss: the clusterProfile object is designe to be consumed by other controllers.  So nothing will "break" here but those fields do not expect to be changed.

pkg/controllers/clusterinventory/clusterprofile/controller.go
ryanzhang-oss: no, name can't change for any k8s object.

PR #412, opened 2026-01-14
fix: make public updaterun test util functions importable

Arvindthiru:

Follow up for https://github.com/kubefleet-dev/kubefleet/pull/397. This doesn't work since the public utils are present in a *_test.go file. https://pkg.go.dev/go.goms.io/fleet@v0.17.1/test/e2e

Only exposing the complex status function as public util this time around. The other functions listed below are tied with eventuallyDuration, eventualnterval and we need to pass duration, interval as arguments so restored them back to their previous state

- updateClusterStagedUpdateRunState
- validateAndApproveClusterApprovalRequests
- updateStagedUpdateRunState
- validateAndApproveNamespacedApprovalRequests

Fixes #

I have:

### DIFF
--- a/test/utils/updaterun/updaterun_status.go
+++ b/test/utils/updaterun/updaterun_status.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2025 The KubeFleet Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


test/utils/updaterun/updaterun_status.go
```diff
@@ -0,0 +1,310 @@
+package updaterun
```
britaniar, line 17: we are missing the copyright

PR #414, opened 2026-01-15
feat: improve the default setup for hub agent leader election to allow better scalability/stability

michaelawyu:

This PR makes the following changes:

* Use a separate config (rate limiter setup) for leader election, so that lease renewals will not be starved by regular controllers under heavy load.
* Apply a default lease duration of 60s (increased from 15s), a default lease renewal duration of 45s (increases from 10s), and a default lease renewal retry period of 5s (increased from 2s). This helps the hub agent to hold onto leadership longer + send less renewal requests), and have a better chance at renewing leases (9 attempts in 45 seconds vs ~5 attempts in 10 seconds), as in recent performance tests we have identified that unexpected leader election failures can lead to frequent agent restarts under heavy load.

I have:

N/A

Please refer to the Jan 2026 performance test report for more information.

### DIFF
--- a/cmd/hubagent/main.go
+++ b/cmd/hubagent/main.go
@@ -124,8 +124,8 @@ func main() {
 		LeaderElection:       opts.LeaderElectionOpts.LeaderElect,
 		LeaderElectionConfig: leaderElectionCfg,
 		// If leader election is enabled, the hub agent by default uses a setup
-		// with a lease duration of 90 secs, a renew deadline of 75 secs, and a retry period of 5 secs.
-		// This setup gives the hub agent up to 15 attempts/75 seconds to renew its leadership lease
+		// with a lease duration of 60 secs, a renew deadline of 45 secs, and a retry period of 5 secs.
+		// This setup gives the hub agent up to 9 attempts/45 seconds to renew its leadership lease
 		// before it loses the leadership and restarts.
 		//
 		// These values are set significantly higher than the controller-runtime defaults
@@ -136,7 +136,7 @@ func main() {
 		//
 		// Note (chenyu1): a minor side effect with the higher values is that when the agent does restart,
 		// (or in the future when we do run multiple hub agent replicas), the new leader might have to wait a bit
-		// longer (up to 90 seconds) to acquire the leadership, which should still be acceptable in most scenarios.
+		// longer (up to 60 seconds) to acquire the leadership, which should still be acceptable in most scenarios.
 		LeaseDuration:           &opts.LeaderElectionOpts.LeaseDuration.Duration,
 		RenewDeadline:           &opts.LeaderElectionOpts.RenewDeadline.Duration,
 		RetryPeriod:             &opts.LeaderElectionOpts.RetryPeriod.Duration,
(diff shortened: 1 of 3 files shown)


cmd/hubagent/main.go
```diff
@@ -102,16 +103,41 @@ func main() {
...
+	// This aims to improve the availability of the hub agent; originally all access to the API
+	// server would share the same rate limiting configuration, and under adverse conditions (e.g.,
+	// large volume of concurrent placements) controllers would exhause all tokens in the rate limiter,
+	// and effectively starve the leader election process (the runtime can no longer renew leases),
+	// which would trigger the hub agent to restart even though the system remains functional.
+	defaultCfg := ctrl.GetConfigOrDie()
+	leaderElectionCfg := rest.CopyConfig(defaultCfg)
+
+	defaultCfg.QPS, defaultCfg.Burst = float32(opts.HubQPS), opts.HubBurst
+	leaderElectionCfg.QPS, leaderElectionCfg.Burst = float32(opts.LeaderElectionQPS), opts.LeaderElectionBurst
```
ytimocin: Should the outputs of float32 be validated before moving on?

ytimocin: Could an invalid config cause panic down the line?

ryanzhang-oss: +1, looks like we missed many option validation?

michaelawyu: Yeah, most of the options do not have validation. I could submit a separate PR for this (and thanks, ytimochin, for raising the issue).

As a side note, for this specific parameter, QPS < 0 -> rate limiting disabled, QPS == 0 -> default QPS value would apply, QPS > 1 -> use the configured QPS.

ryanzhang-oss: is that the go client default the invalid QPS?

michaelawyu: Resolving as the refactored CLI setup with validation has been added. I have pushed additional commits to this PR to make sure of the new setup. PTAL.

cmd/hubagent/options/leaderelection.go
```diff
@@ -68,23 +82,23 @@ func (o *LeaderElectionOptions) AddFlags(flags *flag.FlagSet) {
 	flags.DurationVar(
 		&o.LeaseDuration.Duration,
 		"leader-lease-duration",
-		15*time.Second,
+		90*time.Second,
```
ryanzhang-oss: have we tested this value? It seems a bit high

michaelawyu: Hi Ryan! In the performance test we used a value even higher (180s), but if 90 secs is a concern, would you like me to lower this to 60 secs?

The requirement is basically the lease duration must be higher than the renew deadline.

michaelawyu: Or to be on the safer side, would it be better if we rollout this in multiple stages?

e.g.,

for now, set things to 30 seconds/25 seconds/5 seconds; when it rolls out fully,
next, set things to 45 seconds/40 seconds/5 seconds; ...

It shouldn't affect our performance test plan -> I could do an override on that specific environment.

ryanzhang-oss: let's try 60/45/5?

michaelawyu: Got it. Has tweaked the numbers.

PR #419, opened 2026-01-27
feat: add GVK whitelisting support for status back-reporting

michaelawyu:

This PR adds GVK whitelisting support for status back-reporting so that the hub agent will only back-report status information to the original resource for a specific set of GVKs. This helps most in cases where the Fleet hub cluster runs controllers that would reconcile such objects and add their own status, which would lead to contention with the Fleet hub agent.

I have:

- [x] Unit tests
- [x] Integration tests

This PR is a prerequisite before we enable the status back-reporting feature in the system.

### DIFF
--- a/cmd/hubagent/options/options.go
+++ b/cmd/hubagent/options/options.go
@@ -118,6 +118,27 @@ type Options struct {
 	ResourceSnapshotCreationMinimumInterval time.Duration
 	// ResourceChangesCollectionDuration is the duration for collecting resource changes into one snapshot.
 	ResourceChangesCollectionDuration time.Duration
+	// EnableStatusBackReporting enables the status back-reporting feature, which allows users to view object
+	// status of placed objects on the member cluster side directly on the hub cluster side.
+	EnableStatusBackReporting bool
+	// StatusMirroringToOriginalResourceAllowedGVKsRawStr is the raw string input that specifies a list of GVKs that are allowed
+	// to have their object status back-reported from the member cluster side to the **original resource** on
+	// the hub cluster side. The whitelist has no effect if the status back-reporting destination is set to the Work API.
+	//
+	// The string is a collection of segments, separated by commas; each segment is of the format
+	// `API_GROUP/VERSION/KIND`, where `API_GROUP`, `VERSION`, and `KIND` are the API group, version, and
+	// kind of a Kubernetes API resource respectively, such as `apps/v1/Deployment`. For API resources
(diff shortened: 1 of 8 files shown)


cmd/hubagent/options/options.go
```diff
@@ -193,4 +214,9 @@ func (o *Options) AddFlags(flags *flag.FlagSet) {
 	flags.DurationVar(&o.ResourceChangesCollectionDuration, "resource-changes-collection-duration", 15*time.Second,
 		"The duration for collecting resource changes into one snapshot. The default is 15 seconds, which means that the controller will collect resource changes for 15 seconds before creating a resource snapshot.")
 	o.RateLimiterOpts.AddFlags(flags)
+	flags.BoolVar(&o.EnableStatusBackReporting, "enable-status-back-reporting", false, "If set, the status back-reporting feature is enabled. This allows users to view object status of placed objects on the member cluster side directly on the hub cluster side.")
```
ryanzhang-oss, line 217: if we start this as false, we need to change OM charts to enable it

pkg/utils/gvkwhitelist/gvkwhitelist.go
```diff
@@ -0,0 +1,181 @@
...
+	return &WhitelistedGVKs{
+		root: &tNode{
+			children: map[string]*tNode{},
+		},
+		shortcuts: map[schema.GroupVersionKind]bool{},
+	}
+}
+
+// Register adds a new GVK into the whitelist.
+func (w *WhitelistedGVKs) Register(apiGroup, apiVersion, kind string) {
```
ryanzhang-oss, line 60: the current PR gets the GVK list from a flag which makes it static, why do we need all the locks?

pkg/utils/gvkwhitelist/gvkwhitelist.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// TO-DO (chenyu1): evaluate if the shortcuts can really help improve performance, as there
+	// might be additional overhead associated with the guarding mutex.
+	shortcuts map[schema.GroupVersionKind]bool
+
+	mu sync.Mutex
+}
+
+// NewWhitelistedGVKs returns a new WhitelistedGVKs instance.
+func NewWhitelistedGVKs() *WhitelistedGVKs {
```
ryanzhang-oss, line 50: we already have something similar in checking if a GVK should be propagated, can we not reuse it?

PR #427, opened 2026-02-04
fix: clusterProfile controller should handle the MC condition

ryanzhang-oss:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterinventory/clusterprofile/controller_integration_test.go
+++ b/pkg/controllers/clusterinventory/clusterprofile/controller_integration_test.go
@@ -118,7 +118,7 @@ var _ = Describe("Test ClusterProfile Controller", func() {
 		}, eventuallyTimeout, interval).Should(BeTrue(), "clusterProfile is not created")
 	})
 
-	It("Should recreate a clusterProfile when it is deleted by the user", func() {
+	It("Should recreate a clusterProfile when it is deleted by the user but properties should not show if MC property collection is not succeeded", func() {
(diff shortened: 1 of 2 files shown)


pkg/controllers/clusterinventory/clusterprofile/suite_test.go
```diff
@@ -92,7 +97,7 @@ var _ = BeforeSuite(func() {
 		Metrics: server.Options{
 			BindAddress: "0",
 		},
-		Logger: textlogger.NewLogger(textlogger.NewConfig(textlogger.Verbosity(4))),
+		Logger: zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true), zap.Level(zapcore.Level(-5))),
```
michaelawyu: Hi Ryan! Should this just be the common logger (the one created in line 66?)?

pkg/controllers/clusterinventory/clusterprofile/controller_integration_test.go
```diff
@@ -101,6 +119,18 @@ var _ = Describe("Test ClusterProfile Controller", func() {
 	})
 
 	It("Should recreate a clusterProfile when it is deleted by the user", func() {
+		By("Mark the member cluster as joined")
```
michaelawyu, line 122: Hi Ryan! Just a nit: the name of this spec might need a bit change to reflect that re-creation will happen but some data won't be populated until the properties are added.

PR #429, opened 2026-02-05
feat: Create struct for Resource Snapshot Creation to use across controllers

britaniar:

Fixes #

I have:
- Created a struct for resource struct creation
- Moved related resource snapshot creation functions to pkg/utils/controller/resource_snapshot_reolver.go
- Move related UTs

- UTs
- Manually checked if it would work overall once we make the changes to the controller and agent.

Next Steps:
1. Update Placement controller and agent to use new struct
2. Refactor function used for placement (default) to ignore External and not create a resource snapshot.

### DIFF
--- a/test/apis/placement/v1beta1/api_validation_integration_test.go
+++ b/test/apis/placement/v1beta1/api_validation_integration_test.go
@@ -1409,7 +1409,7 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 			err := hubClient.Create(ctx, &strategy)
 			var statusErr *k8sErrors.StatusError
 			Expect(errors.As(err, &statusErr)).To(BeTrue(), fmt.Sprintf("Create updateRunStrategy call produced error %s. Error type wanted is %s.", reflect.TypeOf(err), reflect.TypeOf(&k8sErrors.StatusError{})))
-			Expect(statusErr.ErrStatus.Message).Should(MatchRegexp("Too long: may not be longer than 63"))
+			Expect(statusErr.ErrStatus.Message).Should(MatchRegexp("Too long: may not be more than 63 bytes"))
 		})
 
 		It("Should deny creation of ClusterStagedUpdateStrategy with invalid stage config - stage name with invalid characters", func() {


pkg/utils/controller/resource_snapshot_resolver.go
```diff
@@ -19,18 +19,453 @@ package controller
...
 	"github.com/kubefleet-dev/kubefleet/pkg/scheduler/queue"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/annotations"
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/labels"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/resource"
+	fleettime "github.com/kubefleet-dev/kubefleet/pkg/utils/time"
 )
 
+// The max size of an object in k8s is 1.5MB because of ETCD limit https://etcd.io/docs/v3.3/dev-guide/limit/.
+// We choose 800KB as the soft limit for all the selected resources within one resourceSnapshot object because of this test in k8s which checks
+// if object size is greater than 1MB https://github.com/kubernetes/kubernetes/blob/db1990f48b92d603f469c1c89e2ad36da1b74846/test/integration/master/synthetic_master_test.go#L337
```
ryanzhang-oss, line 46: FYI: This is not true, the actual limit is 1.5MB. That test did test 1.5MB in stead of 1M. However, since we add a lot more stuff in work, this limit is actually kinda high. 500k is probably better but we don't need to address this here.

pkg/utils/controller/resource_snapshot_resolver.go
```diff
@@ -19,18 +19,453 @@ package controller
...
+}
+
+// WithConfig sets the configuration for the resolver and returns the resolver for method chaining
+func (r *ResourceSnapshotResolver) WithConfig(config *ResourceSnapshotConfig) *ResourceSnapshotResolver {
+	r.Config = config
+	return r
+}
+
+// WithSnapshotFunction sets the getOrCreateResourceSnapshot function and returns the resolver for method chaining
+func (r *ResourceSnapshotResolver) WithSnapshotFunction(fn GetOrCreateSnapshotFunction) *ResourceSnapshotResolver {
```
ryanzhang-oss: is this used anywhere?

pkg/utils/controller/resource_snapshot_resolver.go
```diff
@@ -19,18 +19,453 @@ package controller
...
+	r.GetOrCreateSnapshotFn = fn
+	return r
+}
+
+// WithDefaultSnapshotFunction sets the default getOrCreateResourceSnapshot implementation
+func (r *ResourceSnapshotResolver) WithDefaultSnapshotFunction() *ResourceSnapshotResolver {
```
ryanzhang-oss: where is GetOrCreateSnapshotFn used? Does it need to be public when we have this function?

pkg/utils/controller/resource_snapshot_resolver.go
```diff
@@ -19,18 +19,453 @@ package controller
...
+// NewResourceSnapshotResolver creates a new ResourceSnapshotResolver with the universal fields
+func NewResourceSnapshotResolver(client client.Client, scheme *runtime.Scheme) *ResourceSnapshotResolver {
+	return &ResourceSnapshotResolver{
+		Client: client,
+		Scheme: scheme,
+	}
+}
+
+// WithConfig sets the configuration for the resolver and returns the resolver for method chaining
+func (r *ResourceSnapshotResolver) WithConfig(config *ResourceSnapshotConfig) *ResourceSnapshotResolver {
```
ryanzhang-oss: does Config need to be public with this function?

pkg/utils/controller/resource_snapshot_resolver.go
```diff
@@ -19,18 +19,453 @@ package controller
...
+	// ResourceSnapshotCreationMinimumInterval is the minimum interval to create a new resourcesnapshot
+	// to avoid too frequent updates.
+	ResourceSnapshotCreationMinimumInterval time.Duration
+
+	// ResourceChangesCollectionDuration is the duration for collecting resource changes into one snapshot.
+	ResourceChangesCollectionDuration time.Duration
+}
+
+// NewResourceSnapshotConfig creates a ResourceSnapshotConfig with static values
+func NewResourceSnapshotConfig(creationInterval, collectionDuration time.Duration) *ResourceSnapshotConfig {
```
ryanzhang-oss, line 77: ditto

pkg/controllers/placement/controller.go
```diff
@@ -223,7 +213,7 @@ func (r *Reconciler) handleUpdate(ctx context.Context, placementObj fleetv1beta1
 		return ctrl.Result{}, err
 	}
 
-	createResourceSnapshotRes, latestResourceSnapshot, err := r.getOrCreateResourceSnapshot(ctx, placementObj, envelopeObjCount,
+	createResourceSnapshotRes, latestResourceSnapshot, err := r.ResourceSnapshotResolver.GetOrCreateSnapshotFn(ctx, placementObj, envelopeObjCount,
```
ryanzhang-oss: i am not entirely sure why we need create a function pointer but just call it direclty. In this case, why not call the  DefaultGetOrCreateResourceSnapshot function directly?

pkg/utils/controller/resource_snapshot_resolver.go
britaniar: This is copied over from the placement controller https://github.com/kubefleet-dev/kubefleet/blob/4cef5888b6171608d66636e672d3db11be78fef7/pkg/controllers/placement/controller.go#L54C1-L57C65
Need to update the comment another time...

pkg/controllers/placement/controller.go
britaniar: Done

pkg/utils/controller/resource_snapshot_resolver.go
```diff
@@ -106,7 +81,7 @@ func NewResourceSnapshotConfig(creationInterval, collectionDuration time.Duratio
 	}
 }
 
-// DefaultGetOrCreateResourceSnapshot is the default implementation
+// DefaultGetOrCreateResourceSnapshot is the default implementation.
 func (r *ResourceSnapshotResolver) DefaultGetOrCreateResourceSnapshot(ctx context.Context, placement fleetv1beta1.PlacementObj, envelopeObjCount int, resourceSnapshotSpec *fleetv1beta1.ResourceSnapshotSpec, revisionHistoryLimit int) (ctrl.Result, fleetv1beta1.ResourceSnapshotObj, error) {
```
ryanzhang-oss: I wonder what does "Default" here mean? Do we have other `GetOrCreateResourceSnapshot` function?

britaniar: We may eventually for update run. We do that implementation later. For now, I have restored to the original and we can talk about the refactor and sync for the design changes later as we each take on a controller.

PR #437, opened 2026-02-12
feat: add stage level metrics for update run

britaniar:

Fixes #

I have:
- Added two stage level histogram metrics:
     1. One to see how long a user takes to approve the before/after stage approval requests
     2. How long it takes a stage to complete excluding all stage tasks.
 - Added an integration case for testing only before stage task (missed previously and added to test the new metric)
 - Updated integrations tests to validate new emitted metrics.

- Integration Test

What the metrics look like:

```
fleet_workload_update_run_stage_cluster_updating_duration_seconds_bucket{namespace="prod",name="run-42",le="10"} 1
fleet_workload_update_run_stage_cluster_updating_duration_seconds_bucket{...,le="30"} 4
fleet_workload_update_run_stage_cluster_updating_duration_seconds_bucket{...,le="60"} 7
...
fleet_workload_update_run_stage_cluster_updating_duration_seconds_bucket{...,le="+Inf"} 9

fleet_workload_update_run_stage_cluster_updating_duration_seconds_sum{...} 1450
fleet_workload_update_run_stage_cluster_updating_duration_seconds_count{...} 9
```

```
fleet_workload_update_run_approval_request_latency_seconds_bucket{namespace="prod",name="run-42",stageTask="approve",le="60"} 3
fleet_workload_update_run_approval_request_latency_seconds_bucket{...,le="300"} 8
fleet_workload_update_run_approval_request_latency_seconds_bucket{...,le="900"} 12
...
fleet_workload_update_run_approval_request_latency_seconds_bucket{...,le="+Inf"} 15

fleet_workload_update_run_approval_request_latency_seconds_sum{...} 12450
fleet_workload_update_run_approval_request_latency_seconds_count{...} 15
```

### DIFF
--- a/pkg/controllers/updaterun/execution.go
+++ b/pkg/controllers/updaterun/execution.go
@@ -485,8 +485,7 @@ func (r *Reconciler) handleStageApprovalTask(
 				// Approved state should not change once the approval is accepted.
 				klog.V(2).InfoS("The approval request has been approval-accepted, ignoring changing back to unapproved", "approvalRequestTask", requestRef, "stage", updatingStage.Name, "updateRun", updateRunRef)
 			}
-			markStageTaskRequestApproved(stageTaskStatus, updateRun.GetGeneration())
-			recordApprovalRequestLatency(updatingStage, updateRun, stageTaskType)
+			markStageTaskRequestApproved(stageTaskStatus, updateRun, stageTaskType)
 		} else {
 			// retriable error
 			klog.ErrorS(err, "Failed to create the approval request", "approvalRequest", requestRef, "stage", updatingStage.Name, "updateRun", updateRunRef)
@@ -834,14 +833,16 @@ func markStageTaskRequestCreated(stageTaskStatus *placementv1beta1.StageTaskStat
 }
 
 // markStageTaskRequestApproved marks the Approval for the before or after stage task as Approved in memory.
-func markStageTaskRequestApproved(stageTaskStatus *placementv1beta1.StageTaskStatus, generation int64) {
+func markStageTaskRequestApproved(stageTaskStatus *placementv1beta1.StageTaskStatus, updateRun placementv1beta1.UpdateRunObj, taskType string) {
 	meta.SetStatusCondition(&stageTaskStatus.Conditions, metav1.Condition{
 		Type:               string(placementv1beta1.StageTaskConditionApprovalRequestApproved),
 		Status:             metav1.ConditionTrue,
-		ObservedGeneration: generation,
+		ObservedGeneration: updateRun.GetGeneration(),
 		Reason:             condition.StageTaskApprovalRequestApprovedReason,
 		Message:            "ApprovalRequest object is approved",
 	})
+
+	recordApprovalRequestLatency(stageTaskStatus, updateRun, taskType)
 }
 
 // markAfterStageWaitTimeElapsed marks the TimeWait after stage task as TimeElapsed in memory.
--- a/pkg/controllers/updaterun/metrics.go
+++ b/pkg/controllers/updaterun/metrics.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/apimachinery/pkg/api/meta"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 
 	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
@@ -69,16 +68,12 @@ func emitUpdateRunStatusMetric(updateRun placementv1beta1.UpdateRunObj) {
 
 // recordApprovalRequestLatency records the time from approval request creation to user approval.
 func recordApprovalRequestLatency(
-	updatingStage *placementv1beta1.StageConfig,
+	stageTaskStatus *placementv1beta1.StageTaskStatus,
 	updateRun placementv1beta1.UpdateRunObj,
 	taskType string,
 ) {
-	stageStatus := findStageStatus(updateRun, updatingStage.Name)
-	if stageStatus == nil {
-		return
-	}
-
-	approvalCreatedCond, approvalApprovedCond := findApprovalConditions(stageStatus, taskType)
+	approvalCreatedCond := meta.FindStatusCondition(stageTaskStatus.Conditions, string(placementv1beta1.StageTaskConditionApprovalRequestCreated))
+	approvalApprovedCond := meta.FindStatusCondition(stageTaskStatus.Conditions, string(placementv1beta1.StageTaskConditionApprovalRequestApproved))
 
 	// Only record latency when both approval request created and approved conditions are true,
 	// and their observed generation is the same as the update run generation to ensure the recorded latency is accurate.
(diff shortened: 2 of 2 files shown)


pkg/metrics/hub/metrics.go
```diff
@@ -42,6 +42,22 @@ var (
...
+		Help: "The latency from approval request creation to user approval in seconds",
+		// Buckets: 1min, 5min, 15min, 30min, 1hr, 2hr, 6hr, 12hr, 24hr
+		Buckets: []float64{60, 300, 900, 1800, 3600, 7200, 21600, 43200, 86400},
+	}, []string{"namespace", "name", "stageTask"})
+
+	// FleetUpdateRunStageClusterUpdatingDurationSeconds tracks the duration of each stage of an update run, excluding the execution time of stage tasks.
+	FleetUpdateRunStageClusterUpdatingDurationSeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "fleet_workload_update_run_stage_cluster_updating_duration_seconds",
+		Help: "The duration the stage of an update run in seconds without stage tasks execution time",
+		// Buckets: 10s, 30s, 1min, 2min, 5min, 10min, 30min, 1hr
```
jwtty: The minimum is 15s per current implementation.

pkg/controllers/updaterun/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+// recordApprovalRequestLatency records the time from approval request creation to user approval.
+func recordApprovalRequestLatency(
+	approvalRequest placementv1beta1.ApprovalRequestObj,
+	updateRun placementv1beta1.UpdateRunObj,
+	stageTaskType string,
+) {
+	approvedCond := meta.FindStatusCondition(
+		approvalRequest.GetApprovalRequestStatus().Conditions,
+		string(placementv1beta1.ApprovalRequestConditionApproved))
+	if approvedCond == nil {
```
jwtty: Also make sure the condition state is True.

pkg/controllers/updaterun/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+		return
+	}
+	latencySeconds := approvedCond.LastTransitionTime.Sub(
+		approvalRequest.GetCreationTimestamp().Time).Seconds()
```
jwtty: approveRequests can be recreated if user mistakenly deleted and this latency won't be accurate. Better to use updateRun side to keep track: The time difference between "ApprovalRequestCreated" condition and "ApprovalRequestApproved" condition.

pkg/metrics/hub/metrics.go
```diff
@@ -42,6 +42,22 @@ var (
...
 		Help: "Last update timestamp of update run status in seconds",
 	}, []string{"namespace", "name", "state", "condition", "status", "reason"})
+
+	// FleetUpdateRunApprovalRequestLatencySeconds tracks how long users take to approve approval requests.
+	FleetUpdateRunApprovalRequestLatencySeconds = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "fleet_workload_update_run_approval_request_latency_seconds",
+		Help: "The latency from approval request creation to user approval in seconds",
+		// Buckets: 1min, 5min, 15min, 30min, 1hr, 2hr, 6hr, 12hr, 24hr
+		Buckets: []float64{60, 300, 900, 1800, 3600, 7200, 21600, 43200, 86400},
+	}, []string{"namespace", "name", "stageTask"})
```
jwtty: "type" or "taskType" might be better for the 3rd label. Before I read the impl, I was not sure what this field means.

pkg/controllers/updaterun/controller.go
```diff
@@ -239,6 +239,8 @@ func (r *Reconciler) handleDelete(ctx context.Context, updateRun placementv1beta
 
 	// Delete the update run status metric.
 	hubmetrics.FleetUpdateRunStatusLastTimestampSeconds.DeletePartialMatch(prometheus.Labels{"namespace": updateRun.GetNamespace(), "name": updateRun.GetName()})
+	hubmetrics.FleetUpdateRunStageClusterUpdatingDurationSeconds.DeletePartialMatch(prometheus.Labels{"namespace": updateRun.GetNamespace(), "name": updateRun.GetName()})
```
jwtty: Move these functions to metrics.go too?

pkg/controllers/updaterun/execution.go
```diff
@@ -292,6 +292,13 @@ func (r *Reconciler) handleStageCompletion(
 ) (time.Duration, error) {
 	updateRunRef := klog.KObj(updateRun)
 
+	// Only record metric the first time cluster updating completes (before marking as waiting).
+	progressingCond := meta.FindStatusCondition(updatingStageStatus.Conditions, string(placementv1beta1.StageUpdatingConditionProgressing))
+	if progressingCond == nil || progressingCond.Reason == condition.StageUpdatingStartedReason {
```
jwtty: Does this cover the stop->resume scenario?

pkg/metrics/hub/metrics.go
ryanzhang-oss: What happens to two update run with the same name running on two different fleet? Will their metrics merge?

pkg/controllers/updaterun/controller.go
ryanzhang-oss: will the rest of the metrics of this updateRun get deleted before they are scraped by Prometheus?

pkg/metrics/hub/metrics.go
britaniar: They will not merge. Through kusto there is another label "cluster_id" which has their cppNamespace.

pkg/controllers/updaterun/controller.go
britaniar: The metrics are scraped every 15s so unless we complete the deletion before 15s after the metric was emitted then we should be okay.

pkg/controllers/updaterun/execution.go
britaniar: Yes. We will only enter here if we are done updating clusters. I am updating as there is one scenario we might hit, which is when we stop during the last cluster updating. In that case, the cluster finishes updating then the condition is progressing false. Then when we resume the condition may still be progressing false at this point since we have not more clusters updating and then the condition updated as waiting after this point.

pkg/controllers/updaterun/controller.go
ryanzhang-oss: that sounds like NOT okay? We will lose 15 s of metrics before the updateRun is deleted?

pkg/controllers/updaterun/execution.go
```diff
@@ -478,6 +486,7 @@ func (r *Reconciler) handleStageApprovalTask(
 				klog.V(2).InfoS("The approval request has been approval-accepted, ignoring changing back to unapproved", "approvalRequestTask", requestRef, "stage", updatingStage.Name, "updateRun", updateRunRef)
 			}
 			markStageTaskRequestApproved(stageTaskStatus, updateRun.GetGeneration())
+			recordApprovalRequestLatency(updatingStage, updateRun, stageTaskType)
```
ryanzhang-oss: why not call this inside markStageTaskRequestApproved function so it's clear the recording only happens once.

ryanzhang-oss: it seems that by calling it inside the approved, it can also simply some of the recordApprovalRequestLatency logic?

pkg/controllers/updaterun/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+	// and their observed generation is the same as the update run generation to ensure the recorded latency is accurate.
+	if !condition.IsConditionStatusTrue(approvalCreatedCond, updateRun.GetGeneration()) || !condition.IsConditionStatusTrue(approvalApprovedCond, updateRun.GetGeneration()) {
+		return
+	}
+
+	latencySeconds := approvalApprovedCond.LastTransitionTime.Sub(approvalCreatedCond.LastTransitionTime.Time).Seconds()
+	hubmetrics.FleetUpdateRunApprovalRequestLatencySeconds.WithLabelValues(
+		updateRun.GetNamespace(),
+		updateRun.GetName(),
+		taskType,
```
ryanzhang-oss, line 88: I wonder if the taskType is more important than the stage name? Here we omit the stage name but keep the taskType which I believe is only beforeStage and afterStage?

pkg/controllers/updaterun/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+// emitUpdateRunStatusMetric emits the update run status metric based on status conditions in the updateRun.
+func emitUpdateRunStatusMetric(updateRun placementv1beta1.UpdateRunObj) {
+	generation := updateRun.GetGeneration()
+	state := updateRun.GetUpdateRunSpec().State
+
+	updateRunStatus := updateRun.GetUpdateRunStatus()
+	succeedCond := meta.FindStatusCondition(updateRunStatus.Conditions, string(placementv1beta1.StagedUpdateRunConditionSucceeded))
+	if succeedCond != nil && succeedCond.ObservedGeneration == generation {
+		hubmetrics.FleetUpdateRunStatusLastTimestampSeconds.WithLabelValues(updateRun.GetNamespace(), updateRun.GetName(), string(state),
+			string(placementv1beta1.StagedUpdateRunConditionSucceeded), string(succeedCond.Status), succeedCond.Reason).SetToCurrentTime()
```
ryanzhang-oss, line 47: is the status not enough? The reason as a string might add a lot more cardinality to the data if someone is not aware that it's emited as a metrics. Status can only be true/false/unknown.

pkg/controllers/updaterun/execution.go
```diff
@@ -292,6 +292,14 @@ func (r *Reconciler) handleStageCompletion(
 ) (time.Duration, error) {
 	updateRunRef := klog.KObj(updateRun)
 
+	// Only record metric the first time cluster updating completes in the stage (before marking as waiting).
+	progressingCond := meta.FindStatusCondition(updatingStageStatus.Conditions, string(placementv1beta1.StageUpdatingConditionProgressing))
+	if progressingCond != nil && (!condition.IsConditionStatusFalse(progressingCond, updateRun.GetGeneration()) &&
```
ryanzhang-oss: Why checking 
> !condition.IsConditionStatusFalse(progressingCond, updateRun.GetGeneration()

Which seems to mean the progressingCond == true or unknown? I don't get it

pkg/controllers/updaterun/metrics.go
britaniar: We have to provide reason so we know why it is false. For our alerts the only one we would need the reason is for stuck. For a stuck update run he progressing is false but the update run also has a progressing condition as false for "stopped" state so we need to be able to know the difference for the alert to ensure it is stuck not stopped.

pkg/controllers/updaterun/metrics.go
britaniar: Trying to save on cardinality. This is a histogram so it would be a metric for each task type that includes all the stages.

PR #452, opened 2026-02-18
feat: new selection mode so that CRP can select some namespace resources

weng271190436:

new selection mode so that CRP can select some namespace resources

The motivation is allow managed namespace to select the namespace + some fleet managed resources. Currently the fleet managed resources are Network Policy and Resource Quota.

Going with option 1 of this [implementation one-pager](https://microsoft.sharepoint.com/:fl:/g/contentstorage/x8FNO-xtskuCRX2_fMTHLRvNdDK5qA9Irc36goWnhvw/IQBEvEvVSqhWRrtroiuDajw1AXfdHU8EwMo737vUU4leQTo?e=6d4K1m&nav=cz0lMkZjb250ZW50c3RvcmFnZSUyRng4Rk5PLXh0c2t1Q1JYMl9mTVRITFJ2TmRESzVxQTlJcmMzNmdvV25odncmZD1iJTIxcnN6eVZPX21sa3FvSDNrenVleU41ekUwVDZzejQ5TkZnVnFzdDBLTExBUFR4cU95TDBScVNvZTZfd3kxOGo0eCZmPTAxR0JXRDdaQ0VYUkY1S1NWSUtaRExXMjVDRk9CV1VQQlYmYz0lMkYmYT1Mb29wQXBwJnA9JTQwZmx1aWR4JTJGbG9vcC1wYWdlLWNvbnRhaW5lciZ4PSU3QiUyMnclMjIlM0ElMjJUMFJUVUh4dGFXTnliM052Wm5RdWMyaGhjbVZ3YjJsdWRDNWpiMjE4WWlGeWMzcDVWazlmYld4cmNXOUlNMnQ2ZFdWNVRqVjZSVEJVTm5ONk5EbE9SbWRXY1hOME1FdE1URUZRVkhoeFQzbE1NRkp4VTI5bE5sOTNlVEU0YWpSNGZEQXhSMEpYUkRkYVEwWTFSa3RQVEVoU1R6VldSVW96VGxGSlMwSXpOazlWTTBrJTNEJTIyJTJDJTIyaSUyMiUzQSUyMjg5OGZhNTUyLTFjMmYtNDNkNC04YTg3LWI3ODYyOWM1YmExMyUyMiU3RA%3D%3D)

Fixes: #487

I have:

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
+        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
         with:
           egress-policy: audit
 
--- a/cmd/hubagent/main.go
+++ b/cmd/hubagent/main.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 	clusterinventory "sigs.k8s.io/cluster-inventory-api/apis/v1alpha1"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -101,8 +102,18 @@ func main() {
 	// Set up controller-runtime logger
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
-	config := ctrl.GetConfigOrDie()
-	config.QPS, config.Burst = float32(opts.CtrlMgrOpts.HubQPS), opts.CtrlMgrOpts.HubBurst
+	// Create separate configs for the general access purpose and the leader election purpose.
+	//
+	// This aims to improve the availability of the hub agent; originally all access to the API
+	// server would share the same rate limiting configuration, and under adverse conditions (e.g.,
+	// large volume of concurrent placements) controllers would exhause all tokens in the rate limiter,
+	// and effectively starve the leader election process (the runtime can no longer renew leases),
+	// which would trigger the hub agent to restart even though the system remains functional.
+	defaultCfg := ctrl.GetConfigOrDie()
+	leaderElectionCfg := rest.CopyConfig(defaultCfg)
+
+	defaultCfg.QPS, defaultCfg.Burst = float32(opts.CtrlMgrOpts.HubQPS), opts.CtrlMgrOpts.HubBurst
+	leaderElectionCfg.QPS, leaderElectionCfg.Burst = float32(opts.LeaderElectionOpts.LeaderElectionQPS), opts.LeaderElectionOpts.LeaderElectionBurst
 
 	mgrOpts := ctrl.Options{
 		Scheme: scheme,
@@ -110,12 +121,27 @@ func main() {
 			SyncPeriod:       &opts.CtrlMgrOpts.ResyncPeriod.Duration,
 			DefaultTransform: cache.TransformStripManagedFields(),
 		},
-		LeaderElection:          opts.LeaderElectionOpts.LeaderElect,
-		LeaderElectionID:        "136224848560.hub.fleet.azure.com",
-		LeaderElectionNamespace: opts.LeaderElectionOpts.ResourceNamespace,
+		LeaderElection:       opts.LeaderElectionOpts.LeaderElect,
+		LeaderElectionConfig: leaderElectionCfg,
+		// If leader election is enabled, the hub agent by default uses a setup
+		// with a lease duration of 60 secs, a renew deadline of 45 secs, and a retry period of 5 secs.
+		// This setup gives the hub agent up to 9 attempts/45 seconds to renew its leadership lease
+		// before it loses the leadership and restarts.
+		//
+		// These values are set significantly higher than the controller-runtime defaults
+		// (15 seconds, 10 seconds, and 2 seconds respectively), as under heavy loads the hub agent
+		// might have difficulty renewing its lease in time due to API server side latencies, which
+		// might further lead to unexpected leadership losses (even when it is the only candidate
+		// running) and restarts.
+		//
+		// Note (chenyu1): a minor side effect with the higher values is that when the agent does restart,
+		// (or in the future when we do run multiple hub agent replicas), the new leader might have to wait a bit
+		// longer (up to 60 seconds) to acquire the leadership, which should still be acceptable in most scenarios.
 		LeaseDuration:           &opts.LeaderElectionOpts.LeaseDuration.Duration,
 		RenewDeadline:           &opts.LeaderElectionOpts.RenewDeadline.Duration,
 		RetryPeriod:             &opts.LeaderElectionOpts.RetryPeriod.Duration,
+		LeaderElectionID:        "136224848560.hub.fleet.azure.com",
+		LeaderElectionNamespace: opts.LeaderElectionOpts.ResourceNamespace,
 		HealthProbeBindAddress:  opts.CtrlMgrOpts.HealthProbeBindAddress,
 		Metrics: metricsserver.Options{
 			BindAddress: opts.CtrlMgrOpts.MetricsBindAddress,
@@ -128,7 +154,7 @@ func main() {
 	if opts.CtrlMgrOpts.EnablePprof {
 		mgrOpts.PprofBindAddress = fmt.Sprintf(":%d", opts.CtrlMgrOpts.PprofPort)
 	}
-	mgr, err := ctrl.NewManager(config, mgrOpts)
+	mgr, err := ctrl.NewManager(defaultCfg, mgrOpts)
 	if err != nil {
 		klog.ErrorS(err, "unable to start controller manager.")
 		exitWithErrorFunc()
@@ -186,7 +212,7 @@ func main() {
 	}
 
 	ctx := ctrl.SetupSignalHandler()
-	if err := workload.SetupControllers(ctx, &wg, mgr, config, opts); err != nil {
+	if err := workload.SetupControllers(ctx, &wg, mgr, defaultCfg, opts); err != nil {
 		klog.ErrorS(err, "unable to set up controllers")
 		exitWithErrorFunc()
 	}
--- a/cmd/hubagent/options/leaderelection.go
+++ b/cmd/hubagent/options/leaderelection.go
@@ -18,6 +18,8 @@ package options
 
 import (
 	"flag"
+	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/kubefleet-dev/kubefleet/pkg/utils"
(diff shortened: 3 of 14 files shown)


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -185,7 +185,28 @@ type ResourceSelectorTerm struct {
...
+	//   - NamespaceOnly: Only the namespace object itself
+	//   - NamespaceWithResources: The namespace AND all resources within it (default)
+	//   - NamespaceWithResourceSelectors: The namespace AND resources specified by additional selectors
+	//
+	// When SelectionScope is NamespaceWithResourceSelectors, you can define additional ResourceSelectorTerms
+	// (after the namespace selector) to specify which resources to include. These additional selectors can
+	// target both namespace-scoped resources (within the selected namespace) and cluster-scoped resources.
+	//
+	// Important requirements for NamespaceWithResourceSelectors mode:
+	// - The namespace selector MUST be the FIRST selector (index 0) in the ResourceSelectors array
```
weng271190436: The namespace selector doesn't have to be first for it to work but I think it is the most intuitive way to organize Namespace and other resources in a CRP by putting Namespace first.

Also can't think of a use case where Namespace should not be first.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -185,7 +185,28 @@ type ResourceSelectorTerm struct {
...
+	// - Exactly one namespace must be selected (by name or label matching one namespace)
```
weng271190436: Or I can restrict it to be selected by name only similar to NamespaceAccessibleCRP

apis/placement/v1beta1/clusterresourceplacement_types.go
ryanzhang-oss: What does this restriction buy us?

apis/placement/v1beta1/clusterresourceplacement_types.go
ryanzhang-oss: Name only sounds good, it's not easy to check label

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -185,7 +185,31 @@ type ResourceSelectorTerm struct {
...
+	// - Both requirements are validated at webhook admission time
+	// - If the selected namespace is deleted after CRP creation, the controller will report an error condition
+	//
+	// Example using NamespaceWithResourceSelectors:
+	// - First selector: {Group: "", Version: "v1", Kind: "Namespace", Name: "prod", SelectionScope: "NamespaceWithResourceSelectors"}
+	// - Second selector: {Group: "apps", Version: "v1", Kind: "Deployment", LabelSelector: {app: "frontend"}}
+	// - Third selector: {Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "ClusterRole", Name: "admin"}
```
ryanzhang-oss, line 214: this is a cluster scoped resource. How can we group the namespaced resources that are part of the namespace and the rest? Otherwise, it's just a list of mixed resources.

pkg/utils/validator/placement.go
```diff
@@ -65,6 +66,51 @@ var (
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// validateNamespaceWithResourceSelectorsMode validates that when NamespaceWithResourceSelectors mode is used,
+// there is exactly one namespace selector with this mode.
+func validateNamespaceWithResourceSelectorsMode(resourceSelectors []placementv1beta1.ResourceSelectorTerm) error {
```
ryanzhang-oss: need to add a test in the API only validation suite test/apis/placement/v1beta1/api_validation_integration_test.go

pkg/utils/controller/resource_selector_resolver.go
```diff
@@ -248,16 +280,58 @@ func (rs *ResourceSelectorResolver) gatherSelectedResource(placementKey types.Na
...
+			// Case 2: Non-namespace selector
+			if placementKey.Namespace != "" {
+				// RP: select resource with namespace from placementKey
+				objs, err = rs.fetchResources(selector, placementKey)
+			} else {
+				// CRP: check if cluster-scoped or namespace-scoped
+				isClusterScoped := rs.InformerManager.IsClusterScopedResources(gvk)
+				if isClusterScoped {
+					// Cluster-scoped resource: select with selector only
+					objs, err = rs.fetchResources(selector, placementKey)
```
ryanzhang-oss: fetchResources already have this check inside, no need to do this again. There are enough if/else here already.

test/e2e/placement_namespace_with_resourceselectors_test.go
```diff
@@ -0,0 +1,490 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	placementv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
+)
+
+// Tests for NamespaceWithResourceSelectors SelectionScope mode
+var _ = Describe("CRP with NamespaceWithResourceSelectors selecting single namespace and specific resources", Ordered, func() {
```
ryanzhang-oss, line 55: Just a though, those 4 cases have a lot in common, could probably use some refactoring to reduce duplication.

apis/placement/v1beta1/clusterresourceplacement_types.go
weng271190436: Force people to write intuitive yaml. Putting the namespace first calls out that the namespace is the pre-requisite for resources that come after without requiring the users to refer to the documentation.

apis/placement/v1beta1/clusterresourceplacement_types.go
weng271190436: I want to cover the use case where platform team want to deploy namespace + namespace-scoped resources under that namespace + some other cluster scoped resources (e.g. namespace-1-reader ClusterRole so that user can do kubectl get namespace-1 with this role but can't do kubectl get namespace-2)

One can argue that the user can just use 2 CRPs for this scenario, but grouping cluster scoped resources that support a namespace with the namespace under one CRP is a better way to organize them.

weng271190436: Also, there is existing known use case where people want to deploy a namespace + some cluster scoped resources. I think it is natural for these people to say "what if I still want to deploy the namespace + some cluster scoped resources, but also want to select a few namespace scoped resources within the namespace"

test/e2e/placement_namespace_with_resourceselectors_test.go
weng271190436: extracted some common functions

pkg/utils/controller/resource_selector_resolver.go
weng271190436: I removed the check where selectedNamespace is empty and throws error because fetchResources does that. But don't think I can remove the isClusterScoped check and pass different namespacedKey based on that unless I misunderstood fetchResources

apis/placement/v1beta1/clusterresourceplacement_types.go
weng271190436: discussed offline and removed this restriction

apis/placement/v1beta1/clusterresourceplacement_types.go
weng271190436: Sticking to this but without "namespace selector must come first" restriction

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -624,6 +624,321 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
...
+					},
+				},
+			}
+			err := hubClient.Create(ctx, &crp)
+			var statusErr *k8sErrors.StatusError
+			Expect(errors.As(err, &statusErr)).To(BeTrue(), fmt.Sprintf("Create CRP call produced error %s. Error type wanted is %s.", reflect.TypeOf(err), reflect.TypeOf(&k8sErrors.StatusError{})))
+			Expect(statusErr.ErrStatus.Message).Should(MatchRegexp("when using NamespaceWithResourceSelectors mode, only one namespace selector is allowed"))
+		})
+
+		It("should allow creation of CRP with NamespaceWithResourceSelectors selecting by name", func() {
```
ryanzhang-oss: why is this allow case in the deny test group?

pkg/utils/controller/resource_selector_resolver.go
ryanzhang-oss: I double checked the fetchResources, the only thing it cares is the namespace and the placement namespace must be the same as the selector namespace. I don't think we need this part either.

weng271190436: The issue is this part of fetchResources

```go
    if isNamespacedResource && placementKey.Namespace == "" {
		// If it's a namespace-scoped resource but placement has no namespace, return error.
		err := fmt.Errorf("invalid placement %s: cannot select namespace-scoped resource %v in a clusterResourcePlacement", placementKey, gvr)
		klog.ErrorS(err, "Invalid resource selector", "selector", selector)
		return nil, NewUserError(err)
	} else if !isNamespacedResource && placementKey.Namespace != "" {
		// If it's a cluster-scoped resource but placement has a namespace, return error.
		err := fmt.Errorf("invalid placement %s: cannot select cluster-scoped resource %v in a resourcePlacement", placementKey, gvr)
		klog.ErrorS(err, "Invalid resource selector", "selector", selector)
		return nil, NewUserError(err)
	}
```

if I have a CRP with a namespace-scoped resource I need to do 
```go
namespacedKey := types.NamespacedName{Name: placementKey.Name, Namespace: selectedNamespace}
objs, err = rs.fetchResources(selector, namespacedKey)
```

otherwise fetchResources will give me "invalid placement %s: cannot select namespace-scoped resource %v in a clusterResourcePlacement"

if I have a CRP with cluster-scoped resource I can continue with
```go
objs, err = rs.fetchResources(selector, placementKey)
```

weng271190436: Discussed offline. Simplified

test/apis/placement/v1beta1/api_validation_integration_test.go
weng271190436: Thanks for catching. Fixed

pkg/utils/controller/resource_selector_resolver.go
```diff
@@ -237,6 +237,38 @@ func generateRawContent(object *unstructured.Unstructured) ([]byte, error) {
...
+			// NamespaceWithResourceSelectors mode requires exactly one namespace
+			if len(namespaces) == 0 {
+				err := fmt.Errorf("invalid clusterResourcePlacement %s: NamespaceWithResourceSelectors mode requires exactly one namespace, but no namespaces were selected", placementKey.Name)
+				klog.ErrorS(err, "No namespaces selected with NamespaceWithResourceSelectors mode", "placement", placementKey.Name)
+				return nil, NewUserError(err)
+			}
+			if len(namespaces) > 1 {
+				err := fmt.Errorf("invalid clusterResourcePlacement %s: NamespaceWithResourceSelectors mode requires exactly one namespace, but %d namespaces were selected", placementKey.Name, len(namespaces))
+				klog.ErrorS(err, "Multiple namespaces selected with NamespaceWithResourceSelectors mode", "placement", placementKey.Name, "namespaceCount", len(namespaces))
+				return nil, NewUserError(err)
```
ryanzhang-oss, line 264: IIRC, we have CEL to not allow label selector so this should not happen?

pkg/utils/controller/resource_selector_resolver.go
```diff
@@ -530,6 +591,34 @@ func (rs *ResourceSelectorResolver) fetchResources(selector placementv1beta1.Res
 	return selectedObjs, nil
 }
 
+// fetchSelectedNamespaces retrieves the namespace names based on the namespace selector.
+func (rs *ResourceSelectorResolver) fetchSelectedNamespaces(selector placementv1beta1.ResourceSelectorTerm, placementName string) ([]string, error) {
+	klog.V(2).InfoS("start to fetch selected namespaces", "selector", selector, "placement", placementName)
+
+	// Use fetchResources to get namespace objects (handles label selector, name selector, deletion timestamps)
+	placementKey := types.NamespacedName{Name: placementName}
+	objs, err := rs.fetchResources(selector, placementKey)
```
ryanzhang-oss, line 600: this should only return at most one?

ryanzhang-oss (approved):

PR #458, opened 2026-02-20
feat: Lint helm

### LINKED ISSUE #212: [BUG] Helm charts do not pass `helm lint`
### **Describe the bug**
The charts as currently shown in the charts folder do not currently pass `helm lint`. This can cause some subtle usage problems, especially when trying to package them for standalone usage.

For example, the [insertion of a YAML file](https://github.com/kubefleet-dev/kubefleet/blob/24aee3272af3e647caa740a1c71e459f96b168e5/charts/member-agent/templates/crds/appliedworks.yaml#L3) has an incorrect indent.

Usage of relative paths / symlinks is problematic if the whole repo is not cloned or if the chart is attempted to be packaged.

It would be good if the charts all passed `helm lint` (even if there's no icon included) to ensure they can stand alone.

ryanzhang-oss:

Fixes #212 

I have:

### DIFF
--- a/.github/workflows/code-lint.yml
+++ b/.github/workflows/code-lint.yml
@@ -14,10 +14,9 @@ on:
 
 env:
   # Common versions
-  GO_VERSION: '1.24.13'
+  GO_VERSION: "1.24.13"
 
 jobs:
-
   detect-noop:
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/code-lint.yml
```diff
@@ -58,13 +57,33 @@ jobs:
...
+    if: needs.detect-noop.outputs.noop != 'true'
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v6.0.2
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v4
+        with:
+          version: v3.17.0
```
weng271190436, line 84: is this version the latest that can work with our helm chart? Just checked seems helm has v4 now and helm 3's latest version is v3.19.5 https://github.com/helm/helm/releases/tag/v3.19.5

I assume v4 doesn't work with our charts but v3.19.5 might work?

ryanzhang-oss: my client is 3.18.4 I am not sure if we want to always use the latest one.

PR #459, opened 2026-02-23
feat: refactor hub agent flag handling logic + add per-flag validation logic

michaelawyu:

This PR refactors the flag handling logic on the hub agent end and adds per-flag validation logic.

I have:

- [x] Unit tests

### DIFF
--- a/cmd/hubagent/options/featureflags.go
+++ b/cmd/hubagent/options/featureflags.go
@@ -24,10 +24,6 @@ import (
 
 // FeatureFlags is a set of feature flags the KubeFleet hub agent exposes.
 type FeatureFlags struct {
-	// Enable the hub agent to watch the KubeFleet v1alpha1 APIs or not.
-	// TO-DO (weiweng): remove this field soon. Only kept for backward compatibility.
-	EnableV1Alpha1APIs bool
-
 	// Enable the hub agent to watch the KubeFleet v1beta1 APIs or not. This flag is kept only for
 	// compatibility reasons; it has no effect at this moment, as KubeFleet v1alpha1 APIs have
 	// been removed and the v1beta1 APIs are the storage version in use.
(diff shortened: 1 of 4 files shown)


cmd/hubagent/options/featureflags.go
```diff
@@ -0,0 +1,149 @@
...
+	"flag"
+	"fmt"
+	"strconv"
+)
+
+// FeatureFlags is a set of feature flags the KubeFleet hub agent exposes.
+type FeatureFlags struct {
+	// Enable the hub agent to watch the KubeFleet v1alpha1 APIs or not.
+	// TO-DO (weiweng): remove this field soon. Only kept for backward compatibility.
+	EnableV1Alpha1APIs bool
```
ryanzhang-oss: can we move this now?

cmd/hubagent/options/featureflags.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// Enable the hub agent to watch the KubeFleet v1beta1 APIs or not. This flag is kept only for
+	// compatibility reasons; it has no effect at this moment, as KubeFleet v1alpha1 APIs have
+	// been removed and the v1beta1 APIs are the storage version in use.
+	EnableV1Beta1APIs bool
```
ryanzhang-oss, line 30: and even this one

cmd/hubagent/options/featureflags.go
```diff
@@ -0,0 +1,149 @@
...
+package options
+
+import (
+	"flag"
+	"fmt"
+	"strconv"
+)
+
+// FeatureFlags is a set of feature flags the KubeFleet hub agent exposes.
+type FeatureFlags struct {
```
ryanzhang-oss, line 26: pretty much all of them (except one) should probably be removed? Feature gates are supposed to stay forever

cmd/hubagent/options/featureflags.go
michaelawyu: Hi Ryan! We might need to drop the argument from the downstream deployment charts first, otherwise the agents will panic. I can send a PR.

cmd/hubagent/options/featureflags.go
michaelawyu: Same for this one as well. -> will drop this after the flag is removed from the artifacts.

cmd/hubagent/options/featureflags.go
michaelawyu: Yeah; would it be OK if I send it as a separate PR?

cmd/hubagent/options/featureflags.go
michaelawyu: Update: this flag has been dropped.

PR #461, opened 2026-02-23
feat: wire namespace affinity scheduler plugin

### LINKED ISSUE #501: [Feature] namespace affinity scheduler plugin that schedules resource placement to member clusters where namespace exists
There are cases where a namespace is not available on all member clusters. In this case kubefleet should schedule resource placement on member clusters where the namespace is available

weng271190436:

Namespace affinity scheduler plugin that schedules RP to clusters with namespace based on MemberCluster namespaces status.

The user scenario:
1. Platform team only place namespace to a few clusters.
2. Application team use RP to place their services to all or some of the clusters where their namespace is available. 

Fixes #501

I have:


test/e2e/mixed_placement_test.go
```diff
@@ -79,57 +82,8 @@ var _ = Describe("mixed ClusterResourcePlacement and ResourcePlacement negative
...
-					SelectedResources:     appConfigMapIdentifiers(),
-					ObservedResourceIndex: "0",
-				}
-				if diff := cmp.Diff(rp.Status, wantStatus, placementStatusCmpOptions...); diff != "" {
-					return fmt.Errorf("RP status diff (-got, +want): %s", diff)
-				}
-				return nil
-			}
-			Eventually(rpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update RP %s status as expected", rpName)
+			rpStatusUpdatedActual := rpStatusUpdatedActual(appConfigMapIdentifiers(), []string{memberCluster2EastCanaryName, memberCluster3WestProdName}, nil, "0")
```
weng271190436, line 86: Previously we expect memberCluster1EastProdName to fail. Now with namespace affinity plugin, memberCluster1EastProdName won't be selected so the test expectation changes

test/e2e/enveloped_object_placement_test.go
```diff
@@ -551,6 +551,10 @@ var _ = Describe("Test NamespaceOnly placement through CRP then resource placeme
 		}
 	})
 
+	It("should wait for namespace collection to sync on all member clusters", func() {
```
weng271190436, line 554: Namespace collection can take a while to finish so placing this in all RP tests where a CRP was created to place a namespace to member clusters where RP resources go

pkg/scheduler/profile/profile_test.go
```diff
@@ -0,0 +1,133 @@
...
+			namespaceaffinity.Plugin{},
+			sameplacementaffinity.Plugin{},
+			topologyspreadconstraints.Plugin{},
+			tainttoleration.Plugin{})); diff != "" {
+		t.Errorf("NewDefaultProfile() mismatch (-got +want):\n%s", diff)
+	}
+}
+
+// TestNewProfileWithOptions tests the creation of a scheduling profile with custom options.
+func TestNewProfileWithOptions(t *testing.T) {
```
ryanzhang-oss, line 82: what does this test actually test?

pkg/scheduler/profile/profile.go
```diff
@@ -51,13 +52,14 @@ func NewProfile(opts Options) *framework.Profile {
 		clusterAffinityPlugin = *opts.ClusterAffinityPlugin
```
ryanzhang-oss, line 52: I wonder why is ClusterAffinityPlugin customizable while others are static?

test/e2e/enveloped_object_placement_test.go
ryanzhang-oss: What happens if the namepace is not reported in the MC yet? The scheduler should be triggered again when the namespace is reported. Does the eventually block in the following IT not enough?

test/e2e/resource_placement_namespace_affinity_test.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+		})
+
+		It("should not place the ConfigMap on the cluster without the namespace", func() {
+			checkIfRemovedConfigMapFromMemberClusters([]*framework.Cluster{memberCluster3WestProd})
+		})
+	})
+})
+
+// Test scenario (PickN):
```
ryanzhang-oss, line 123: can we add one test that the NS is only on 1 cluster and Pick N =2?

test/e2e/utils_test.go
```diff
@@ -1733,6 +1800,35 @@ func createNamespaceOnlyCRP(crpName string) {
 	createCRPWithApplyStrategy(crpName, nil, namespaceOnlySelector())
 }
 
+// createNamespaceOnlyCRPForTwoClusters creates a CRP with PickFixed policy that places the
+// namespace on cluster-1 and cluster-2 only. Used by namespace affinity e2e tests to set up
+// a state where only 2 out of 3 member clusters have the target namespace.
+func createNamespaceOnlyCRPForTwoClusters(crpName string) {
```
michaelawyu: Hi weng271190436! I think maybe it's time for us to switch to a more generic warpper for create RP/CRP utilities in the E2E code 😂 We have a bit too many of different variants.

michaelawyu: Something like `createCRP(name, resourceSelectors, placementPolicy, strategy)` would suffice.

michaelawyu: Of course it's not a blocker for this PR in anyway; we can always address this later

test/e2e/resource_placement_drift_diff_test.go
```diff
@@ -96,6 +96,9 @@ var _ = Describe("take over existing resources using RP", Label("resourceplaceme
 		By("Validate CRP status is as expected")
 		crpStatusUpdatedActual := crpStatusUpdatedActual(workNamespaceIdentifiers(), allMemberClusterNames, nil, "0")
 		Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP status as expected")
+
+		By("waiting for namespace to be collected on all member clusters")
+		waitForNamespaceCollectionOnClusters(nsName, allMemberClusterNames)
```
michaelawyu: Hi weng271190436! Just a nit: I understand that this might be a must considering that it takes ~15 seconds for a namespace to be reported and our eventually block runs only for 10 seconds; but since it is a necessary step, would it be better if we make it a separate `It` block when applicable?

test/e2e/utils_test.go
```diff
@@ -836,6 +841,68 @@ func createNamespace() {
...
+// - CRPs dynamically create namespaces on member clusters
+// - ResourcePlacements need to use those newly created namespaces
+// - Namespace affinity plugin filtering is enabled
+func waitForNamespaceCollectionOnClusters(namespaceName string, clusterNames []string) {
+	GinkgoHelper()
+
+	// Skip waiting entirely when namespace collection is not enabled on any cluster.
+	// The namespace affinity plugin skips filtering in this case (backward compatibility),
+	// so there is no race to protect against.
+	if len(clusterNames) == 0 || !isNamespaceCollectionEnabled(clusterNames[0]) {
```
michaelawyu, line 880: Hi weng271190436! The namespace collection is enabled on a per-member basis, but here it only checks for one cluster? I understand that we might have some assumption that the collection is enabled in a all-or-nothing case, but if that's case we might need a comment here just to be clear.

test/e2e/setup.sh
```diff
@@ -217,6 +217,7 @@ do
             --set workApplierRequeueRateLimiterMaxFastBackoffDelaySeconds=5 \
             --set propertyProvider=$PROPERTY_PROVIDER \
             --set region=${REGIONS[$i]} \
+            --set enableNamespaceCollectionInPropertyProvider=true \
```
michaelawyu, line 219: Hi weng271190436! Since this is enabled per member agent and as a part of the property provider, we might need a check in the setup as well to make sure that it is running properly.

pkg/scheduler/profile/profile.go
michaelawyu, line 51: Yeah, a note on this part: originally we added the profile support so that developers (or users) can switch scheduler profiles as needed (e.g., for some placements, use profile A, and the others profile B). Each profile can have its own plugin chain (e.g., B has cluster affinity support, A does not); it seems that the options part was added to achieve the same purpose? Of course we haven't exposed the profile switch API yet.

michaelawyu: I checked the code and the options is not in use at all. Was it added back then to faciliate the development of the RP API?

test/e2e/utils_test.go
weng271190436: Make sense! Done

pkg/scheduler/profile/profile.go
weng271190436: Doesn't seem to be RP related. But it was added back recently https://teams.microsoft.com/l/message/19:13f7d68c3be840eb9281e41b6d1c15ea@thread.skype/1763091181077?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=e121dbfd-0ec1-40ea-8af5-26075f6a731b&parentMessageId=1763068248219&teamName=Azure%20Container%20Compute&channelName=SIG%20Multi-Cluster%20-%20Dataplane&createdTime=1763091181077

weng271190436: I think I get it. We want to avoid merge conflict when backporting so allowing customization here while using the customization capability in Azure/fleet

https://github.com/Azure/fleet/pull/1221/changes

pkg/scheduler/profile/profile.go
weng271190436: We want to avoid merge conflict when backporting so allowing customization here while using the customization capability in Azure/fleet

Azure/fleet uses a customized cluster affinity plugin to support cluster affinity based on VM SKU availability

https://github.com/Azure/fleet/pull/1221/changes

test/e2e/enveloped_object_placement_test.go
weng271190436: When the namespace is finally reported in MC, the member cluster watcher can detect it and triggers placement reconciliation.

The eventually block can work if we wait long enough. The current 10 second wait time is not enough. And waiting for namespace collection makes it more deterministic

test/e2e/setup.sh
weng271190436: The member clusters are not joined to the hub during setup.sh. If they are joined I can verify that the namespaces are populated on the hub membercluster CRs. But since they are not joined, I can't think of a good way to make sure that it is running properly here. When they are joined, I check NamespaceCollectionSucceededCondType in utils_test.go to make sure that it is working.

pkg/scheduler/profile/profile_test.go
weng271190436: I am testing calling NewProfile with option

test/e2e/utils_test.go
weng271190436: I created a method createGenericCRP because createCRP already exists.

I refactored utils in utils_test.go to use createGenericCRP. 

There are quite some places where crp is explicitly constructed without util methods. I didn't change those, thinking those might be out of scope

test/e2e/resource_placement_namespace_affinity_test.go
weng271190436: added

test/e2e/enveloped_object_placement_test.go
weng271190436: After pulling most waitForNamespaceCollectionOnClusters into its own It block based on 
michaelawyu's comment, many of the waits are after createRP because many createRP happens inside BeforeAll. So waitForNamespaceCollectionOnClusters doesn't have to happen before createRP.

There are exceptions. For example in "Test cluster scale out and shrink using pickN policy with namespaced staged update run," we expect PickN where N = 1 to pick kind-cluster-3. This can also be deterministic if the namespaces are ready before createRP.

test/e2e/mixed_placement_test.go
```diff
@@ -271,6 +234,9 @@ var _ = Describe("mixed ClusterResourcePlacement and ResourcePlacement negative
 				Expect(memberCluster.KubeClient.Create(ctx, &ns)).To(Succeed())
 			}
 
+			By("waiting for namespace to be collected on all member clusters")
```
weng271190436, line 237: Didn't convert this to It block (based on a comment) because the RP created right below depends on namespaces created above. So need to wait inside BeforeAll

test/e2e/staged_updaterun_test.go
```diff
@@ -679,6 +691,12 @@ var _ = Describe("test RP rollout with staged update run", Label("resourceplacem
 		updateRunNames := []string{}
 
 		BeforeAll(func() {
+			// Ensure the namespace collection is synced on all member clusters before creating the RP.
+			// We expect kind-cluster-3 to be picked by PickN with N=1 policy due to the alphabetical order of cluster names.
+			// If kind-cluster-3 does not have namespace synced during schedule time, the RP might pick kind-cluster-1 or kind-cluster-2.
+			By("should wait for namespace collection to sync on all member clusters")
```
weng271190436, line 697: Here also needs to wait inside BeforeAll. Commented on why

test/e2e/resource_placement_with_custom_config_test.go
```diff
@@ -71,6 +71,9 @@ var _ = Describe("validating RP when using customized resourceSnapshotCreationMi
 		crpStatusUpdatedActual := crpStatusUpdatedActual(workNamespaceIdentifiers(), allMemberClusterNames, nil, "0")
 		Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP status as expected")
 
+		By("waiting for namespace collection to sync on all member clusters")
```
weng271190436, line 74: In custom test, we expect "should not update RP status immediately". If I wait for namespace collection after the RP creation, the RP is likely to be updated when we expect it to be not updated yet.

test/e2e/resource_placement_drift_diff_test.go
weng271190436: I pulled them into It blocks in most cases. There are 3 cases where waiting in It block doesn't work and I commented below

Makefile
```diff
@@ -57,7 +57,7 @@ STATICCHECK_VER := v0.6.1
 STATICCHECK_BIN := staticcheck
 STATICCHECK := $(abspath $(TOOLS_BIN_DIR)/$(STATICCHECK_BIN)-$(STATICCHECK_VER))
 
-GOIMPORTS_VER := latest
+GOIMPORTS_VER := v0.42.0
```
weng271190436, line 60: fix

go: golang.org/x/tools/cmd/goimports@latest: golang.org/x/tools@v0.43.0 requires go >= 1.25.0 (running go 1.24.13; GOTOOLCHAIN=local)

test/e2e/setup.sh
michaelawyu: Hi Wei! I mean that there was a separate part of the check at the environment setup step, `checkIfAzurePropertyProviderIsWorking()`, for verifying if property provider is working properly before the tests start up.

michaelawyu: Come to think of that, for the test steps that require the property provider, we also need to mark them specifically so that they can be skipped if the environment has not provider available (`isAzurePropertyProviderEnabled`).

PR #463, opened 2026-02-24
feat: add namespace affinity scheduler plugin (not wired yet)

weng271190436:

add namespace affinity scheduler plugin to implement [this design](https://microsoft.sharepoint.com/:w:/r/teams/aks-caravel/_layouts/15/doc2.aspx?sourcedoc=%7BEC4121C8-9672-4AD6-93B7-7B0DC7CA7F1D%7D&file=CRP%20and%20RP%20affinity%20design.docx&action=default&mobileredirect=true)

The use case is that after platform team place a namespace to some clusters using CRP,  application team can place resources using RP to clusters with that namespace 

I have:

### DIFF
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -172,7 +172,7 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		Scheme:                   mgr.GetScheme(),
 		UncachedReader:           mgr.GetAPIReader(),
 		ResourceSelectorResolver: resourceSelectorResolver,
-		ResourceSnapshotResolver: *resourceSnapshotResolver,
+		ResourceSnapshotResolver: resourceSnapshotResolver,
 	}
 
 	rateLimiter := options.DefaultControllerRateLimiter(opts.RateLimiterOpts)
@@ -310,8 +310,10 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 			}
 			klog.Info("Setting up clusterStagedUpdateRun controller")
 			if err = (&updaterun.Reconciler{
-				Client:          mgr.GetClient(),
-				InformerManager: dynamicInformerManager,
+				Client:                   mgr.GetClient(),
+				InformerManager:          dynamicInformerManager,
+				ResourceSelectorResolver: resourceSelectorResolver,
+				ResourceSnapshotResolver: resourceSnapshotResolver,
 			}).SetupWithManagerForClusterStagedUpdateRun(mgr); err != nil {
 				klog.ErrorS(err, "Unable to set up clusterStagedUpdateRun controller")
 				return err
@@ -326,8 +328,10 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 				}
 				klog.Info("Setting up stagedUpdateRun controller")
 				if err = (&updaterun.Reconciler{
-					Client:          mgr.GetClient(),
-					InformerManager: dynamicInformerManager,
+					Client:                   mgr.GetClient(),
+					InformerManager:          dynamicInformerManager,
+					ResourceSelectorResolver: resourceSelectorResolver,
+					ResourceSnapshotResolver: resourceSnapshotResolver,
 				}).SetupWithManagerForStagedUpdateRun(mgr); err != nil {
 					klog.ErrorS(err, "Unable to set up stagedUpdateRun controller")
 					return err
--- a/pkg/controllers/placement/suite_test.go
+++ b/pkg/controllers/placement/suite_test.go
@@ -126,7 +126,7 @@ var _ = BeforeSuite(func() {
 		UncachedReader:           mgr.GetAPIReader(),
 		Recorder:                 mgr.GetEventRecorderFor(controllerName),
 		ResourceSelectorResolver: resourceSelectorResolver,
-		ResourceSnapshotResolver: *resourceSnapshotResolver,
+		ResourceSnapshotResolver: resourceSnapshotResolver,
 	}
 	opts := options.RateLimitOptions{
 		RateLimiterBaseDelay:  5 * time.Millisecond,
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -60,6 +60,12 @@ type Reconciler struct {
 	recorder record.EventRecorder
 	// the informer contains the cache for all the resources we need to check the resource scope.
 	InformerManager informer.Manager
+
+	// ResourceSelectorResolver selects resources for placement.
+	ResourceSelectorResolver controller.ResourceSelectorResolver
+
+	// ResourceSnapshotResolver gets or creates resource snapshots.
+	ResourceSnapshotResolver controller.ResourceSnapshotResolver
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtime.Result, error) {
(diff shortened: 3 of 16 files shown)


apis/cluster/v1beta1/membercluster_types.go
```diff
@@ -141,6 +141,14 @@ type MemberClusterStatus struct {
 	// +optional
 	ResourceUsage ResourceUsage `json:"resourceUsage,omitempty"`
 
+	// Namespaces is a map of namespace names to their associated work names for namespaces
+	// that are managed by Fleet (i.e., have AppliedWork owner references when created).
+	// The key is the namespace name and the value is the work name from the AppliedWork owner reference.
+	// If the namespace does not have an AppliedWork owner reference, the value will be an empty string.
+	// This field is copied from the corresponding InternalMemberCluster object.
+	// +optional
+	Namespaces map[string]string `json:"namespaces,omitempty"`
```
weng271190436, line 150: unlike the original design, I am just saving the namespaces as status rather than label. Think status is a better place because label is desired state and status is observed state. The namespaces on each member cluster are observed states.

pkg/scheduler/framework/plugins/namespaceaffinity/filtering.go
```diff
@@ -0,0 +1,93 @@
...
+	cluster *clusterv1beta1.MemberCluster,
+) (status *framework.Status) {
+	// Get the target namespace for this ResourcePlacement.
+	nsName := ps.GetNamespace()
+
+	// Check if namespace collection is enabled for this cluster.
+	// The condition can have three states:
+	// 1. Missing: namespace collection is not enabled (backward compatibility - skip filtering)
+	// 2. True: namespace collection is working normally
+	// 3. False: namespace collection is enabled but degraded (limit reached - still use the data)
```
weng271190436, line 64: When limit reached, I still respect the namespaces listed because at least I am not trying to schedule the namespaces to random clusters where the namespace is not available. But this also means that "limit reached" state can have unexpected consequence and it might not be easy for users to figure out what's wrong.

pkg/scheduler/framework/plugins/namespaceaffinity/filtering.go
```diff
@@ -0,0 +1,93 @@
...
+	namespaceCollectionEnabled := false
+	for _, cond := range cluster.Status.Conditions {
```
michaelawyu: Hi weng271190436! Just a nit: we've been using `meta.FindStatusCondition` utility pretty commonly in the code base; we also have a condition utility package (`IsConditionStatusTrue`/`IsConditionStatusTrue` functions) that might be of use. Of course there is nothing wrong with the current style and this certainly is not a blocker.

apis/cluster/v1beta1/membercluster_types.go
michaelawyu, line 150: Just a thought not directly related to this PR: our member cluster objects (MC and IMC) are getting larger and larger (it is not uncommon for a production cluster to have hundreds/thousands namespaces) -> we need to be a bit careful on any frequent writes on such objects.

pkg/scheduler/watchers/membercluster/watcher.go
```diff
@@ -253,6 +253,17 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 				}
 			}
 
+			// Capture namespace collection changes.
+			//
+			// This is important for the namespace affinity plugin to trigger rescheduling
+			// when namespaces are added or removed from clusters.
+			oldNamespaces := oldCluster.Status.Namespaces
+			newNamespaces := newCluster.Status.Namespaces
+			if !equality.Semantic.DeepEqual(oldNamespaces, newNamespaces) {
```
michaelawyu, line 262: Hi weng271190436! Just a side note about this part: this setup would lead to a situation where one addition of a namespace on a member cluster would trigger all PickAll placements (which is the most commonly used placement type) on the hub cluster for a rescheduling attempt -> which can be quite expensive in a large scale deployment, esp. considering that most of the time this will yield no chanages on the scheduling results.

michaelawyu: But I do want to say that this is not a blocker of any sort for the PR. We probably need to re-think the watcher part a bit when we've gotten more time.

test/scheduler/namespaceaffinity_integration_test.go
```diff
@@ -0,0 +1,335 @@
...
+
+		// Update namespaces map
+		mc.Status.Namespaces = namespaces
+
+		// Update condition based on whether namespace collection is enabled
+		if enabled {
+			// Add/update the NamespaceCollectionSucceeded condition to True
+			conditionExists := false
+			for i := range mc.Status.Conditions {
+				if mc.Status.Conditions[i].Type == propertyprovider.NamespaceCollectionSucceededCondType {
```
michaelawyu: Hi weng271190436! A nit: similar to a comment below, this feels like a good place to use `meta.SetStatusCondition` utility.

test/scheduler/namespaceaffinity_integration_test.go
```diff
@@ -0,0 +1,335 @@
...
+					Reason:             propertyprovider.NamespaceCollectionSucceededReason,
+					Message:            propertyprovider.NamespaceCollectionSucceededMsg,
+					LastTransitionTime: metav1.Now(),
+				})
+			}
+		} else {
+			// Remove the condition entirely (namespace collection not enabled)
+			// This is different from ConditionFalse which means degraded/limit reached
+			filteredConditions := make([]metav1.Condition, 0)
+			for i := range mc.Status.Conditions {
```
michaelawyu: And for this one, `meta.RemoveStatusCondition`, if you feel like it :D

michaelawyu: (And the one below).

apis/cluster/v1beta1/membercluster_types.go
ryanzhang-oss: IIRC, P95 of the AKS cluster have less than like 30 NS

ryanzhang-oss: even P99 isn't very high

pkg/scheduler/watchers/membercluster/watcher.go
weng271190436: I think this is a good point because for a single cluster namespace change, it should only affect a few resource placements. And pratically it affects 0 RP most of the time and 1 RP in some cases. Reconciling CRP + RP is very wasteful.

Maybe can do this as a followup:
```go
return ctrl.NewControllerManagedBy(mgr).
        Named("membercluster-scheduler-watcher").
        WithEventFilter(customPredicate).
        Watches(
            &clusterv1beta1.MemberCluster{},
            handler.Funcs{
                UpdateFunc: r.handleClusterUpdate,
                DeleteFunc: r.handleClusterDelete,
            },
        ).
        Complete(r)
```

And within UpdateFunc and DeleteFunc, do the following in a loop:
1. find CRP/RPs affected by Label change
2. find CRP/RPs affected by Taint change
3. find CRP/RPs affected by Namespace change
4. ...

Union all affected CRP/RPs and queue

apis/cluster/v1beta1/membercluster_types.go
weng271190436: I think storing all namespaces is necessary because:
1. I want to make scheduling decision based on availbility of each namespace
->
2. I need to track all available namespaces on each cluster
->
3. I need to store all available namespaces on each cluster somewhere available to scheduler

On 1, one thing we can do is to only consider fleet-owned namespaces (can tell by owner reference). If user removes owner reference the namespace is no longer considered fleet-owned and workload won't be scheduled there. I think this is most promising but changes functionality.

On 2, nothing I can think of to optimize.

On 3, maybe can split MC and IMC into multiple objects if the concern is MC and IMC size and we don't want to send large payload to api server. But there is overhead when you have more objects. It is unclear if splitting is going to help.

pkg/scheduler/framework/plugins/namespaceaffinity/filtering.go
weng271190436: thank you for the suggestion. Fixed!

PR #478, opened 2026-03-04
feat: add CEL validation rules to MemberCluster taint API types

### LINKED ISSUE #490: Feature: improve reliability by moving away from using webhook
We are trying to move away from this as the webhook could go down anytime depending on the API Server.

Instead of relying on the webhook we want to convert as much of the validation to CEL. [Common Expression Language in Kubernetes | Kubernetes](https://kubernetes.io/docs/reference/using-api/cel/)

ytimocin:

Adds CRD CEL validation for `MemberCluster` taints in both API versions (`v1` and `v1beta1`), replacing the webhook-based taint validation with declarative CRD-level validation.

### What changed

- API type updates:
  - Added CEL rules on `MemberClusterSpec.Taints` and `Taint` in:
    - `apis/cluster/v1/membercluster_types.go`
    - `apis/cluster/v1beta1/membercluster_types.go`
  - Rules cover:
    - taint key name segment format and max length
    - taint key prefix DNS-subdomain format and max length
    - taint value label-value format and max length
    - taint uniqueness by `(key, value, effect)`
  - Added `MaxLength`, `MinLength`, and `MaxItems` constraints to bound CEL cost
- Regenerated CRD schema:
  - `config/crd/bases/cluster.kubernetes-fleet.io_memberclusters.yaml`
- Removed webhook taint validation:
  - `pkg/webhook/membercluster/membercluster_validating_webhook.go` — CREATE/UPDATE now returns `Allowed` since CEL handles validation before the webhook is reached. DELETE path (ServiceExport check) is unchanged.
  - `pkg/utils/validator/membercluster.go` — Deleted (dead code)
  - `pkg/utils/validator/membercluster_test.go` — Deleted (dead code)
- Added integration validation coverage:
  - `test/apis/cluster/v1/api_validation_integration_test.go`
  - `test/apis/cluster/v1beta1/api_validation_integration_test.go`
  - Comprehensive valid/invalid taint CEL cases for both versions.
- Updated e2e assertion:
  - `test/e2e/webhook_test.go`
  - Adjusted expected error message to match CEL validation output.

### Why

- Enforces invalid taints earlier at CRD admission time (schema validation runs before webhooks).
- Removes runtime dependency on the webhook for taint validation — if the webhook is unavailable, taint validation still works.
- Simplifies webhook code by removing redundant validation logic.

### Breaking change note

Error responses for invalid taints have changed:
- **Before**: Webhook returned `403 Forbidden` with messages from `k8s.io/apimachinery/pkg/util/validation` (e.g., `"name part must consist of..."`)
- **After**: CRD validation returns `422 Unprocessable Entity` with CEL rule messages (e.g., `"taint key name segment must consist of..."`)

If external tooling parses taint validation error messages or status codes, it will need to be updated.

- `make reviewable` passes clean
- Integration tests:
  - `go test ./test/apis/cluster/v1 -run TestAPIs -count=1 -- -ginkgo.focus="MemberCluster taint CEL validation" -ginkgo.v`
  - `go test ./test/apis/cluster/v1beta1 -run TestAPIs -count=1 -- -ginkgo.focus="MemberCluster taint CEL validation" -ginkgo.v`

- CEL validation runs at schema validation time, before validating webhooks. Invalid taints are rejected before reaching the webhook.
- The webhook DELETE path is preserved — it performs cross-object validation (checking `InternalServiceExport` resources) that cannot be expressed in CEL.
- Estimated total CEL cost is ~4.4M out of the 10M per-object budget (~56% headroom).

I have:


apis/cluster/v1/membercluster_types.go
```diff
@@ -136,12 +137,21 @@ type MemberClusterStatus struct {
 
 // Taint attached to MemberCluster has the "effect" on
 // any ClusterResourcePlacement that does not tolerate the Taint.
+// +kubebuilder:validation:XValidation:rule="(self.key.contains('/') ? self.key.substring(self.key.indexOf('/') + 1) : self.key).matches('^[A-Za-z0-9]([A-Za-z0-9._-]*[A-Za-z0-9])?$')",message="taint key name segment must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character"
```
Arvindthiru, line 140: The CEL validation for name and label seem too complex ? Currently we depend on validation functions provided by k8s which guarantee that we are following established conventions once this CEL validation is merged we would also need to maintain consistency between our validation and k8s (very unlikely but still something to keep in mind)

PR #489, opened 2026-03-05
feat: refactor member agent flag handling logic + add per-flag validation logic

michaelawyu:

This PR refactors the flag handling logic on the hub agent end and adds per-flag validation logic.

I have:

N/A

N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: "[Feature]"
+labels: enhancement
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context or screenshots about the feature request here.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -13,6 +13,7 @@ Fixes #
 
 I have:
 
+- [ ] Associated this change with a known KubeFleet Issue (Bug, Feature, etc).
 - [ ] Run `make reviewable` to ensure this PR is ready for review.
 
 ### How has this code been tested
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,33 @@ updates:
       interval: "weekly"
     commit-message:
       prefix: "chore"
+
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "chore"
+    labels:
+      - "dependencies"
+    open-pull-requests-limit: 10
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
+    groups:
+      kubernetes:
+        patterns:
+          - "k8s.io/*"
+          - "sigs.k8s.io/*"
+      azure-sdk:
+        patterns:
+          - "github.com/Azure/*"
+          - "github.com/AzureAD/*"
+      all-others:
+        patterns:
+          - "*"
+        exclude-patterns:
+          - "k8s.io/*"
+          - "sigs.k8s.io/*"
+          - "github.com/Azure/*"
+          - "github.com/AzureAD/*"
--- a/.github/workflows/chart.yml
+++ b/.github/workflows/chart.yml
@@ -57,18 +57,19 @@ jobs:
         run: |
           set -euo pipefail
 
-          RELEASE_VERSION="${{ needs.export-registry.outputs.version }}"
-
+          RELEASE_TAG="${{ needs.export-registry.outputs.tag }}"
+          CHART_VERSION="${{ needs.export-registry.outputs.version }}"
           OCI_REGISTRY="${{ needs.export-registry.outputs.registry }}/charts"
-          make helm-push REGISTRY="${OCI_REGISTRY}" TAG="${RELEASE_VERSION}"
+
+          make helm-push REGISTRY="${OCI_REGISTRY}" TAG="${RELEASE_TAG}" CHART_VERSION="${CHART_VERSION}"
 
       - name: Verify chart appVersion matches release tag
         run: |
           set -euo pipefail
 
-          RELEASE_VERSION="${{ needs.export-registry.outputs.version }}"
-          CHART_VERSION="${RELEASE_VERSION}"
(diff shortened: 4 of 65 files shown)


docker/member-agent.Dockerfile
```diff
@@ -13,13 +13,13 @@ COPY go.sum go.sum
 RUN go mod download
 
 # Copy the go source
-COPY cmd/memberagent/main.go main.go
+COPY cmd/memberagent cmd/memberagent/
```
ryanzhang-oss, line 16: is there any reason for this change?

cmd/memberagent/options/options.go
```diff
@@ -0,0 +1,68 @@
...
+	// cluster to the member cluster.
+	ApplierOpts ApplierOptions
+
+	// KubeFleet cluster property provider related options.
+	PropertyProviderOpts PropertyProviderOptions
+
+	// The fields below are added only for backwards compatibility reasons.
+	// Their values are never read.
+	UseV1Alpha1APIs bool
+	UseV1Beta1APIs  bool
```
ryanzhang-oss: where are those still used?

docker/member-agent.Dockerfile
michaelawyu: Hi Ryan! This is needed as previously we copy only the `main.go` file for the member agent; however, right now a new folder `cmd/memberagent/options` is added in the same pattern as the hub agent, and we need to copy this as well otherwise things won't build.

cmd/memberagent/options/options.go
michaelawyu: Hi Ryan! The v1beta1 API flag is still in use in the upstream chart; PR #502 will drop it. I saw that the v1alpha1 API flag has been removed everywhere -> will remove it from the code.

cmd/memberagent/options/ctrlmanager.go
```diff
@@ -0,0 +1,301 @@
...
+	return (*HubBurstValueWithValidation)(p)
+}
+
+type MemberQPSValueWithValidation float64
+
+func (v *MemberQPSValueWithValidation) String() string {
+	return fmt.Sprintf("%f", *v)
+}
+
+func (v *MemberQPSValueWithValidation) Set(s string) error {
```
weng271190436, line 251: Do you think it makes sense to extract common code from 

```go
func (v *MemberQPSValueWithValidation) Set(s string) error
```
and

```go
func (v *HubQPSValueWithValidation) Set(s string) error
```

?

cmd/memberagent/main.go
```diff
@@ -203,23 +152,26 @@ func main() {
...
-	if *enablePprof {
-		memberOpts.PprofBindAddress = fmt.Sprintf(":%d", *pprofPort)
-		hubOpts.PprofBindAddress = fmt.Sprintf(":%d", *hubPprofPort)
+	if opts.CtrlManagerOptions.EnablePprof {
+		memberOpts.PprofBindAddress = fmt.Sprintf(":%d", opts.CtrlManagerOptions.MemberManagerOpts.PprofPort)
+		hubOpts.PprofBindAddress = fmt.Sprintf(":%d", opts.CtrlManagerOptions.HubManagerOpts.PprofPort)
 	}
 
-	if err := Start(ctrl.SetupSignalHandler(), hubConfig, memberConfig, hubOpts, memberOpts); err != nil {
+	if err := Start(ctrl.SetupSignalHandler(), hubConfig, memberConfig, hubOpts, memberOpts, *opts); err != nil {
```
weng271190436, line 174: Is there a reason why *opts is passed by value but the existing pattern of hubOpts and memberOpts is passed by reference?

cmd/memberagent/options/applier.go
```diff
@@ -0,0 +1,415 @@
...
+		"work-applier-requeue-rate-limiter-attempts-with-fixed-delay",
+		"The number of attempts with a fixed delay before the KubeFleet member agent switches to exponential backoff when re-processing a placement. Default is 1. The value must be in the range [1, 40].")
+
+	flags.Var(
+		newRequeueFixedDelaySecondsValue(5, &o.RequeueRateLimiterFixedDelaySeconds),
+		"work-applier-requeue-rate-limiter-fixed-delay-seconds",
+		"The fixed delay in seconds for the KubeFleet member agent when re-processing a placement before switching to exponential backoff. Default is 5 seconds. The value must be in the range [2, 300].")
+
+	flags.Var(
+		newRequeueExpBaseForSlowBackoffValue(1.2, &o.RequeueRateLimiterExponentialBaseForSlowBackoff),
```
weng271190436, line 132: Though we don't have cross validation between slow backoff and fast backoff, the correct way to configure these is to have fast backoff value <= slow backoff value (or strictly less than)?

cmd/memberagent/options/applier.go
```diff
@@ -0,0 +1,415 @@
...
+	return (*ResForceDeletionWaitTimeMinutes)(p)
+}
+
+type RequeueAttemptsWithFixedDelay int
+
+func (v *RequeueAttemptsWithFixedDelay) String() string {
+	return fmt.Sprintf("%d", *v)
+}
+
+func (v *RequeueAttemptsWithFixedDelay) Set(s string) error {
```
weng271190436, line 203: Shall we add unit tests for these? The validation logic of these Set functions are straight forward but still make sense to have unit tests?

cmd/memberagent/options/ctrlmanager.go
michaelawyu: Hi weng271190436! I was thinking the same thing; kept them separate for now just in case the validation logic might become different and for the reason that the duplication seems to be minor. Please let me know if this is a concern; I don't particularly have a strong opinion on this TBH.

cmd/memberagent/main.go
michaelawyu: Hi weng271190436! All the opts here (`ctrl.Options` for `hubOpts` and `memberOpts`; `options.Options` for `globalOpts`) are values I believe.

cmd/memberagent/options/applier.go
michaelawyu: Ah, I must have forgotten to prompt the preparation for the cross-field validation code. Let me add that.

michaelawyu: Added.

cmd/memberagent/options/applier.go
michaelawyu: Hi weng271190436! Yeah, there should be UTs; they will be sent as a separate PR, similar to #467 .

cmd/memberagent/main.go
weng271190436: oh right! It just looks different because opts is a pointer but hubOpts and memberOpts are values.

cmd/memberagent/options/ctrlmanager.go
weng271190436: Even if they become different, most likely just this part will be different "if qps < 10.0 || qps > 10000.0 {"?

And that can still be extracted into a helper where lower bound and upper bound can be passed in. But not big deal

PR #511, opened 2026-03-16
fix(chart): add resource requests/limits to member-agent initContainer

### LINKED ISSUE #510: Missing `resources`on fleet-member-agent initContainer
The fleet-member-agent deployment includes an initContainer without any resource requests or limits defined (resources: {}). This came up during policy testing on Arc‑enabled clusters joined to a fleet and is more of a best‑practice gap than a functional error, but it has some practical impact in policy‑restricted environments.

The initContainer currently has no CPU/memory requests or limits
This can trigger policy or compliance warnings in clusters that enforce resource requirements
Adding basic defaults would improve out‑of‑box behavior.

Image

BillionClaw:

Fixes #510

This PR adds proper resource requests and limits to the fleet-member-agent deployment's initContainer. This addresses policy compliance warnings in restricted environments (e.g., Arc-enabled clusters) that enforce resource requirements.

Changes:
- Added `initContainerResources` values to values.yaml with sensible defaults
- Added `init-provider-token` initContainer to the deployment template
- The initContainer uses the same image as refresh-token for consistency
- Resources are only applied when `useCAAuth` is false (when the refresh token flow is used)

- Verified the Helm chart template syntax is valid
- The initContainer resources follow the same pattern as other containers in the chart
- Default resource values are conservative (50m CPU, 64Mi memory requests)

The initContainer is added conditionally only when `useCAAuth` is false, which is when the refresh-token sidecar and provider-token volume are also used. This maintains consistency with the existing deployment logic.

I have:
- [x] Associated this change with a known KubeFleet Issue (Bug, etc).
- [ ] Run `make reviewable` to ensure this PR is ready for review. (Chart changes only)

### DIFF
--- a/charts/member-agent/templates/deployment.yaml
+++ b/charts/member-agent/templates/deployment.yaml
@@ -16,6 +16,18 @@ spec:
     spec:
       restartPolicy: Always
       serviceAccountName: {{ include "member-agent.fullname" . }}-sa
+      {{- if not .Values.useCAAuth }}
+      initContainers:
+        - name: init-provider-token
+          image: "{{ .Values.refreshtoken.repository }}:{{ .Values.refreshtoken.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.refreshtoken.pullPolicy }}
+          command: ["sh", "-c", "echo 'Initializing provider token'"]
+          resources:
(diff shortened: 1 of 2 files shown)


michaelawyu:
Closing this PR as it should be targeted at a different repo.

PR #527, opened 2026-03-19
ci: add GitHub issues to Azure DevOps sync workflow

ytimocin:

This pull request introduces a new GitHub Actions workflow, `.github/workflows/sync-issues-to-ado.yml`, that securely and automatically syncs approved GitHub issues to Azure DevOps (ADO) as work items. The workflow emphasizes security, least privilege, and supply-chain safety by using direct API calls and robust input validation. It only syncs issues labeled with `ado-sync` by maintainers, and provides a manual trigger for ad-hoc syncing.

**Security and Trust Boundaries**

- Only issues labeled `ado-sync` by maintainers are eligible for syncing. Manual syncs are also label-gated to prevent bypasses.
- The workflow is scoped to run only on the upstream repository, not forks, and the `sync` job declares `environment: issues` so secrets sit behind environment-level access approval per the [secure-use guidance](https://docs.github.com/en/actions/reference/security/secure-use). The required-reviewers rule on that environment is what makes the gate non-decorative — see "Before first run" below.
- All user and input data (issue numbers, titles, bodies) is validated and safely escaped to prevent shell, script, HTML, and WIQL injection (env-var pass-through, `jq --arg`, `@html` escaping, random-delimiter heredocs for `GITHUB_OUTPUT`).

**Supply-Chain and Permissions Hardening**

- Replaces third-party ADO actions with direct REST API calls, removing that supply-chain link.
- Pinned `actions/github-script` to a full-length commit SHA.
- Uses `permissions: { issues: write }` at workflow level. Splitting into two jobs so the ADO PAT and the GitHub-write token never coexist in one process is tracked in #680.

**Review feedback addressed (commit `bda8ac7f`)**

- Gate `labeled` events on `github.event.label.name == 'ado-sync'` so adding any other label to an `ado-sync` issue no longer re-triggers a sync.
- Paginate `listComments` so the link-back marker is found even when an issue has >30 comments.
- URL-encode `ADO_ORG` and `ADO_PROJECT` in both the curl base URL and the JS link-back URL.
- Validate `ISSUE_NUMBER` is strictly numeric **before** any `gh issue view` call on the manual-dispatch path.

I have:

- [ ] Associated this change with a known KubeFleet Issue (Bug, Feature, etc). _No precursor issue exists; follow-up hardening tracked in #680._

Locally tested with [act](https://github.com/nektos/act):

```sh
act issues \
  -e test-issue-event.json \
  -W .github/workflows/sync-issues-to-ado.yml \
  -n \
  --container-architecture linux/amd64 \
  --env GITHUB_REPOSITORY=kubefleet-dev/kubefleet
```

- `-e test-issue-event.json` — fake issue event payload
- `-n` — dry run (no containers actually started)
- `--env GITHUB_REPOSITORY=...` — overrides the repo guard so the job isn't skipped on a fork

The workflow has not yet run end-to-end against a real ADO instance. The **first labelled issue post-merge is the first true e2e**, so the configuration in "Before first run" must be in place before any issue gets labelled `ado-sync` in production.

### Before first run (required configuration)

The job declares `environment: issues`, so the secrets below should be configured on that environment (not just at repo level) so they inherit the environment's access-approval rule.

**Environment secrets to add to the `issues` environment** (only names are listed; values are set in repo settings — never paste real values into a PR/issue/comment):

| Name | Purpose |
|---|---|
| `ADO_PAT` | Azure DevOps Personal Access Token. Scope: **Work Items: Read & Write**. Project-scoped to the target project. Set an expiration. |
| `ADO_ORGANIZATION` | Azure DevOps organization (URL path component). |
| `ADO_PROJECT` | Azure DevOps project name. Used for the URL path and the WIQL `[System.TeamProject]` filter. |
| `ADO_AREA_PATH` | `System.AreaPath` value applied to new work items. |
| `ADO_ITERATION_PATH` | `System.IterationPath` value applied to new work items. |

The auto-provided `GITHUB_TOKEN` (referenced as `${{ github.token }}` and implicitly used by `actions/github-script`) does not need to be added — it is issued per run.

**Repo configuration (no YAML, all in Settings):**

1. Create the `issues` environment under Settings → Environments. Configure all five secrets above on it.
2. Configure the **required reviewers** rule on `issues` (per the [secure-use guidance](https://docs.github.com/en/actions/reference/security/secure-use)). Without it, the environment gate is decorative — anyone with write access could trigger a `workflow_dispatch` and reach the PAT.
3. Restrict the environment's **deployment branch policy** to `main`, so a `workflow_dispatch` from a feature branch cannot run a modified workflow that exfiltrates the PAT.
4. Create the `ado-sync` label in the repo. The workflow gates on it; without the label no issue is syncable.
5. Designate a maintainer (or rotation team) responsible for `ADO_PAT` rotation before its expiration.

**Bonus pre-flight verification:**

- Confirm the configured `ADO_PAT` authenticates against ADO (one-shot `curl` against `_apis/projects` returns 200) before labelling the first issue.
- After the first real run, manually skim the workflow run logs to confirm no unredacted secret material appears (per the secure-use page's log-redaction guidance).

Follow-up hardening tracked in #680: job split for tighter `permissions` scoping, GitHub-App-vs-PAT evaluation for the ADO credential, and `CODEOWNERS` for `.github/workflows/`.

### DIFF
--- a/.github/workflows/sync-issues-to-ado.yml
+++ b/.github/workflows/sync-issues-to-ado.yml
@@ -34,11 +34,20 @@ jobs:
   # Only users with write access can add labels, so labeling itself is the trust boundary.
   # The "issues" environment requires reviewer approval before credentials are accessed.
   sync:
-    # [Repo scoping] Only runs on the upstream repo, not on forks
+    # [Repo scoping] Only runs on the upstream repo, not on forks.
+    # The `labeled` event is gated on the specific label name so that adding any unrelated
+    # label to an issue that already carries `ado-sync` does not re-trigger a sync; closed and
+    # reopened still gate on label-presence because the action payload omits label.name there.
     if: >-
       github.repository == 'kubefleet-dev/kubefleet'
       && (
-        (github.event_name == 'issues' && contains(github.event.issue.labels.*.name, 'ado-sync'))
+        (
+          github.event_name == 'issues'
+          && (
+            (github.event.action == 'labeled' && github.event.label.name == 'ado-sync')
+            || (github.event.action != 'labeled' && contains(github.event.issue.labels.*.name, 'ado-sync'))
+          )
+        )
         || github.event_name == 'workflow_dispatch'
       )
     runs-on: ubuntu-latest
@@ -59,7 +68,19 @@ jobs:
           set -euo pipefail
           if [ "${EVENT_NAME}" = "workflow_dispatch" ]; then
             ISSUE_NUMBER="${INPUT_ISSUE_NUMBER}"
-            # [Label gate] Validate label even on manual dispatch to prevent bypass
+          else
+            ISSUE_NUMBER="${EVENT_ISSUE_NUMBER}"
+          fi
+
+          # Validate issue number is strictly numeric BEFORE any downstream call,
+          # to prevent malformed input from reaching gh / curl / shell expansions.
+          if ! [[ "${ISSUE_NUMBER}" =~ ^[0-9]+$ ]]; then
+            echo "::error::Invalid issue number: not a positive integer"
+            exit 1
+          fi
+
+          if [ "${EVENT_NAME}" = "workflow_dispatch" ]; then
+            # [Label gate] Validate label even on manual dispatch to prevent bypass.
             LABELS=$(gh issue view "${ISSUE_NUMBER}" \
               --repo "${GITHUB_REPOSITORY}" \
               --json labels --jq '.labels[].name')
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-issues-to-ado.yml
```diff
@@ -0,0 +1,252 @@
...
+        || github.event_name == 'workflow_dispatch'
+      )
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    environment: issues
+    steps:
+      - name: Resolve issue details
+        id: issue
+        env:
+          GH_TOKEN: ${{ github.token }}
```
Tatsinnit, line 60: 💡 Thoughts, only - but something you/we should consider https://docs.github.com/en/actions/tutorials/authenticate-with-github_token

.github/workflows/sync-issues-to-ado.yml
```diff
@@ -0,0 +1,252 @@
...
+      # (any public GitHub user) to push content changes into ADO after the initial label
+      # approval, bypassing human review. Use workflow_dispatch to re-sync if needed.
+      - closed # Updates ADO work item state to Closed
+      - reopened # Updates ADO work item state back to New
+      - labeled # Triggers sync when maintainer applies ado-sync label
+
+  # [Workflow dispatch] Manual trigger for ad-hoc syncing of specific issues
+  workflow_dispatch:
+    inputs:
+      issue_number:
```
Tatsinnit, line 16: 💡Thought Share: Consider requiring review for access to secrets 

https://docs.github.com/en/actions/reference/security/secure-use

PR #553, opened 2026-03-26
test: add UTs for the refactored member agent CLI flags setup

michaelawyu:

This PR adds UTs to the refactored member agent CLI flag setup as mentioned in #489.

I have:

- [NA] Associated this change with a known KubeFleet Issue (Bug, Feature, etc).

N.A

N/A

### DIFF
--- a/cmd/memberagent/options/options_test.go
+++ b/cmd/memberagent/options/options_test.go
@@ -18,6 +18,7 @@ package options
 
 import (
 	"flag"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
@@ -216,14 +217,14 @@ func TestCtrlManagerOptions(t *testing.T) {
 			flagSetName:      "hubQPSOutOfRangeTooSmall",
 			args:             []string{"--hub-api-qps=9.9"},
 			wantErred:        true,
-			wantErrMsgSubStr: "QPS limit is set to an invalid value",
+			wantErrMsgSubStr: fmt.Sprintf("QPS limit is set to an invalid value (%f), must be a value in the range [10.0, 1000.0]", 9.9),
 		},
 		{
 			name:             "hub client QPS out of range (too large)",
 			flagSetName:      "hubQPSOutOfRangeTooLarge",
 			args:             []string{"--hub-api-qps=1000.1"},
 			wantErred:        true,
-			wantErrMsgSubStr: "QPS limit is set to an invalid value",
+			wantErrMsgSubStr: fmt.Sprintf("QPS limit is set to an invalid value (%f), must be a value in the range [10.0, 1000.0]", 1000.1),
 		},
 		{
 			name:             "hub client burst parse error",
@@ -237,14 +238,14 @@ func TestCtrlManagerOptions(t *testing.T) {
 			flagSetName:      "hubBurstOutOfRangeTooSmall",
 			args:             []string{"--hub-api-burst=9"},
 			wantErred:        true,
-			wantErrMsgSubStr: "burst limit is set to an invalid value",
+			wantErrMsgSubStr: fmt.Sprintf("burst limit is set to an invalid value (%d), must be a value in the range [10, 2000]", 9),
 		},
 		{
 			name:             "hub client burst out of range (too large)",
 			flagSetName:      "hubBurstOutOfRangeTooLarge",
 			args:             []string{"--hub-api-burst=2001"},
 			wantErred:        true,
-			wantErrMsgSubStr: "burst limit is set to an invalid value",
+			wantErrMsgSubStr: fmt.Sprintf("burst limit is set to an invalid value (%d), must be a value in the range [10, 2000]", 2001),
 		},
 		{
 			name:        "negative member client QPS value",
@@ -287,14 +288,14 @@ func TestCtrlManagerOptions(t *testing.T) {
 			flagSetName:      "memberQPSOutOfRangeTooSmall",
 			args:             []string{"--member-api-qps=9.9"},
 			wantErred:        true,
-			wantErrMsgSubStr: "QPS limit is set to an invalid value",
+			wantErrMsgSubStr: fmt.Sprintf("QPS limit is set to an invalid value (%f), must be a value in the range [10.0, 10000.0]", 9.9),
 		},
 		{
 			name:             "member client QPS out of range (too large)",
 			flagSetName:      "memberQPSOutOfRangeTooLarge",
 			args:             []string{"--member-api-qps=10000.1"},
 			wantErred:        true,
-			wantErrMsgSubStr: "QPS limit is set to an invalid value",
+			wantErrMsgSubStr: fmt.Sprintf("QPS limit is set to an invalid value (%f), must be a value in the range [10.0, 10000.0]", 10000.1),
 		},
 		{
 			name:             "member client burst parse error",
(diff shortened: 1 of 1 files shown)


cmd/memberagent/options/options_test.go
```diff
@@ -0,0 +1,711 @@
...
+			name:             "hub client burst out of range (too large)",
+			flagSetName:      "hubBurstOutOfRangeTooLarge",
+			args:             []string{"--hub-api-burst=2001"},
+			wantErred:        true,
+			wantErrMsgSubStr: "burst limit is set to an invalid value",
+		},
+		{
+			name:        "negative member client QPS value",
+			flagSetName: "memberQPSNegative",
+			args:        []string{"--member-api-qps=-5"},
```
weng271190436, line 253: No need to address but now I am thinking:

If a user passes --member-api-qps=-5, is it more likely that the user made a typo, or is it more likely that the user means no QPS throttling?

cmd/memberagent/options/options_test.go
```diff
@@ -0,0 +1,711 @@
...
+			args:             []string{"--member-api-qps=9.9"},
+			wantErred:        true,
+			wantErrMsgSubStr: "QPS limit is set to an invalid value",
+		},
+		{
+			name:             "member client QPS out of range (too large)",
+			flagSetName:      "memberQPSOutOfRangeTooLarge",
+			args:             []string{"--member-api-qps=10000.1"},
+			wantErred:        true,
+			wantErrMsgSubStr: "QPS limit is set to an invalid value",
```
weng271190436: nit: I think the later part of the error message is an important behavior that is worth testing. For better user experience, it is not enough to say it is an invalid value. The valid range is also important.

cmd/memberagent/options/options_test.go
```diff
@@ -0,0 +1,711 @@
...
+			args:             []string{"--member-api-burst=abc"},
+			wantErred:        true,
+			wantErrMsgSubStr: "failed to parse int value",
+		},
+		{
+			name:             "member client burst out of range (too small)",
+			flagSetName:      "memberBurstOutOfRangeTooSmall",
+			args:             []string{"--member-api-burst=9"},
+			wantErred:        true,
+			wantErrMsgSubStr: "burst limit is set to an invalid value",
```
weng271190436: nit: I think the later part of the error message is an important behavior that is worth testing. For better user experience, it is not enough to say it is an invalid value. The valid range is also important.

cmd/memberagent/options/options_test.go
```diff
@@ -0,0 +1,711 @@
...
+		},
+		{
+			name:             "member client burst out of range (too large)",
+			flagSetName:      "memberBurstOutOfRangeTooLarge",
+			args:             []string{"--member-api-burst=20001"},
+			wantErred:        true,
+			wantErrMsgSubStr: "burst limit is set to an invalid value",
```
weng271190436: nit: I think the later part of the error message is an important behavior that is worth testing. For better user experience, it is not enough to say it is an invalid value. The valid range is also important.

cmd/memberagent/options/options_test.go
```diff
@@ -0,0 +1,711 @@
...
+			args:             []string{"--deletion-wait-time=abc"},
+			wantErred:        true,
+			wantErrMsgSubStr: "failed to parse integer value",
+		},
+		{
+			name:             "deletion wait time out of range (too small)",
+			flagSetName:      "deletionWaitTimeOutOfRangeTooSmall",
+			args:             []string{"--deletion-wait-time=0"},
+			wantErred:        true,
+			wantErrMsgSubStr: "resource force deletion wait time in minutes is set to an invalid value",
```
weng271190436: nit: I think the later part of the error message is an important behavior that is worth testing. For better user experience, it is not enough to say it is an invalid value. The valid range is also important.

Same can be said about below wantErrMsgSubStr which contains "invalid value"

cmd/memberagent/options/options_test.go
```diff
@@ -0,0 +1,711 @@
...
+				EnablePriorityQueue:                                                   false,
+				RequeueRateLimiterAttemptsWithFixedDelay:                              1,
+				RequeueRateLimiterFixedDelaySeconds:                                   5,
+				RequeueRateLimiterExponentialBaseForSlowBackoff:                       1.2,
+				RequeueRateLimiterInitialSlowBackoffDelaySeconds:                      2,
+				RequeueRateLimiterMaxSlowBackoffDelaySeconds:                          15,
+				RequeueRateLimiterExponentialBaseForFastBackoff:                       1.5,
+				RequeueRateLimiterMaxFastBackoffDelaySeconds:                          900,
+				RequeueRateLimiterSkipToFastBackoffForAvailableOrDiffReportedWorkObjs: true,
+				PriorityLinearEquationCoEffA:                                          -3,
```
weng271190436, line 377: Just a question and nothing to address here. Are these default values tuned such that they are good for a "typical" use case? I think it is difficult to justify why pick -3 instead of -5 for this placement age coefficient so most users won't know how to customize these. Did we pick -3 because it yields good performance in the load tests?

cmd/memberagent/options/options_test.go
michaelawyu: Hi weng271190436! In `client-go`, the current validation logic is essentially: if QPS == 0, fall back to the default value; if < 0, do not set up the rate limiter at all (no flow control). Here we basically follow the same pattern, but just assume that all negative values == -1 and use a more strict range. I totally understand your point though; but I guess it is a bit difficult to make guesses.

cmd/memberagent/options/options_test.go
michaelawyu: Hi weng271190436! Originally we just added a few static priority levels in the system, but later it was suggested that we should give users a bit more flexibility. The default values here (-3 * age in minutes + 100), as explained in the comments/notes, give a placement around ~30 minutes of time with prioritized processing (with -5 it is ~20 minutes).

michaelawyu: Note that (for obvious reasons) prioritized processing cannot help if placements appear faster than the speed at which our controller can handle (though in reality this typically won't occur; we do assume a cap on the # of placements). The settings are primarily helpful in cases where our agents restart/resync, as this might bring about a period where the system becomes less responsive (due to the controllers becoming busy re-processing existing placements). To gain the full effect, the time of prioritized processing should be a bit longer than the time it takes for an agent to complete resyncing -> but the time needed for resyncing can vary significantly across deployments, esp. on the upstream side. 30 minutes should be enough given what perf test data we have on hand (in the 1K/1K scenario resync takes ~10-20 minutes), but there's really no guarantee.

cmd/memberagent/options/options_test.go
michaelawyu: Done.

cmd/memberagent/options/options_test.go
michaelawyu: Done. Thanks 🙏

cmd/memberagent/options/options_test.go
michaelawyu: Done. Thanks 🙏

cmd/memberagent/options/options_test.go
michaelawyu: Done. Thanks 🙏

PR #567, opened 2026-04-05
fix: refresh binding status when work is first observed with reported status

### LINKED ISSUE #558: [BUG] Placement status might fail to sync when caches are lagging
### **Describe the bug**
Placements might see a stale status `ApplyInProgress`, despite the fact that object might have become applied and/or available, when the caches are lagging behind. This happens because KubeFleet's work generator controller always assumes the `ApplyInProgress` state right after a work object has been created, and it will ignore CREATE events; if a Work object's status has stablized when the cache first sees the object, the binding object will never receive a status update, and the placement status is stuck.

### **Environment**
N/A

### **To Reproduce**
This is most often triggered under heavy load.

### **Expected behavior**
Placements should have synchronized status.

officialasishkumar:

This PR makes the work generator recover when its cache first observes a `Work` object after the member side has already reported status. In that case the create event now requeues the parent binding so the binding and placement status can be refreshed instead of remaining stuck at `ApplyInProgress`.

To avoid changing the normal fresh-work flow, the create handler only enqueues when the created `Work` already carries reported status.

Fixes #558

I have:

- `go test ./pkg/controllers/workgenerator -run TestWorkHandlerCreateFunc -count=1`
- `go test ./pkg/controllers/workgenerator -run TestGetWorkNamePrefixFromSnapshotName -count=1`

- The new create-event path is intentionally gated on `Work` objects that already have reported status to avoid regressing the existing `ApplyInProgress` behavior for freshly created works without member-side status yet.

### DIFF
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -1470,7 +1470,7 @@ func extractDiffedResourcePlacementsFromWork(work *fleetv1beta1.Work) []fleetv1b
 }
 
 // SetupWithManagerForClusterResourceBinding sets up the controller with the Manager.
-// It watches clusterResourceBinding events and also update/delete events for work.
+// It watches clusterResourceBinding events and also create/update/delete events for work.
 func (r *Reconciler) SetupWithManagerForClusterResourceBinding(mgr controllerruntime.Manager) error {
 	r.recorder = mgr.GetEventRecorderFor("cluster resource binding work generator")
 	return controllerruntime.NewControllerManagedBy(mgr).Named("cluster-resource-binding-work-generator").
@@ -1481,7 +1481,7 @@ func (r *Reconciler) SetupWithManagerForClusterResourceBinding(mgr controllerrun
 }
 
 // SetupWithManagerForResourceBinding sets up the controller with the Manager.
-// It watches resourceBinding events and also update/delete events for work.
+// It watches resourceBinding events and also create/update/delete events for work.
 func (r *Reconciler) SetupWithManagerForResourceBinding(mgr controllerruntime.Manager) error {
 	r.recorder = mgr.GetEventRecorderFor("resource binding work generator")
 	return controllerruntime.NewControllerManagedBy(mgr).Named("resource-binding-work-generator").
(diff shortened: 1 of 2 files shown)


pkg/controllers/workgenerator/controller.go
```diff
@@ -1498,8 +1498,48 @@ func shouldIgnoreWork(enqueueCRB bool, parentNamespaceName string) bool {
 	return false
 }
 
+func hasReportedWorkStatus(work *fleetv1beta1.Work) bool {
```
michaelawyu, line 1501: Hi Asish! Thanks for the patch (and sorry for the delayed response) 🙏

michaelawyu: A minor comment on this one: as the work applier refreshes the status as a whole unit, it should be fine just to check if the status is set (or if .conditions is set).

pkg/controllers/workgenerator/controller.go
```diff
@@ -1498,8 +1498,48 @@ func shouldIgnoreWork(enqueueCRB bool, parentNamespaceName string) bool {
...
+				return
+			}
+			if !hasReportedWorkStatus(work) {
+				klog.V(5).InfoS("Ignoring the work create event as the work has no reported status yet", "work", klog.KObj(work))
+				return
+			}
+			parentBindingName, exist := evt.Object.GetLabels()[fleetv1beta1.ParentBindingLabel]
+			if !exist {
+				klog.ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("created work has no binding parent")),
+					"Could not find the parent binding label", "created work", evt.Object, "existing label", evt.Object.GetLabels())
```
michaelawyu, line 1534: Hi Asish! For brevity reasons, maybe it's best to use `klog.KObj(evt.Object) instead here.

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -52,6 +57,136 @@ var statusCmpOptions = []cmp.Option{
...
+			}
+
+			workHandlerFuncs(tt.enqueueCRB).CreateFunc(context.Background(), event.CreateEvent{Object: tt.work}, queue)
+
+			validateEnqueueBehavior(t, queue, tt.wantEnqueueKeys)
+		})
+	}
+}
+
+func validateEnqueueBehavior(t *testing.T, queue *controllertest.Queue, wantEnqueueKeys []reconcile.Request) {
```
michaelawyu, line 170: Hi Asish! Just a style nit about this: if there is no other UTs using this helper function, do you think it would be OK if we inline it? As the data involved here is simple and the queue is ordered, we could just do a direct diff if you are fine with it.

michaelawyu (commented):
Added some minor comments, PTAL. Thanks again for contributing to KubeFleet!

PR #577, opened 2026-04-06
fix: update correct stage status when update run fails in "Stop" state

britaniar:

I have:  addressed a bug found from AI review comments on https://github.com/kubefleet-dev/kubefleet/pull/571

 **Bug Fix:** Fixed a bug in pkg/controllers/updaterun/stop.go where updatingStageStatus was not being set before the deferred checkIfErrorStagedUpdateAborted function captured it. 


- UT


 **Problem**                                                                                                                                                                                                                                                                                                            
     In the stop function, the variable updatingStageStatus (line 41) was declared as nil and the defer statement on line 44-46 captured it immediately. When stopUpdatingStage was called, the updatingStageStatus remained nil in the stop function's scope, so if an errStagedUpdatedAborted error occurred, the deferred function couldn't properly mark the stage as failed.

### DIFF
--- a/pkg/controllers/updaterun/stop_test.go
+++ b/pkg/controllers/updaterun/stop_test.go
@@ -35,15 +35,15 @@ import (
 
 func TestStop(t *testing.T) {
 	tests := []struct {
-		name                         string
-		updateRun                    *placementv1beta1.ClusterStagedUpdateRun
-		updatingStageIndex           int
-		toBeUpdatedBindings          []placementv1beta1.BindingObj
-		toBeDeletedBindings          []placementv1beta1.BindingObj
-		wantErr                      bool
-		wantFinished                 bool
-		wantStageSucceededCond       *metav1.Condition
-		wantDeleteStageSucceededCond *metav1.Condition
+		name                             string
+		updateRun                        *placementv1beta1.ClusterStagedUpdateRun
+		updatingStageIndex               int
+		toBeUpdatedBindings              []placementv1beta1.BindingObj
+		toBeDeletedBindings              []placementv1beta1.BindingObj
+		wantErr                          bool
+		wantFinished                     bool
+		wantStageSucceededCond           *metav1.Condition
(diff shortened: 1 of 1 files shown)


pkg/controllers/updaterun/stop_test.go
```diff
@@ -33,6 +33,207 @@ import (
...
+				if succeededCond == nil {
+					t.Fatalf("stop() expected StageUpdatingConditionSucceeded condition to be set, got nil")
+				}
+				if diff := cmp.Diff(*tt.wantStageSucceededCond, *succeededCond, cmpOptions...); diff != "" {
+					t.Errorf("stop() StageUpdatingConditionSucceeded mismatch (-want +got):\n%s", diff)
+				}
+			}
+
+			// Verify deletion stage succeeded condition if expected.
+			if tt.wantDeleteStageSucceededCond != nil {
```
weng271190436: Can also verify DeletionStageStatus is nil when we don't want it to be populated? The bug is that we can populate DeletionStageStatus incorrectly so I think it is good to verify that it is not populated when it is not supposed to be.

pkg/controllers/updaterun/stop_test.go
```diff
@@ -33,6 +33,207 @@ import (
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       "test-update-run",
+					Generation: 1,
+				},
+				Spec: placementv1beta1.UpdateRunSpec{
+					PlacementName:         "test-placement",
+					ResourceSnapshotIndex: "1",
+				},
+				Status: placementv1beta1.UpdateRunStatus{
+					StagesStatus: []placementv1beta1.StageUpdatingStatus{}, // Empty - all stages done.
```
weng271190436: In practice StagesStatus should still be populated when all stages done? It looks like this is to simulate the case where updatingStageIndex >= len(StagesStatus)

If I am understanding correctly, maybe just say that in the comment because "Empty - all stages done" can be confusing.

PR #578, opened 2026-04-06
feat: auto-add member-cluster-name label to MemberCluster for override selection by cluster name

### LINKED ISSUE #101: Allow `ResourceOverride` to select a cluster by cluster name
We have some use cases where certain specific member clusters require overrides. However, there's no obvious way to indicate that a `ResourceOverride` should apply to a _specific cluster by name_ - only `labelSelector` is currently supported (#100).

While we could easily add a label with the cluster name to the `MemberCluster`, we then run into issues with use of the API or Terraform provider (at least as far as the Azure Kubernetes Fleet Manager implementation is concerned - [I was directed to file issues here](https://github.com/Azure/fleet/issues/1146#issuecomment-2957557410), so here I am).

Via the Terraform provider or API, there's not a clear way to add a label to the `MemberCluster` entities. We could add something post-linking, but from an API/IaC perspective, that's configuration drift and may result in the labels being removed later.

Having support for a `propertySelector` on `ResourceOverride` where one of the allowed properties is `name` would solve this issue.

ytimocin:

Automatically adds a kubernetes-fleet.io/member-cluster-name label to MemberCluster objects during reconciliation, enabling users to select clusters by name in ResourceOverride and ClusterResourceOverride via labelSelector.

- **New clusters**: Label is added on first reconcile (same mechanism as the finalizer)
- **Existing clusters after upgrade**: Label is added on the next heartbeat-triggered reconcile — no manual action, no downtime
- **Self-healing**: If removed, the next reconcile restores it
- **No infinite reconcile loop**: GenerationChangedPredicate filters out metadata-only updates

Fixes #101 

I have:

Locally, yes. Also added necessary tests and updated existing ones.

N/A

### DIFF
--- a/apis/placement/v1beta1/commons.go
+++ b/apis/placement/v1beta1/commons.go
@@ -74,6 +74,11 @@ const (
 	// MemberClusterFinalizer is used to make sure that we handle gc of all the member cluster resources on the hub cluster.
 	MemberClusterFinalizer = FleetPrefix + "membercluster-finalizer"
 
+	// MemberClusterNameLabel is a label automatically added to MemberCluster objects
+	// with the value set to the MemberCluster's name. This enables selecting clusters
+	// by name in ResourceOverride and ClusterResourceOverride via labelSelector.
+	MemberClusterNameLabel = FleetPrefix + "member-cluster-name"
+
 	// WorkFinalizer is used by the work generator to make sure that the binding is not deleted until the work objects
 	// it generates are all deleted, or used by the work controller to make sure the work has been deleted in the member
 	// cluster.
--- a/apis/placement/v1beta1/commons.go
+++ b/apis/placement/v1beta1/commons.go
@@ -74,10 +74,10 @@ const (
 	// MemberClusterFinalizer is used to make sure that we handle gc of all the member cluster resources on the hub cluster.
 	MemberClusterFinalizer = FleetPrefix + "membercluster-finalizer"
 
-	// MemberClusterNameLabel is a label automatically added to MemberCluster objects
+	// MemberNameLabel is a label automatically added to MemberCluster objects
 	// with the value set to the MemberCluster's name. This enables selecting clusters
 	// by name in ResourceOverride and ClusterResourceOverride via labelSelector.
-	MemberClusterNameLabel = FleetPrefix + "member-cluster-name"
+	MemberNameLabel = FleetPrefix + "member-name"
 
 	// WorkFinalizer is used by the work generator to make sure that the binding is not deleted until the work objects
 	// it generates are all deleted, or used by the work controller to make sure the work has been deleted in the member
--- a/pkg/controllers/membercluster/v1beta1/membercluster_controller_test.go
+++ b/pkg/controllers/membercluster/v1beta1/membercluster_controller_test.go
@@ -64,7 +64,7 @@ const (
 	propertyProviderConditionMessage2 = "property provider condition 2 message"
 )
 
-func TestEnsureMemberClusterNameLabel(t *testing.T) {
+func TestEnsureMemberNameLabel(t *testing.T) {
 	tests := map[string]struct {
 		r             *Reconciler
 		memberCluster *clusterv1beta1.MemberCluster
@@ -81,12 +81,12 @@ func TestEnsureMemberClusterNameLabel(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "mc1",
 					Labels: map[string]string{
-						placementv1beta1.MemberClusterNameLabel: "mc1",
+						placementv1beta1.MemberNameLabel: "mc1",
 					},
 				},
 			},
 			wantLabels: map[string]string{
-				placementv1beta1.MemberClusterNameLabel: "mc1",
+				placementv1beta1.MemberNameLabel: "mc1",
 			},
 		},
 		"no labels at all": {
@@ -103,7 +103,7 @@ func TestEnsureMemberClusterNameLabel(t *testing.T) {
 				},
 			},
 			wantLabels: map[string]string{
-				placementv1beta1.MemberClusterNameLabel: "mc1",
+				placementv1beta1.MemberNameLabel: "mc1",
 			},
 		},
 		"labels exist but name label is missing": {
@@ -123,8 +123,8 @@ func TestEnsureMemberClusterNameLabel(t *testing.T) {
 				},
 			},
 			wantLabels: map[string]string{
-				"existing-label":                        "value",
-				placementv1beta1.MemberClusterNameLabel: "mc1",
+				"existing-label":                 "value",
+				placementv1beta1.MemberNameLabel: "mc1",
 			},
 		},
 		"label present with wrong value": {
@@ -139,12 +139,12 @@ func TestEnsureMemberClusterNameLabel(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "mc1",
 					Labels: map[string]string{
-						placementv1beta1.MemberClusterNameLabel: "wrong-name",
+						placementv1beta1.MemberNameLabel: "wrong-name",
 					},
 				},
 			},
 			wantLabels: map[string]string{
-				placementv1beta1.MemberClusterNameLabel: "mc1",
+				placementv1beta1.MemberNameLabel: "mc1",
 			},
 		},
 		"update error": {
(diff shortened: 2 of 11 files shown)


pkg/controllers/membercluster/v1beta1/membercluster_controller_test.go
```diff
@@ -63,6 +64,157 @@ const (
...
+				}
+			}
+		})
+	}
+}
+
+func TestReconcileEnsureMemberClusterNameLabelError(t *testing.T) {
+	// This test verifies that Reconcile returns an error when ensureMemberClusterNameLabel fails.
+	// The MC already has the finalizer (so ensureFinalizer is a no-op) but no label.
+	updateErr := errors.New("update failed")
```
weng271190436, line 191: Did you intend to use `updateErr` in `if err == nil || !strings.Contains(err.Error(), "update failed") {` but forgot to do it? Otherwise MockUpdate can just be

```go
			MockUpdate: func(ctx context.Context, obj client.Object, opts ...client.UpdateOption) error {
				return updateErr
			},
```

test/e2e/placement_cro_test.go
```diff
@@ -952,3 +954,136 @@ var _ = Context("creating clusterResourceOverride for a namespace-only CRP", Ord
...
+		By(fmt.Sprintf("deleting clusterResourceOverride %s", croName))
+		cleanupClusterResourceOverride(croName)
+	})
+
+	It("should update CRP status as expected", func() {
+		wantCRONames := []string{croSnapShotName}
+		crpKey := types.NamespacedName{Name: crpName}
+		// The CRO targets only one cluster via clusterSelector, so only the targeted
+		// cluster should have ApplicableClusterResourceOverrides set.
+		crpStatusUpdatedActual := func() error {
```
weng271190436: In actuals_test.go, can you create a helper for constructing crpStatusUpdatedActual?

```go
// crpStatusWithSingleClusterOverrideUpdatedActual is like crpStatusWithOverrideUpdatedActual but
// applies the override only to overrideCluster while other selected clusters receive no override.
func crpStatusWithSingleClusterOverrideUpdatedActual(
	wantSelectedResourceIdentifiers []placementv1beta1.ResourceIdentifier,
	wantSelectedClusters []string,
	wantObservedResourceIndex string,
	overrideCluster string,
	wantClusterResourceOverrides []string,
	wantResourceOverrides []placementv1beta1.NamespacedName) func() error {
	crpKey := types.NamespacedName{Name: fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())}
	croMap := map[string][]string{overrideCluster: wantClusterResourceOverrides}
	roMap := map[string][]placementv1beta1.NamespacedName{overrideCluster: wantResourceOverrides}
	return placementStatusWithOverrideUpdatedActual(crpKey, wantSelectedResourceIdentifiers, wantSelectedClusters,
		wantObservedResourceIndex, croMap, roMap)
}
```

And then change `placementStatusWithOverrideUpdatedActual` to consume 2 maps: one from cluster name to slice of strings that represent applicable CRO snapshots, one from cluster name to slice of placementv1beta1.NamespacedNames that represent applicable RO snapshots

like below

```go
func placementStatusWithOverrideUpdatedActual(
	placementKey types.NamespacedName,
	wantSelectedResourceIdentifiers []placementv1beta1.ResourceIdentifier,
	wantSelectedClusters []string,
	wantObservedResourceIndex string,
	wantClusterResourceOverrides map[string][]string,
	wantResourceOverrides map[string][]placementv1beta1.NamespacedName,
) func() error {
	return func() error {
		placement, err := retrievePlacement(placementKey)
		if err != nil {
			return err
		}
		hasOverride := len(wantClusterResourceOverrides) > 0 || len(wantResourceOverrides) > 0
		var wantPlacementStatus []placementv1beta1.PerClusterPlacementStatus
		for _, name := range wantSelectedClusters {
			perClusterHasOverride := len(wantClusterResourceOverrides[name]) > 0 || len(wantResourceOverrides[name]) > 0
			wantPlacementStatus = append(wantPlacementStatus, placementv1beta1.PerClusterPlacementStatus{
				ClusterName:                        name,
				Conditions:                         perClusterRolloutCompletedConditions(placement.GetGeneration(), true, perClusterHasOverride),
				ApplicableResourceOverrides:        wantResourceOverrides[name],
				ApplicableClusterResourceOverrides: wantClusterResourceOverrides[name],
				ObservedResourceIndex:              wantObservedResourceIndex,
			})
		}
```

And adapt crpStatusWithOverrideUpdatedActual and rpStatusWithOverrideUpdatedActual to use this updated placementStatusWithOverrideUpdatedActual

test/e2e/placement_cro_test.go
```diff
@@ -952,3 +954,136 @@ var _ = Context("creating clusterResourceOverride for a namespace-only CRP", Ord
...
+			return nil
+		}
+		Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP %s status as expected", crpName)
+	})
+
+	It("should place the selected resources on all member clusters", checkIfPlacedWorkResourcesOnAllMemberClusters)
+
+	It("should have override annotation only on the targeted cluster", func() {
+		wantAnnotations := map[string]string{croTestAnnotationKey: croTestAnnotationValue}
+		Eventually(func() error {
```
weng271190436: Do we need eventually here? If in the last step already verifies work resources were placed on all member clusters. Why eventually?

test/e2e/placement_cro_test.go
```diff
@@ -952,3 +954,136 @@ var _ = Context("creating clusterResourceOverride for a namespace-only CRP", Ord
...
+			return validateAnnotationOfConfigMapOnCluster(targetCluster, wantAnnotations)
+		}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to override configmap annotation on targeted cluster %s", targetCluster.ClusterName)
+	})
+
+	It("should not have override annotation on other clusters", func() {
+		for _, cluster := range allMemberClusters {
+			if cluster.ClusterName == targetCluster.ClusterName {
+				continue
+			}
+			Expect(validateNamespaceNoAnnotationOnCluster(cluster, croTestAnnotationKey)).Should(Succeed(),
```
weng271190436: This should be consistently to verify that the annotation is consistently not there rather than just "not there yet"?

apis/placement/v1beta1/commons.go
```diff
@@ -74,6 +74,11 @@ const (
 	// MemberClusterFinalizer is used to make sure that we handle gc of all the member cluster resources on the hub cluster.
 	MemberClusterFinalizer = FleetPrefix + "membercluster-finalizer"
 
+	// MemberClusterNameLabel is a label automatically added to MemberCluster objects
+	// with the value set to the MemberCluster's name. This enables selecting clusters
+	// by name in ResourceOverride and ClusterResourceOverride via labelSelector.
+	MemberClusterNameLabel = FleetPrefix + "member-cluster-name"
```
weng271190436: In Azure Kubernetes Fleet Manager, we have these 2 labels on the member cluster (there are more but these 2 are the most relevant ones)

```
fleet.azure.com/cluster-name=testmember-0
fleet.azure.com/member-name=testmember-0
```
fleet.azure.com/cluster-name is the AKS cluster name
fleet.azure.com/member-name is the fleet member name

I think this should be named 

```
kubernetes-fleet.io/member-name=testmember-0
```
, which follows the existing Azure label convention

weng271190436: @sjwaight I think having `kubernetes-fleet.io/member-name=testmember-0` in addition to 

```
fleet.azure.com/cluster-name=testmember-0
fleet.azure.com/member-name=testmember-0
```

is acceptable in Azure Kubernetes Fleet Manager's member cluster CRs?

pkg/controllers/membercluster/v1beta1/membercluster_controller_test.go
ytimocin: Yes, my bad. Addressed.

apis/placement/v1beta1/commons.go
sjwaight: I think in an ideal world we would have just the kubernetes-fleet.io labels and retire the fleet.azure.com ones for cluster name and member name, as these concepts are portable between KubeFleet and Fleet Manager. I would recommend we add both: kubernetes-fleet.io/member-name and kubernetes-fleet.io/cluster-name (if it's not already there). We can add these to our documentation for Fleet Manager, but keep the existing labels to not break the current API contract.

serbrech: `fleet.azure.com` labels are owned by the azure service, and the service is the source of truth for them (we set these labels). this ensures parity in label selectors between controlplane and dataplane.

The Kubefleet system should not care more than that about them.

Having built-in labels in kubefleet make sense. They should mirror the ones from the service, when they serve the same purpose. That's what the namespacing of labels is for :).

We will certainly have labels that only kubefleet can define in the future too, because it can get them from the members directly.

I think that we should also always consider built-in kubernetes labels. for example topology has a region and zone label that might be relevant?

https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion

weng271190436 (dismissed):

weng271190436 (approved):

PR #621, opened 2026-04-16
feat: determine failure type when possible for update run

britaniar:

Fixes #

I have:
- Included failure type label for "fleet_workload_update_run_status_last_timestamp_seconds" to help determine what failed update runs were due to user error. 
- Determination is based on if the error has the user error wrapper. 

- UT
- Integration

This is to help with alerts that are based on the number of failed update run. We want excluded those with known user error, so we don't have to investigate.

### DIFF
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -155,7 +155,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 			// errStagedUpdatedAborted cannot be retried.
 			if errors.Is(validateErr, errStagedUpdatedAborted) {
 				reconcileErr = validateErr
-				return runtime.Result{}, r.recordUpdateRunFailed(ctx, updateRun, validateErr)
+				return runtime.Result{}, r.recordUpdateRunFailed(ctx, updateRun, validateErr.Error())
 			}
 			return runtime.Result{}, validateErr
 		}
@@ -178,7 +178,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		if errors.Is(execErr, errStagedUpdatedAborted) {
 			// errStagedUpdatedAborted cannot be retried.
 			reconcileErr = execErr
-			return runtime.Result{}, r.recordUpdateRunFailed(ctx, updateRun, execErr)
+			return runtime.Result{}, r.recordUpdateRunFailed(ctx, updateRun, execErr.Error())
 		}
 
 		if finished {
@@ -194,7 +194,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		if errors.Is(stopErr, errStagedUpdatedAborted) {
 			// errStagedUpdatedAborted cannot be retried.
 			reconcileErr = stopErr
-			return runtime.Result{}, r.recordUpdateRunFailed(ctx, updateRun, stopErr)
+			return runtime.Result{}, r.recordUpdateRunFailed(ctx, updateRun, stopErr.Error())
 		}
 
 		if finished {
@@ -210,7 +210,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		klog.ErrorS(unexpectedErr, "Invalid updateRun state", "state", state, "updateRun", runObjRef)
 		// This is an internal error - unsupported state should not happen
 		reconcileErr = unexpectedErr
-		return runtime.Result{}, r.recordUpdateRunFailed(ctx, updateRun, unexpectedErr)
+		return runtime.Result{}, r.recordUpdateRunFailed(ctx, updateRun, unexpectedErr.Error())
 	}
 	return runtime.Result{}, nil
 }
@@ -297,7 +297,7 @@ func (r *Reconciler) recordUpdateRunSucceeded(ctx context.Context, updateRun pla
 }
 
 // recordUpdateRunFailed records the failed condition in the updateRun status.
-func (r *Reconciler) recordUpdateRunFailed(ctx context.Context, updateRun placementv1beta1.UpdateRunObj, err error) error {
+func (r *Reconciler) recordUpdateRunFailed(ctx context.Context, updateRun placementv1beta1.UpdateRunObj, message string) error {
 	updateRunStatus := updateRun.GetUpdateRunStatus()
 	meta.SetStatusCondition(&updateRunStatus.Conditions, metav1.Condition{
 		Type:               string(placementv1beta1.StagedUpdateRunConditionProgressing),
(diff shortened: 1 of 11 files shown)


pkg/controllers/updaterun/controller.go
```diff
@@ -109,8 +109,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		return runtime.Result{}, err
 	}
 
+	// Track errors for metrics emission. The error is used to determine the failure type
+	// (user_error vs internal_error) in the emitted metrics.
+	var reconcileErr error
 	// Emit the update run status metric based on status conditions in the updateRun.
-	defer emitUpdateRunStatusMetric(updateRun)
+	defer func() { emitUpdateRunStatusMetric(updateRun, reconcileErr) }()
```
Arvindthiru, line 117: Why do we have to wrap the emitUpdateRunStatusMetric within another inline function?

pkg/controllers/updaterun/metrics.go
```diff
@@ -35,30 +38,60 @@ func deleteUpdateRunMetrics(updateRun placementv1beta1.UpdateRunObj) {
...
+
+	// For False status, check if the error is a user error.
+	if err != nil && errors.Is(err, controller.ErrUserError) {
+		return hubmetrics.UpdateRunFailureTypeUserError
+	}
+
+	// Failed condition that is not a user error is an internal error.
+	if err != nil {
+		return hubmetrics.UpdateRunFailureTypeInternalError
+	}
```
Arvindthiru:
```suggestion
	if err != nil {
		// For False status, check if the error is a user error.
	    if errors.Is(err, controller.ErrUserError) {
	        return hubmetrics.UpdateRunFailureTypeUserError
	    }
	    // Failed condition that is not a user error is an internal error.
		return hubmetrics.UpdateRunFailureTypeInternalError
	}
```

pkg/controllers/updaterun/metrics.go
```diff
@@ -35,30 +38,60 @@ func deleteUpdateRunMetrics(updateRun placementv1beta1.UpdateRunObj) {
...
 
+// determineFailureType determines the type of failure based on the condition status and error.
+// It returns:
+// - "none" for successful, in-progress, waiting, or stopping conditions
+// - "user_error" for known customer configuration errors (when err wraps controller.ErrUserError)
+// - "internal_error" for errors that require investigation
+func determineFailureType(cond *metav1.Condition, generation int64, err error) hubmetrics.UpdateRunFailureType {
+	klog.InfoS("Condition:", "type", cond.Type, "status", cond.Status, "reason", cond.Reason, "message", cond.Message)
+	// If the condition status is True (success) or Unknown (in-progress), there's no failure.
+	if !condition.IsConditionStatusFalse(cond, generation) {
```
Arvindthiru: Might be better to first check to see if the err is not nil and then check the positive case just to be on the safer side?

pkg/controllers/updaterun/controller.go
```diff
@@ -288,7 +297,7 @@ func (r *Reconciler) recordUpdateRunSucceeded(ctx context.Context, updateRun pla
 }
 
 // recordUpdateRunFailed records the failed condition in the updateRun status.
-func (r *Reconciler) recordUpdateRunFailed(ctx context.Context, updateRun placementv1beta1.UpdateRunObj, message string) error {
+func (r *Reconciler) recordUpdateRunFailed(ctx context.Context, updateRun placementv1beta1.UpdateRunObj, err error) error {
```
Arvindthiru: Why are we updating this function signature ?

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -517,6 +522,7 @@ var _ = Describe("Updaterun initialization tests", func() {
 
 			By("Validating the initialization not failed due to no selected cluster")
```
Arvindthiru: This needs to be fixed, the check below is confusing

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -87,6 +89,7 @@ var _ = Describe("Updaterun initialization tests", func() {
 		stageUpdatingWaitTime = time.Second * 3
 		clusterUpdatingWaitTime = time.Second * 2
 
+		failureType = hubmetrics.UpdateRunFailureTypeInternalError // All following tests will still fail due to not fully set up environment.
```
Arvindthiru: Are we testing negative cases ?. 

"All following tests will still fail due to not fully set up environment." The comment makes it seems like the tests would fail to run. What does "not fully set environment" entail please add more context

pkg/controllers/updaterun/metrics_test.go
```diff
@@ -0,0 +1,123 @@
...
+	hubmetrics "github.com/kubefleet-dev/kubefleet/pkg/metrics/hub"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/controller"
+)
+
+func TestDetermineFailureType(t *testing.T) {
+	tests := []struct {
+		name       string
+		cond       *metav1.Condition
+		generation int64
```
Arvindthiru: nit: do we need generation as a separate test input since it's always 1 ?

pkg/controllers/updaterun/controller.go
britaniar: This ensures the closure reads the current value of reconcileErr when it actually executes (after the function returns), capturing any updates made during reconciliation.  Similar to https://github.com/kubefleet-dev/kubefleet/blob/aa840c6c4ecd18f0e3ce9c674d2d8455678000a7/pkg/controllers/updaterun/execution.go#L68-L71

britaniar: we reassign the variable (reconcileErr = initErr), so the defer doesn't see it without a closure. While for updaterun, the  interface value is copied, but it still points to the same underlying *ClusterStagedUpdateRun struct .

pkg/controllers/updaterun/metrics.go
```diff
@@ -35,30 +37,53 @@ func deleteUpdateRunMetrics(updateRun placementv1beta1.UpdateRunObj) {
...
-func emitUpdateRunStatusMetric(updateRun placementv1beta1.UpdateRunObj) {
+// The err parameter is used to determine the failure type for failed conditions.
+func emitUpdateRunStatusMetric(updateRun placementv1beta1.UpdateRunObj, err error) {
 	generation := updateRun.GetGeneration()
 	state := updateRun.GetUpdateRunSpec().State
 
 	updateRunStatus := updateRun.GetUpdateRunStatus()
 	succeedCond := meta.FindStatusCondition(updateRunStatus.Conditions, string(placementv1beta1.StagedUpdateRunConditionSucceeded))
 	if succeedCond != nil && succeedCond.ObservedGeneration == generation {
+		failureType := determineFailureType(err)
```
jwtty: This line can be moved out? It's shared by all options.

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -87,6 +89,10 @@ var _ = Describe("Updaterun initialization tests", func() {
 		stageUpdatingWaitTime = time.Second * 3
 		clusterUpdatingWaitTime = time.Second * 2
 
+		// All following tests will report internal errors, if not user error, because only the updaterun controller
+		// is running in this test environment. Other controllers (e.g., rollout, work generator,
+		// binding controllers) are not set up, causing operations that depend on them to fail.
+		failureType = hubmetrics.UpdateRunFailureTypeInternalError
```
jwtty, line 95: Do we not have any passed cases?

jwtty (approved):

Issue #216, opened 2025-08-21
Improve docs on use of TLS, Azure cloud config, managed identity in Helm charts


tillig:
The [member-agent Helm chart supports `config.azureCloudConfig`](https://github.com/kubefleet-dev/kubefleet/blob/f6245daf06c0b43f5f852ece38a69b1447272234/charts/member-agent/values.yaml#L30) which takes a variety of parameters that appear to help you work in an AKS environment with concepts like managed identity.

Unfortunately, there's not a "real working example" in the main kubefleet.dev docs, and the README for the chart is a bit short/incomplete.

For example...

- None of the `config` values are actually documented. `config.hubCA`, for example - what format does that take?
- When setting up the `config.azureCloudConfig` section, are all of those values (resource group, etc.) for the _member cluster_ you're setting up, or for the _hub cluster_? Or some mix-and-match?
- What rights do I need to give a managed identity to make this work? Do I need to grant something to the member cluster's node pool identity? What?

It makes setting up a secure version of this (validated TLS, managed identity, etc.) really difficult.

sjwaight:
Related to #296

PR #622, opened 2026-04-17
fix: Member uses TLS to communicate with hub by default

weng271190436:

Member uses TLS to communicate with hub by default. Also update scripts/docs everywhere.

With TLS being the default, we need to set hubCA everywhere

related to #216

I have:

### DIFF
--- a/test/upgrade/setup.sh
+++ b/test/upgrade/setup.sh
@@ -111,6 +111,9 @@ done
 kind export kubeconfig --name $HUB_CLUSTER
 HUB_SERVER_URL="https://$(docker inspect $HUB_CLUSTER-control-plane --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}'):6443"
 
+# Extract the hub cluster CA for secure TLS verification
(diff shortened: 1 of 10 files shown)


test/upgrade/setup.sh
```diff
@@ -111,6 +111,9 @@ done
 kind export kubeconfig --name $HUB_CLUSTER
 HUB_SERVER_URL="https://$(docker inspect $HUB_CLUSTER-control-plane --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}'):6443"
 
+# Extract the hub cluster CA for secure TLS verification
+HUB_CA=$(kubectl config view --raw -o jsonpath='{.clusters[?(@.name=="kind-'$HUB_CLUSTER'")].cluster.certificate-authority-data}')
```
ytimocin, line 115: Can this return empty at any point?

weng271190436: If you look at this line only - yes it could return empty

But on line 42 the hub cluster was created using kind with name $HUB_CLUSTER, I don't think it would happen in practice.

```
kind create cluster --name $HUB_CLUSTER --image=$KIND_IMAGE --kubeconfig=$KUBECONFIG
```

ytimocin (approved):

PR #665, opened 2026-04-24
fix: deterministic envelope Work names to prevent duplicates

### LINKED ISSUE #630: [BUG] Duplicate envelope Works leave the binding stuck in SyncWorkFailed forever
### **Describe the bug**

If two `Work` objects ever exist for the same `(binding, envelope)` in a member cluster's reserved namespace, the workgenerator detects it, declares `UnexpectedBehaviorError`, and **gives up**. Every subsequent reconcile finds the same two Works and fails identically. The binding stays `WorkSynchronized: False, Reason: SyncWorkFailed` forever and the placement never converges on that cluster until an operator manually deletes one of the Works.

**Where in the code:** `pkg/controllers/workgenerator/envelope.go:88-102` — the `len(workList.Items) > 1` branch returns an error without deduplicating.

**How the duplicates happen in the wild:** the envelope Work is named with a fresh `uuid.NewUUID()` on every create (`envelope.go:211`). Two back-to-back reconciles of the same binding can both see an empty `List` from the informer cache (cache hasn't propagated the first Work yet), both build a Work with different UUIDs, and both `Create` succeed — the API server enforces uniqueness by name, not by `(binding, envelope)`, so nothing stops it.

### **Environment**

Observed in Kind-based e2e (`make e2e-tests`, 3 members) on GitHub Actions CI, but the condition — informer cache lag between successive reconciles — is a normal controller-runtime behavior and can surface in any environment.

#### Organic occurrences (CI)

- https://github.com/kubefleet-dev/kubefleet/actions/runs/23621818959 (Mar 26) — `kind-cluster-2`; artifact expired, CI-log fingerprint matches.
- https://github.com/kubefleet-dev/kubefleet/actions/runs/24492820275 (Apr 16) — `kind-cluster-1`; artifact expired, fingerprint matches.
- https://github.com/kubefleet-dev/kubefleet/actions/runs/24839343153 (Apr 23) — `kind-cluster-2`; **hub-agent log in `e2e-logs-default` artifact shows the exact two duplicate Works (`...65ad9bba...` and `...f80ad44b...`), the `envelope.go:101` "Duplicate work objects found" error, and the full stack trace.**

#### Deterministic reproduction (local)

The organic race is timing-dependent, but the *state* it produces can be reproduced by hand. The controller can't tell the difference — it only sees "two Works match these labels". Steps against a running hub + member setup:

1. Apply a CRP that selects a namespace containing a `ResourceEnvelope` (any envelope-selecting CRP).
2. Wait for the envelope `Work` to appear on a member cluster's reserved namespace (e.g. `fleet-member-kind-cluster-1`):
   ```bash
   kubectl --context kind-hub get work -n fleet-member-kind-cluster-1 \
     -l kubernetes-fleet.io/envelope-work
   ```
3. Clone that `Work` under a new name, preserving all labels:
   ```bash
   ORIG=<name-from-step-2>
   kubectl --context kind-hub get work -n fleet-member-kind-cluster-1 $ORIG -o json | \
     jq --arg new "$ORIG-dup" '.metadata.name=$new |
       del(.metadata.uid, .metadata.resourceVersion, .metadata.creationTimestamp,
           .metadata.generation, .metadata.managedFields, .metadata.ownerReferences, .status)' | \
     kubectl --context kind-hub create -f -
   ```
4. Touch the envelope to trigger a binding reconcile:
   ```bash
   kubectl --context kind-hub annotate resourceenvelope -n <ns> <envelope> kick=$(date +%s) --overwrite
   ```
5. Observe the binding flip within seconds:
   ```bash
   kubectl --context kind-hub get clusterresourcebinding <binding> -o jsonpath='{.status.conditions[?(@.type=="WorkSynchronized")]}'
   ```
   Output:
   ```
   "reason":"SyncWorkFailed",
   "message":"Failed to synchronize the work to the latest: 2 work objects found for the same envelope <ns>/<envelope>, only one expected"
   ```

Message format — byte-identical between CI and local repro:
```
Failed to synchronize the work to the latest: 2 work objects found for the same envelope <namespace>/<name>, only one expected
```

Waiting 60+ seconds confirms the reconciler keeps firing (`observedGeneration` ticks up) but never self-heals — both duplicates remain, binding stays stuck.

### **Expected behavior**

- Duplicate envelope Works should not be creatable in the first place — Work identity should be a deterministic function of `(binding, envelope)` so concurrent creates collide at the API server.
- If duplicates do exist, the controller should deduplicate and continue, not permanently fail.

### **Screenshots**

N/A.

### **Additional context**

**Proposed fixes (can be combined):**

1. **Deterministic Work name** — replace `uuid.NewUUID()` at `envelope.go:211` with a hash of `(binding name, envelope type, envelope namespace, envelope name)`. Concurrent creates then collide on name, the API server rejects the second with `AlreadyExists`, and the controller re-lists and falls into the existing `len == 1` update branch (`envelope.go:103-110`). No code parses the UUID portion of the name — identity is carried by labels — so changing the naming scheme is safe.
2. **Self-heal on detection** — in the `len > 1` branch at `envelope.go:88-102`, pick a winner (e.g. oldest by `CreationTimestamp`) and delete the rest, then return a retryable error. Any deployment already stuck on this bug recovers on the next reconcile.

**Broader audit:** any controller path that does "list-by-label → create-with-random-name" without an idempotent name is susceptible to the same cache-staleness race. Worth scanning for other instances.

**Related:** the test-side kube-root-ca.crt flake is tracked separately in #629.

ytimocin:

Envelope Work names were generated with a fresh `uuid.NewUUID()` on every Create, so two back-to-back reconciles of the same binding with a stale informer-cache List could each build a Work with different names and both succeed. The `len > 1` branch then returned `UnexpectedBehaviorError` without any recovery, leaving the binding permanently stuck in `WorkSynchronized=False/SyncWorkFailed` and the placement never converging on that cluster.

`buildNewWorkForEnvelopeCR` now derives the name suffix from a truncated SHA-256 hash of `(envelope type, namespace, name)`. Two concurrent Creates for the same `(binding, envelope)` produce the same name and collide at the API server; the existing `AlreadyExists` handling in `upsertWork` requeues into the `len == 1` update branch.

When the `len > 1` state is still observed (possible only in environments that were stuck before this change), the controller now emits a Warning Event on the binding (reason=`DuplicateEnvelopeWorks`) and a clear log message instructing the operator to delete all but the oldest Work in the affected namespace. See **Special notes** for why auto-deletion was deliberately avoided.

Fixes #630

I have:

**Unit tests** (`pkg/controllers/workgenerator/envelope_test.go`):
- `TestEnvelopeWorkNameSuffix` — determinism, cross-identity uniqueness (different name, different namespace, different type all hash differently), fixed length (16 hex chars), lowercase DNS-1123-safe format.
- `TestCreateOrUpdateEnvelopeCRWorkObj_DuplicateWorksSurfaceWithoutMutation` — when multiple Works match the label query, the function returns `ErrUnexpectedBehavior`, **no Work is deleted**, and a `DuplicateEnvelopeWorks` Event is emitted.
- Existing `TestCreateOrUpdateEnvelopeCRWorkObj` table case updated: "two existing works surface an UnexpectedBehaviorError without mutation".
- All existing tests in the workgenerator package still pass; lint (`make lint`) clean.

**Live-cluster reproduction** (Kind hub + 3 members, fresh build of hub-agent with this change):
1. Created a CRP selecting a namespace containing a `ResourceEnvelope`.
2. Verified all 3 member-cluster namespaces got the same deterministic Work name suffix (`ad22e36a8e622bff`), derived from the envelope identity — confirming Fix 1.
3. Injected a duplicate Work by cloning the real one with a new name but identical labels (simulates the state an environment stuck on the race would be in). Confirmed:
   - Binding transitions to `WorkSynchronized: False, Reason: SyncWorkFailed` with the byte-identical message `"Failed to synchronize the work to the latest: 2 work objects found for the same envelope <ns>/<envelope>, only one expected"`.
   - **Both Works remain** — no auto-delete.
   - `DuplicateEnvelopeWorks` Warning Event is present on the binding with cleanup instructions.
   - Hub-agent log contains `"Duplicate envelope Work objects found; manual cleanup required"`.
4. Manually deleted the extra Work (simulating operator intervention). Binding recovered to `AllWorkSynced` within ~8 seconds.

**Edge case — same envelope name, different namespaces:** two CRPs, each selecting a namespace containing a `ResourceEnvelope` named `shared-name`. The generated Work names differ (`8671af501951ff8c` vs. `f430c5e4da73cda1`), because the namespace is part of the hash input.

**Soak — 40-minute envelope-focused `ginkgo --until-it-fails -p` loop** against the fixed build:
- 37 clean iterations, 740 specs total, 0 failures.
- Duplicate-Work watcher (`kubectl get works -l envelope-work` polled every 3s) fired zero times.
- No `SyncWorkFailed`, no `Duplicate work objects`, no `manual cleanup required` log entries.
- For reference: on unpatched `main`, this CI path flakes at ~7.5% per run. Probability of 37 clean iterations if the bug still existed: `(1 − 0.075)^37 ≈ 5.3%`.

**Why not auto-delete duplicates?** An earlier draft of this change deduplicated on detection (kept the oldest by `CreationTimestamp`, deleted the rest). That was withdrawn after feedback from @michaelawyu: once created, Works can be reconciled by the member agent and lead to resource creation; deleting a Work that the member agent has already applied triggers resource removal on the member side and causes user-visible fluctuation. The current change detects and surfaces but does not mutate. Pre-existing stuck deployments require one-time manual cleanup (delete all but the oldest Work matching the `parent-resource-binding` + `envelope-name` labels in the affected `fleet-member-<cluster>` namespace).

**Relationship to broader work:** @michaelawyu mentioned an adjacent task to "remove the scheduler raw list" covering a similar read/write consistency pattern, and flagged resource-version-based consistent views as the longer-term direction. This change takes a compatible but different angle — pushing the `one Work per (binding, envelope)` invariant into the API server's name-uniqueness check via deterministic naming. It can coexist with a future resource-version-based lookup and does not preclude it.

**Upgrade behavior:** existing envelope Works with UUID-based names are unaffected. On the first reconcile after upgrade, the `len == 1` branch at `envelope.go:103-110` updates the existing Work in place via `refreshWorkForEnvelopeCR` — no rename, no churn. New Works (for new envelopes, new CRPs, or after an envelope is deleted and recreated) get the deterministic name. Clusters end up with a mix, both work correctly; over time, UUID-named Works age out naturally. No migration step required.

**Hash format:** truncated to 8 bytes (16 hex characters) of SHA-256 over `"<type>/<namespace>/<name>"`. Collision probability per binding's envelope set is ~2^-64. No code in the codebase parses the suffix portion of the Work name — identity is carried by labels — so the naming change is safe.

### DIFF
--- a/pkg/controllers/workgenerator/envelope.go
+++ b/pkg/controllers/workgenerator/envelope.go
@@ -215,9 +215,9 @@ func refreshWorkForEnvelopeCR(
 // derived from the envelope's identity. Using a deterministic suffix (instead of a random
 // UUID) means two concurrent Create attempts for the same envelope collide on name at the
 // API server, which enforces the "one Work per (binding, envelope)" invariant server-side.
-// The hash is truncated to 16 hex characters (8 bytes of SHA-256), keeping the overall
-// Work name well under the DNS-1123 subdomain limit; collision probability at the scale
-// of a single binding's envelope set is ~2^-64.
+// The hash is truncated to 8 bytes of SHA-256 (rendered as 16 hex characters), keeping
+// the overall Work name well under the DNS-1123 subdomain limit; collision probability
+// at the scale of a single binding's envelope set is ~2^-64.
 func envelopeWorkNameSuffix(envelopeReader fleetv1beta1.EnvelopeReader) string {
 	// NOTE: ClusterResourceEnvelope has an empty namespace; ResourceEnvelope has a namespace.
 	// Including the type disambiguates an unlikely name collision across kinds.
--- a/pkg/controllers/workgenerator/envelope.go
+++ b/pkg/controllers/workgenerator/envelope.go
@@ -18,13 +18,15 @@ package workgenerator
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"sort"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -87,18 +89,27 @@ func (r *Reconciler) createOrUpdateEnvelopeCRWorkObj(
 	var work *fleetv1beta1.Work
 	switch {
 	case len(workList.Items) > 1:
-		// Multiple matching work objects found; this should never occur under normal conditions.
-		wrappedErr := fmt.Errorf("%d work objects found for the same envelope %v, only one expected", len(workList.Items), envelopeReader.GetEnvelopeObjRef())
-		klog.ErrorS(wrappedErr, "Failed to create or update work object for envelope",
-			"resourceBinding", klog.KObj(binding),
-			"resourceSnapshot", klog.KObj(resourceSnapshot),
-			"envelope", envelopeReader.GetEnvelopeObjRef())
-		// Log the work object names to help debug.
+		// Multiple matching work objects found. This can only happen in environments that
+		// were stuck on this state *before* deterministic Work naming was introduced in
+		// buildNewWorkForEnvelopeCR; newly-raced reconciles are now prevented by the
+		// API server's name-uniqueness check. We deliberately do NOT auto-delete the
+		// extras: both Works may have been applied by the member agent, and deleting one
+		// would trigger member-side resource cleanup and user-visible fluctuation on
+		// resources the survivor still owns. Surface the state for operator cleanup
+		// (manually delete all but the oldest Work in the affected namespace).
 		workNames := make([]string, len(workList.Items))
 		for i := range workList.Items {
 			workNames[i] = workList.Items[i].Name
 		}
-		klog.ErrorS(wrappedErr, "Duplicate work objects found", "works", workNames)
+		wrappedErr := fmt.Errorf("%d work objects found for the same envelope %v, only one expected", len(workList.Items), envelopeReader.GetEnvelopeObjRef())
+		klog.ErrorS(wrappedErr, "Duplicate envelope Work objects found; manual cleanup required — delete all but the oldest Work in the namespace",
+			"works", workNames,
+			"resourceBinding", klog.KObj(binding),
+			"resourceSnapshot", klog.KObj(resourceSnapshot),
+			"envelope", envelopeReader.GetEnvelopeObjRef())
+		r.recorder.Eventf(binding, corev1.EventTypeWarning, "DuplicateEnvelopeWorks",
+			"Multiple Work objects (%v) found for envelope %v in namespace %s; delete all but the oldest to recover",
+			workNames, envelopeReader.GetEnvelopeObjRef(), fmt.Sprintf(utils.NamespaceNameFormat, binding.GetBindingSpec().TargetCluster))
 		return nil, controller.NewUnexpectedBehaviorError(wrappedErr)
 	case len(workList.Items) == 1:
 		klog.V(2).InfoS("Found existing work object for the envelope; updating it",
@@ -200,6 +211,24 @@ func refreshWorkForEnvelopeCR(
 	work.Spec.ApplyStrategy = resourceBinding.GetBindingSpec().ApplyStrategy
 }
 
+// envelopeWorkNameSuffix returns a deterministic suffix for the envelope Work's name,
+// derived from the envelope's identity. Using a deterministic suffix (instead of a random
+// UUID) means two concurrent Create attempts for the same envelope collide on name at the
+// API server, which enforces the "one Work per (binding, envelope)" invariant server-side.
+// The hash is truncated to 16 hex characters (8 bytes of SHA-256), keeping the overall
+// Work name well under the DNS-1123 subdomain limit; collision probability at the scale
+// of a single binding's envelope set is ~2^-64.
+func envelopeWorkNameSuffix(envelopeReader fleetv1beta1.EnvelopeReader) string {
+	// NOTE: ClusterResourceEnvelope has an empty namespace; ResourceEnvelope has a namespace.
+	// Including the type disambiguates an unlikely name collision across kinds.
+	identity := fmt.Sprintf("%s/%s/%s",
+		envelopeReader.GetEnvelopeType(),
+		envelopeReader.GetNamespace(),
+		envelopeReader.GetName())
+	sum := sha256.Sum256([]byte(identity))
+	return hex.EncodeToString(sum[:8])
+}
+
 func buildNewWorkForEnvelopeCR(
 	workNamePrefix string,
 	resourceBinding fleetv1beta1.BindingObj,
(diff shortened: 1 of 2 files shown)


pkg/controllers/workgenerator/envelope.go
```diff
@@ -200,6 +211,24 @@ func refreshWorkForEnvelopeCR(
...
+// of a single binding's envelope set is ~2^-64.
+func envelopeWorkNameSuffix(envelopeReader fleetv1beta1.EnvelopeReader) string {
+	// NOTE: ClusterResourceEnvelope has an empty namespace; ResourceEnvelope has a namespace.
+	// Including the type disambiguates an unlikely name collision across kinds.
+	identity := fmt.Sprintf("%s/%s/%s",
+		envelopeReader.GetEnvelopeType(),
+		envelopeReader.GetNamespace(),
+		envelopeReader.GetName())
+	sum := sha256.Sum256([]byte(identity))
+	return hex.EncodeToString(sum[:8])
```
michaelawyu, line 229: Hi Yetkin! Thanks for this 🙏 I think this looks good, just a few nitpicky stuff -> but let's address this later, cause this improvement is better than what we have.

michaelawyu: A nit would be that when we use the just the first 8 digits that chances of collision are actually not that insignificant (birthday paradox, around 1/80000; of course using 16 digits would be much better) -> but given our usage so far (esp. the envelope feature) this is surely good enough.

pkg/controllers/workgenerator/envelope_test.go
```diff
@@ -725,3 +735,168 @@ func createResourceContent(t *testing.T, obj runtime.Object) *fleetv1beta1.Resou
 		},
 	}
 }
+
+// TestEnvelopeWorkNameSuffix verifies that the suffix is a stable, length-bounded function
+// of the envelope's identity. Determinism is what lets the API server enforce the
+// "one Work per (binding, envelope)" invariant via name uniqueness.
+func TestEnvelopeWorkNameSuffix(t *testing.T) {
```
michaelawyu, line 741: Hi Yetkin! Just a style nit: maybe it's better to use the consistent table-form test cases here?

pkg/controllers/workgenerator/envelope_test.go
```diff
@@ -725,3 +735,168 @@ func createResourceContent(t *testing.T, obj runtime.Object) *fleetv1beta1.Resou
...
+	// Post-condition: all duplicates remain untouched — no auto-deletion.
+	remaining := &fleetv1beta1.WorkList{}
+	if err := fakeClient.List(ctx, remaining, client.InNamespace(workNamespace)); err != nil {
+		t.Fatalf("List Works in %q: %v", workNamespace, err)
+	}
+	if len(remaining.Items) != len(dupNames) {
+		t.Errorf("remaining Works = %d (%v), want %d (no mutation expected)",
+			len(remaining.Items), workNames(remaining.Items), len(dupNames))
+	}
+	for _, n := range dupNames {
```
michaelawyu: Hi Yetkin! A nit: the works have been listed; the subsequent get seems a bit redundant? For fake client we can just directly do cmp.Diff.

michaelawyu (dismissed):
LGTM ;) Just some minor comments.

michaelawyu (approved):
LGTM ;)

PR #667, opened 2026-04-24
refactor: extract IsLatest/LookupLatest policy snapshot helpers

### LINKED ISSUE #643: Scheduler: extract shared helper for latest policy snapshot lookup
## What
`lookupLatestPolicySnapshot` and the "is this the latest snapshot?" check are duplicated between the scheduler and the policy-snapshot watcher. Extract into a shared library.

## Locations
- `pkg/scheduler/scheduler.go:350`
- `pkg/scheduler/watchers/schedulingpolicysnapshot/watcher.go:78`

ytimocin:

Fixes #643.

Moves the two duplicated "latest policy snapshot" lookups into `pkg/utils/controller/policy_snapshot_resolver.go` so the scheduler and the policy-snapshot watcher share the implementation.

Two new helpers:

- `IsLatestPolicySnapshot(snapshot) (bool, error)` — inspects the `IsLatestSnapshotLabel`. Returns `(false, ErrUnexpectedBehavior-wrapped)` for missing or non-bool values. The watcher's `Reconcile` now uses this instead of inline label parsing.
- `LookupLatestPolicySnapshot(ctx, client.Reader, placement) (PolicySnapshotObj, error)` — wraps `FetchLatestPolicySnapshot` and enforces the "exactly one latest snapshot" invariant. Returns a plain error for the (transient) zero case and `ErrUnexpectedBehavior` for the >1 case. The scheduler's `lookupLatestPolicySnapshot` now delegates to it.

The watcher's `buildCustomPredicate` continues to do its own `ParseBool` because it compares old-vs-new label transitions, which doesn't fit the helper's signature. Two now-unused imports (`strconv`, `labels`) are dropped from `scheduler.go`.

I have:

- New table-driven `TestIsLatestPolicySnapshot` covers label true/false, missing label, nil labels map, malformed value.
- New table-driven `TestLookupLatestPolicySnapshot` covers exactly-one cluster-scoped, exactly-one namespaced, zero snapshots (plain error), multiple snapshots (`ErrUnexpectedBehavior`), ignoring snapshots from other placements, and ignoring non-latest snapshots.
- `go test -count=1 -short ./pkg/scheduler/... ./pkg/utils/controller/...` passes locally with envtest. `make reviewable` is clean.

- `LookupLatestPolicySnapshot` uses `client.Reader` to match `FetchLatestPolicySnapshot`'s existing signature.
- The API-error path now logs three times (once inside `FetchLatestPolicySnapshot`, once inside `NewAPIServerError`, once at the scheduler call site). The duplication is acknowledged in a comment on `LookupLatestPolicySnapshot`. Pre-existing behavior was two logs; the third is the helper-level wrap that lets all callers get consistent classification. Happy to drop the scheduler-side log if reviewers prefer.
- The "no latest snapshot" error string is no longer logged inside the helper; the scheduler caller still logs it, so observability for the (most common) zero-snapshot transient case is preserved.

### DIFF
--- a/pkg/utils/controller/policy_snapshot_resolver.go
+++ b/pkg/utils/controller/policy_snapshot_resolver.go
@@ -130,63 +130,49 @@ func IsLatestPolicySnapshot(snapshot fleetv1beta1.PolicySnapshotObj) (bool, erro
 }
 
 // LookupLatestPolicySnapshot returns the single active (latest) policy snapshot associated with the
-// given placement. It is a thin wrapper over FetchLatestPolicySnapshot that enforces the
-// "exactly one latest snapshot" invariant.
+// placement identified by placementKey, enforcing the "exactly one latest snapshot" invariant.
+// A namespaced placementKey selects SchedulingPolicySnapshot; a cluster-scoped key (empty
+// namespace) selects ClusterSchedulingPolicySnapshot.
 //
 // Returns:
 //   - (snapshot, nil) when exactly one latest snapshot exists.
-//   - (nil, error) when zero, more than one, or a List error occurs. The "zero" case may happen
-//     transiently when a placement is newly created or its latest snapshot is being replaced;
-//     callers typically log and let the next event re-trigger reconciliation.
-//
-// The API-error path is logged inside both FetchLatestPolicySnapshot and NewAPIServerError; the
-// duplication is preserved here so callers that wrap the error get consistent classification, at
-// the cost of one extra log line on List failures.
-func LookupLatestPolicySnapshot(ctx context.Context, k8Client client.Reader, placement fleetv1beta1.PlacementObj) (fleetv1beta1.PolicySnapshotObj, error) {
-	placementKey := types.NamespacedName{Namespace: placement.GetNamespace(), Name: placement.GetName()}
-	policySnapshotList, err := FetchLatestPolicySnapshot(ctx, k8Client, placementKey)
-	if err != nil {
+//   - (nil, ErrNoLatestPolicySnapshot-wrapped error) when none exist. This is a transient state for
+//     newly-created placements or in-progress rotations; callers branch on it via errors.Is.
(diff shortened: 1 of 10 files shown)


pkg/utils/controller/policy_snapshot_resolver.go
```diff
@@ -110,6 +110,55 @@ func BuildPolicySnapshot(placementObj fleetv1beta1.PlacementObj, policySnapshotI
...
+//   - (nil, error) when zero, more than one, or a List error occurs. The "zero" case may happen
+//     transiently when a placement is newly created or its latest snapshot is being replaced;
+//     callers typically log and let the next event re-trigger reconciliation.
+//
+// The API-error path is logged inside both FetchLatestPolicySnapshot and NewAPIServerError; the
+// duplication is preserved here so callers that wrap the error get consistent classification, at
+// the cost of one extra log line on List failures.
+func LookupLatestPolicySnapshot(ctx context.Context, k8Client client.Reader, placement fleetv1beta1.PlacementObj) (fleetv1beta1.PolicySnapshotObj, error) {
+	placementKey := types.NamespacedName{Namespace: placement.GetNamespace(), Name: placement.GetName()}
+	policySnapshotList, err := FetchLatestPolicySnapshot(ctx, k8Client, placementKey)
```
weng271190436: FetchLatestPolicySnapshot should use IsLatestPolicySnapshot?

weng271190436: Actually it doesn't make sense to have both FetchLatestPolicySnapshot and LookupLatestPolicySnapshot?

I checked the 2 use cases of FetchLatestPolicySnapshot. They both fetch, and does the case 0, 1, default conditional just like LookupLatestPolicySnapshot anyway. Maybe should just keep `LookupLatestPolicySnapshot `

weng271190436 (approved):

PR #668, opened 2026-04-24
feat: harden property selector validation

### LINKED ISSUE #646: Harden placement selector validation
## What
Several validation gaps exist in the placement selector validator:
- Enforce mutual exclusiveness between `name` and `labelSelector`.
- Enforce validity of the `labelSelector` itself.
- Check for logical contradictions in property selectors (e.g. `x > 5 AND x < 3`).
- Restructure `Eq` validation so operator and value logic live together.

## Locations
- `pkg/utils/controller/resource_selector_resolver.go:526`
- `pkg/utils/controller/resource_selector_resolver.go:556`
- `pkg/utils/validator/placement.go:463`
- `pkg/utils/validator/placement.go:534`

ytimocin:

Fixes #646.

Addresses the four validator-hardening TODOs from the issue. Two were stale (the validator already enforced them); two are real and now have implementations + thorough tests.

**Real changes**

- Bundle operator and value validation into `validateOperatorAndValues` driven by a single `supportedPropertyOperators` table. Replaces the prior split `validateOperator` / `validateValues`. Preserves the historical "operator X requires exactly one value, got N" wording so existing tests stay stable.
- Add per-property logical-contradiction detection (`validateRequirementsConsistency`). Split into named helpers per reviewer feedback: `collectRequirementBounds`, `tightenLower`/`tightenUpper` (with strict-preferred tiebreak), `checkEqVsNe`, `checkInterval`, `checkEqInsideBounds`. Detects:
  - two `Eq` with different values,
  - `Eq` + `Ne` on the same value,
  - empty intervals from the most-restrictive lower vs the most-restrictive upper (incl. boundary cases `Gt x + Lt x`, `Gt x + Lte x`, `Gte x + Lt x`),
  - `Eq` value violating the most-restrictive lower or upper bound.

  More elaborate cases (sparse `Ne` exclusion sets, complex inequality normalisation) are intentionally left for the property-selector evaluator at scheduling time.

**Stale TODOs removed**

- `resource_selector_resolver.go:526` — mutual exclusiveness of `Name` vs `LabelSelector` is already enforced in `pkg/utils/validator/placement.go:92-95` (`validatePlacement`). Replaced with a pointer-comment.
- `resource_selector_resolver.go:556` — `labelSelector` validity is already enforced in `validateLabelSelector` (`placement.go:394`). Replaced with a pointer-comment.

I have:

- `TestValidateOperatorAndValues` (7 cases including unsupported operator, malformed quantity, count mismatch).
- `TestValidateRequirementsConsistency` (~20 cases: each contradiction variant plus negative controls and malformed-input skipping).
- `TestRequirementBoundsTighten` (4 cases for "most restrictive wins" + strict tiebreak).
- Full `pkg/utils/validator/...` and `pkg/utils/controller/...` suites pass; `make reviewable` clean.

- `validateRequirementsConsistency` is **first-error-wins** by design — once `collectRequirementBounds` hits a conflicting Eq it returns; downstream interval/Ne contradictions on the same property won't be reported in the same admission response. Documented in the function comment.
- Boy Scout: also switched two `interface{}` → `any` in the Service `nodePort` stripping block in `resource_selector_resolver.go` since I was already in that file.

### DIFF
--- a/pkg/utils/validator/placement.go
+++ b/pkg/utils/validator/placement.go
@@ -450,17 +450,22 @@ func validatePropertySelector(propertySelector *placementv1beta1.PropertySelecto
 
 func validatePropertySelectorRequirements(propertySelectorRequirements []placementv1beta1.PropertySelectorRequirement) error {
 	var allErr []error
+	// Group requirements by property name so we can also check cross-requirement
+	// contradictions (e.g., Eq 5 alongside Eq 10, or Gt 10 alongside Lt 5).
+	byName := make(map[string][]placementv1beta1.PropertySelectorRequirement, len(propertySelectorRequirements))
 	for _, req := range propertySelectorRequirements {
 		if err := validateName(req.Name); err != nil {
 			allErr = append(allErr, fmt.Errorf("invalid property name %s: %w", req.Name, err))
 		}
-		if err := validateOperator(req.Operator, req.Values); err != nil {
-			allErr = append(allErr, err)
+		if err := validateOperatorAndValues(req.Operator, req.Values); err != nil {
+			allErr = append(allErr, fmt.Errorf("invalid requirement on property %s: %w", req.Name, err))
 		}
-		if err := validateValues(req.Values); err != nil {
-			allErr = append(allErr, fmt.Errorf("invalid values for property %s: %w", req.Name, err))
+		byName[req.Name] = append(byName[req.Name], req)
+	}
+	for name, reqs := range byName {
+		if err := validateRequirementsConsistency(reqs); err != nil {
+			allErr = append(allErr, fmt.Errorf("inconsistent requirements on property %s: %w", name, err))
 		}
-		// TODO: Check for logical contradictions
 	}
 	return apiErrors.NewAggregate(allErr)
 }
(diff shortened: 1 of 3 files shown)


pkg/utils/validator/placement.go
```diff
@@ -530,26 +535,188 @@ func validateName(name string) error {
...
+	spec, ok := supportedPropertyOperators[op]
+	if !ok {
+		return fmt.Errorf("unsupported operator %s", op)
 	}
-	if validOperators[op] && len(values) != 1 {
-		return fmt.Errorf("operator %s requires exactly one value, got %d", op, len(values))
+	if len(values) != spec.requiredValueCount {
+		// Preserve the historical wording for the count==1 case (every supported operator today).
+		if spec.requiredValueCount == 1 {
+			return fmt.Errorf("operator %s requires exactly one value, got %d", op, len(values))
```
weng271190436: I think we should not have this special case. It would be unfortunate if we have an alert that tries to match exactly this "requires exactly one value." But I think that's an unmaintainable alert.

pkg/utils/validator/placement.go
```diff
@@ -530,26 +535,188 @@ func validateName(name string) error {
...
-		placementv1beta1.PropertySelectorGreaterThan:          true,
-		placementv1beta1.PropertySelectorGreaterThanOrEqualTo: true,
-		placementv1beta1.PropertySelectorLessThan:             true,
-		placementv1beta1.PropertySelectorLessThanOrEqualTo:    true,
-		placementv1beta1.PropertySelectorEqualTo:              true,
-		placementv1beta1.PropertySelectorNotEqualTo:           true,
+// supportedPropertyOperators lists the operators recognised on a PropertySelectorRequirement.
+// All currently-supported operators take exactly one resource.Quantity value; this table is the
+// single source of truth for both the operator validity check and the per-operator value rules.
+var supportedPropertyOperators = map[placementv1beta1.PropertySelectorOperator]struct {
```
weng271190436: nit: move this to the top of the file because this is a configuration that people might want to change from time to time. Should make it obvious to reader rather than buried deep in the middle of the file

pkg/utils/validator/placement.go
```diff
@@ -530,26 +535,188 @@ func validateName(name string) error {
...
+}
+
+// collectRequirementBounds parses each well-formed requirement into the normalised bounds view.
+// It returns early with an error if it sees two Eq requirements with different values; this is
+// a first-error-wins design, consistent with the rest of validateRequirementsConsistency. Other
+// inconsistencies are reported by the bound-checking helpers.
+func collectRequirementBounds(reqs []placementv1beta1.PropertySelectorRequirement) (*requirementBounds, error) {
+	out := &requirementBounds{}
+	for _, req := range reqs {
+		if _, ok := supportedPropertyOperators[req.Operator]; !ok || len(req.Values) != 1 {
```
weng271190436: Why `len(req.Values) != 1`? Each operator has a `requiredValueCount` config now. Should we check that instead of 1?

pkg/utils/validator/placement.go
```diff
@@ -530,26 +535,188 @@ func validateName(name string) error {
...
+			continue
+		}
+		q, err := resource.ParseQuantity(req.Values[0])
+		if err != nil {
+			continue
+		}
+		switch req.Operator {
+		case placementv1beta1.PropertySelectorEqualTo:
```
weng271190436: Should these switch case logic be part of the "supportedPropertyOperators"?

Such that validation logic for each operator has a real "single source of truth." And it is impossible for someone adding/removing an operator to forget about updating these switch cases.

Like this:

```go
type operatorSpec struct {
    requiredValueCount int
    applyToBounds      func(*requirementBounds, resource.Quantity) error
}

var supportedPropertyOperators = map[placementv1beta1.PropertySelectorOperator]operatorSpec{
    placementv1beta1.PropertySelectorGreaterThan:          {1, func(rb *requirementBounds, q resource.Quantity) error { rb.tightenLower(boundary{q: q, strict: true}); return nil }},
    placementv1beta1.PropertySelectorGreaterThanOrEqualTo: {1, func(rb *requirementBounds, q resource.Quantity) error { rb.tightenLower(boundary{q: q, strict: false}); return nil }},
    placementv1beta1.PropertySelectorLessThan:             {1, func(rb *requirementBounds, q resource.Quantity) error { rb.tightenUpper(boundary{q: q, strict: true}); return nil }},
    placementv1beta1.PropertySelectorLessThanOrEqualTo:    {1, func(rb *requirementBounds, q resource.Quantity) error { rb.tightenUpper(boundary{q: q, strict: false}); return nil }},
    placementv1beta1.PropertySelectorEqualTo:              {1, (*requirementBounds).applyEq},
    placementv1beta1.PropertySelectorNotEqualTo:           {1, func(rb *requirementBounds, q resource.Quantity) error { rb.neVals = append(rb.neVals, q); return nil }},
}

func (rb *requirementBounds) applyEq(q resource.Quantity) error {
    if rb.eqVal != nil && rb.eqVal.Cmp(q) != 0 {
        return fmt.Errorf("conflicting Eq values %s and %s", rb.eqVal.String(), q.String())
    }
    rb.eqVal = &q
    return nil
}
```

pkg/utils/validator/placement.go
weng271190436: I think handling len(req.Values) > 1 takes more work and is beyond the scope of this PR, might as well not have "requiredValueCount"?

Right now requiredValueCount gives an illusion that the code can handle an operator with arbitrary number of inputs.

And 1 should be a constant on top of the file. And admit in its comment that right now we can only handle operators that take exactly 1 input.

pkg/utils/validator/placement.go
ytimocin: Good idea! Addressed!

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/Azure/karpenter-provider-azure v1.5.1
 	github.com/crossplane/crossplane-runtime/v2 v2.1.0
 	github.com/evanphx/json-patch/v5 v5.9.11
+	github.com/go-logr/logr v1.4.3
```
ytimocin, line 11: Ran `go mod tidy` locally fixed an issue in main.

pkg/utils/validator/placement.go
ytimocin: Addressed! Let me know if updated code looks better!

pkg/utils/validator/placement.go
ytimocin: Addressed!

pkg/utils/validator/placement.go
ytimocin: Good call — done.

pkg/utils/validator/placement.go
weng271190436: you didn't consolidate the applyToBounds logic into the configuration `supportedPropertyOperators`?

weng271190436 (dismissed):

weng271190436 (approved):

PR #669, opened 2026-04-24
test: cover snapshot cross-scope isolation in DeleteResourceSnapshots

### LINKED ISSUE #647: Test resource snapshots sharing a placement name across scopes
## What
Coverage gap: need a way to test that namespaced and cluster-scoped snapshots can coexist when they share the same placement name. Covers both directions.

## Locations
- `pkg/utils/controller/resource_snapshot_resolver_test.go:1774` (namespaced case)
- `pkg/utils/controller/resource_snapshot_resolver_test.go:1796` (cluster-scoped case)

ytimocin:

Fixes #647.

Two existing cases in `TestDeleteResourceSnapshots` mix `ClusterResourceSnapshot` and `ResourceSnapshot` objects sharing the same placement name to exercise cross-scope isolation, but the verification block only lists the placement's own scope — so the "should NOT be deleted" assertions never fired. Two `// TODO: find a way to test this` markers flagged the gap.

This PR adds:

- An `expectedSurvivors []types.NamespacedName` field on the test table so each case can declare which snapshots must still exist after the delete.
- A small `snapshotExists` helper that distinguishes scope by empty-vs-non-empty namespace and uses the controller-runtime fake client's `Get`.
- Cross-scope verification after every `DeleteResourceSnapshots` call. The third "mixed environment" case now also gets explicit cross-scope and cross-namespace survivor assertions.

The two `// TODO: find a way to test this` comments are removed.

I have:

`go test -count=1 -run TestDeleteResourceSnapshots ./pkg/utils/controller/...` passes. The fake client correctly stores `ClusterResourceSnapshot` under the empty-namespace bucket and `ResourceSnapshot` under its actual namespace, so the helper's scope discrimination matches production semantics. `make reviewable` is clean.

- The first two test cases (single-scope deletes) have empty `expectedSurvivors` and a comment explaining the cross-scope check is vacuous there; the mixed-environment cases below carry the real assertions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/Azure/karpenter-provider-azure v1.5.1
 	github.com/crossplane/crossplane-runtime/v2 v2.1.0
 	github.com/evanphx/json-patch/v5 v5.9.11
+	github.com/go-logr/logr v1.4.3
 	github.com/google/go-cmp v0.7.0
 	github.com/onsi/ginkgo/v2 v2.23.4
 	github.com/onsi/gomega v1.37.0
@@ -65,7 +66,6 @@ require (
 	github.com/fsnotify/fsnotify v1.9.0 // indirect
 	github.com/fxamacker/cbor/v2 v2.9.0 // indirect
 	github.com/go-errors/errors v1.4.2 // indirect
-	github.com/go-logr/logr v1.4.3 // indirect
 	github.com/go-logr/zapr v1.3.0 // indirect
 	github.com/go-openapi/jsonpointer v0.21.1 // indirect
 	github.com/go-openapi/jsonreference v0.21.0 // indirect
--- a/pkg/utils/controller/resource_snapshot_resolver_test.go
+++ b/pkg/utils/controller/resource_snapshot_resolver_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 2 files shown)


pkg/utils/controller/resource_snapshot_resolver_test.go
```diff
@@ -1716,6 +1723,8 @@ func TestDeleteResourceSnapshots(t *testing.T) {
 					},
 				},
 			},
+			// No cross-scope objects seeded; the cross-scope survivor check is vacuous here.
```
Arvindthiru: nit: do we need this comment here ?

pkg/utils/controller/resource_snapshot_resolver_test.go
```diff
@@ -1864,14 +1883,54 @@ func TestDeleteResourceSnapshots(t *testing.T) {
...
-
 			if len(result.GetResourceSnapshotObjs()) != 0 {
 				t.Errorf("Expected 0 resource snapshots after deletion, got %d", len(result.GetResourceSnapshotObjs()))
 			}
+
+			// Cross-scope verification: snapshots in expectedSurvivors must still exist.
+			// This catches the case where DeleteResourceSnapshots accidentally deletes the
+			// other scope (or a different namespace) when placement names collide.
+			for _, want := range tt.expectedSurvivors {
+				found, err := snapshotExists(context.Background(), k8Client, want)
```
Arvindthiru: The test looks like it lists resourceSnapshots and then gets survivors again why not list both cluster scoped and namespace scoped resource snapshots and then check to see if the survivors exist in the list?

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/Azure/karpenter-provider-azure v1.5.1
 	github.com/crossplane/crossplane-runtime/v2 v2.1.0
 	github.com/evanphx/json-patch/v5 v5.9.11
+	github.com/go-logr/logr v1.4.3
```
ytimocin, line 11: carry-over go mod tidy

pkg/utils/controller/resource_snapshot_resolver_test.go
```diff
@@ -1683,10 +1683,12 @@ func TestDeleteResourceSnapshots(t *testing.T) {
...
 	tests := []struct {
-		name          string
-		placementObj  fleetv1beta1.PlacementObj
-		objects       []client.Object
-		expectedError string
+		name         string
+		placementObj fleetv1beta1.PlacementObj
+		objects      []client.Object
+		// Snapshots that must still exist after the delete. Nil means every seeded object should be deleted.
+		expectedSurvivors []types.NamespacedName
```
Arvindthiru: nit: we should use wantSurvivors, wantError instead of expected to follow convention

Arvindthiru (approved):

PR #670, opened 2026-04-24
fix: don't treat in-progress binding states as terminal failures

### LINKED ISSUE #648: HasBindingFailed treats all condition failures as terminal
## What
`HasBindingFailed` returns `true` on any false condition without inspecting the reason. Recoverable/retriable failures are therefore reported as hard failures, which can drive unnecessary behavior upstream (skipped retries, premature status reporting).

## Proposal
Parse the condition reason to distinguish retriable failures from terminal ones.

## Location
- `pkg/utils/binding/binding.go:30`

ytimocin:

Fixes #648.

`HasBindingFailed` previously returned `true` for any per-cluster binding condition with `Status=False`, regardless of `Reason`. Several **in-progress** reasons on those conditions (Overridden, WorkSynchronized, Applied, Available) are expressed as `Status=False` rather than `Status=Unknown` — most notably `WorkNotSynchronizedYetReason` and `NotAvailableYetReason`. The over-broad classification caused upstream controllers (rollout, updaterun) to bail on bindings that were still progressing.

This PR switches `HasBindingFailed` to consult an explicit **allowlist** of non-terminal reasons (`nonTerminalBindingFailureReasons`) drawn from the `condition` package:

- `OverriddenPendingReason`
- `WorkNotSynchronizedYetReason`
- `ApplyPendingReason`
- `NotAvailableYetReason`

Anything outside that set is treated as terminal — the safer default, since wrongly classifying an unknown reason as "transient" would stall rollouts forever.

I have:

- New `TestHasBindingFailed_NonTerminalReasons` (8 cases) covers each known transient reason returning false, each known terminal reason returning true, an unknown-reason fall-through to terminal, and a mixed-condition case where a transient reason on one condition doesn't mask a terminal failure on another.
- Existing `TestHasBindingFailed` (camelCase reason strings not in the allowlist) still passes — those legacy strings are correctly classified as terminal.
- `TestCheckClusterUpdateResult` in `pkg/controllers/updaterun` was encoding the previous bug — it asserted `WorkNotSynchronizedYet` should produce an error. Updated to match corrected semantics (in-progress → no error) and added a new case covering the terminal `SyncWorkFailedReason` path. This aligns with the existing TODO at `execution.go:626` that called for distinguishing recoverable from non-recoverable failures.
- `pkg/controllers/rollout/...` (unit + Ginkgo integration), `pkg/controllers/updaterun/...`, and `pkg/controllers/clusterresourceplacementeviction/...` all pass locally with envtest. `make reviewable` is clean.

- An earlier draft used `strings.Contains(strings.ToLower(reason), "failed")` for the classifier. That was rejected during review as fragile (typo-prone, convention-drift-prone). The allowlist uses typed constants instead.
- New "Pending" / "NotYet" reasons added to `pkg/utils/condition/reason.go` must also be added to `nonTerminalBindingFailureReasons`. This is documented inline in `binding.go`.
- The updaterun test change is the only behaviour-shift visible to callers; the rollout controller's behaviour also changes (in-progress bindings now stay in `canBeReadyBindings` instead of being dropped), but no existing rollout tests were tied to the buggy behaviour.

### DIFF
--- a/pkg/utils/binding/binding.go
+++ b/pkg/utils/binding/binding.go
@@ -23,14 +23,51 @@ import (
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/condition"
 )
 
-// HasBindingFailed checks if BindingObj has failed based on its conditions.
+// nonTerminalBindingFailureReasons is the closed set of `Reason` strings that the per-cluster
+// binding conditions (Overridden, WorkSynchronized, Applied, Available) can carry while still
+// `Status=False` *without* representing a terminal failure. Each one means the binding is still
+// progressing and the controller should keep waiting rather than treat the binding as failed.
+//
+// The set is intentionally an allowlist: any new `Status=False` reason added to the codebase is
+// treated as terminal until it is explicitly added here. This is the safer default because the
+// alternative (treat unknown reasons as transient) would stall the rollout controller forever
+// on a genuinely-failing binding it does not yet know how to classify.
+//
+// When you add a new in-progress / pending / "not yet" reason to pkg/utils/condition/reason.go
+// for any of the above conditions, also add it here.
+var nonTerminalBindingFailureReasons = map[string]struct{}{
+	condition.OverriddenPendingReason:      {},
+	condition.WorkNotSynchronizedYetReason: {},
+	condition.ApplyPendingReason:           {},
+	condition.NotAvailableYetReason:        {},
+}
+
+// HasBindingFailed reports whether a binding has reached a terminal failure on any of its
+// per-cluster conditions (Overridden, WorkSynchronized, Applied, Available).
+//
+// A condition counts as a terminal failure when its `Status` is `False` *and* its `Reason` is
+// not in `nonTerminalBindingFailureReasons`. The Reason check is necessary because several of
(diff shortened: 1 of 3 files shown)


pkg/utils/binding/binding.go
```diff
@@ -23,14 +23,51 @@ import (
...
+//
+// Unknown reasons fall through to "terminal" by design; see the comment on
+// `nonTerminalBindingFailureReasons`.
 func HasBindingFailed(binding placementv1beta1.BindingObj) bool {
 	for i := condition.OverriddenCondition; i <= condition.AvailableCondition; i++ {
-		if condition.IsConditionStatusFalse(binding.GetCondition(string(i.ResourceBindingConditionType())), binding.GetGeneration()) {
-			// TODO: parse the reason of the condition to see if the failure is recoverable/retriable or not
-			klog.V(2).Infof("binding %s has condition %s with status false", klog.KObj(binding), string(i.ResourceBindingConditionType()))
-			return true
+		c := binding.GetCondition(string(i.ResourceBindingConditionType()))
```
britaniar, line 59: nit: could be its own variable since it is used multiple times in this code.

britaniar (approved):
Minor nit but LGTM!

Issue #627, opened 2026-04-23
[Feature] Convention: log errors where handled, wrap when returning


ytimocin:
**Is your feature request related to a problem? Please describe.**

We have no convention for where to log errors. On-spot `klog.ErrorS` and caller-side logging are both used, so one failure often produces multiple log lines. `NewAPIError` / `NewUnexpectedError` also log while wrapping, which compounds the double-log.

Surfaced in the review thread on #575.

**Describe the solution you'd like**

**Log where the error is handled, not at each return. Wrap otherwise.**
- Handled at `main` → log at `main`.
- Recovered mid-stack → log there: `warning: x failed, handled by z: failed to parse: ...`
- Intermediate layers just `fmt.Errorf("...: %w", err)`.

Concrete changes:
1. Add a slim error wrapper that accumulates key-value pairs and keeps the `%w` chain, so structured logging still works when logging happens once at the top.
2. Remove the log side-effect from `NewAPIError` / `NewUnexpectedError`.
3. Document the rule in `CONTRIBUTING.md`.
4. Migrate controllers incrementally.

**Describe alternatives you've considered**

- **Status quo + stricter review.** Cheap, but we've already drifted.
- **Ban on-spot logging outright.** Loses structured kv-pair ergonomics and leaves `NewAPIError` dependents stranded.

**Additional context**

On-spot `klog.ErrorS` is currently used precisely for its structured kv-pairs — a wrap-and-bubble model has to carry those up the chain or we lose context.

PR #679, opened 2026-04-28
feat: add new error handling utilities

michaelawyu:

EXPERIMENTAL PR: posted for discussion.

This PR adds some utilities that allows better error handling.

The goal is primarily to:

* facilitate an error handling pattern where errors are wrapped in the caller chain, and logged with `klog.ErrorS` at the top level.
* enable additional k-v attribute bookkeeping at each level, so that contextual information can be recorded without having to call `klog.ErrorS` at each level.
* enable error categorization at the bottom level, where categorization should happen, but write the category as part of the output only at the top-level
* enable additional k-v attribute auto-capturing for specific categories

Refs #627 

I have:

N/A -> posted for discussion only; if the pattern is OK, tests will be added later.

### DIFF
--- a/pkg/utils/errors/errors.go
+++ b/pkg/utils/errors/errors.go
@@ -82,15 +82,6 @@ func (e *Error) categoryWithDefault() ErrCategory {
 // Note the output intentionally does not include the additional attributes, so as to keep the cardinality
 // low; to print them out, pass them explicitly to the klog functions.
 func (e *Error) Error() string {
-	// Build a bitmap to determine the error template to use.
-	var bitmap int
-	if e.wrapped != nil {
-		bitmap |= 1 << 0
-	}
-	if len(e.desc) > 0 {
-		bitmap |= 1 << 1
-	}
-
 	switch {
 	case e.wrapped != nil && len(e.desc) == 0:
 		// With wrapped error but no description.
@@ -120,11 +111,13 @@ func Wraps(err error, desc string, kvs ...interface{}) *Error {
 	// Find if any of error down in the chain is already an *Error; if so, use its category and surface its
 	// attributes to the current level.
 	var category ErrCategory
+	var mergedKVs []interface{}
 	var childError *Error
 	if errors.As(err, &childError) {
 		category = childError.category
-		kvs = append(childError.attrs, kvs...)
+		mergedKVs = append(mergedKVs, childError.attrs...)
 	}
+	mergedKVs = append(mergedKVs, kvs...)
 	if len(category) == 0 {
 		category = ErrCategoryUncategorized
 	}
@@ -133,7 +126,7 @@ func Wraps(err error, desc string, kvs ...interface{}) *Error {
 		category: category,
 		wrapped:  err,
 		desc:     desc,
-		attrs:    kvs,
+		attrs:    mergedKVs,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/utils/errors/errors_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Fatalf("Failed to read from bytes buffer: %v", err)
+	}
+	output = output[:n]
+	outputStr := string(output)
+
+	// The full error output looks like the follows:
+	// errors_test.go:53: time=2026-04-29T02:39:44.853+10:00 level=ERROR msg="additional top/controller-level error description" err="a/an Kubernetes API server error occurred: additional high-level error description: failed to create namespace: namespaces \"work\" already exists" k1=v1 cached=false APIServerErrReason=AlreadyExists APIServerResHTTPCode=409 k2=v2 k3=v3
+	wantSubStrings := []string{
+		"msg=\"additional top/controller-level error description\"",
+		"err=\"a/an Kubernetes API server error occurred: additional high-level error description: failed to create namespace: namespaces \\\"work\\\" already exists\"",
```
weng271190436: Should category be a key rather than being part of the err string so that it is easier to query?

weng271190436: This way you also don't need isRootLevel I think?

pkg/utils/errors/errors.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	if len(e.desc) > 0 {
+		bitmap |= 1 << 1
+	}
+	if e.isRootLevel {
+		bitmap |= 1 << 2
+	}
+
+	switch bitmap {
+	case 1:
```
weng271190436: Might be easier to read if use binary literals.

```go
switch bitmap {
case 0b001:
    // wrapped=yes, desc=no, rootLevel=no
    return e.wrapped.Error()
case 0b010:
    // wrapped=no, desc=yes, rootLevel=no
    return e.desc
case 0b011:
    // wrapped=yes, desc=yes, rootLevel=no
    return fmt.Sprintf("%s: %s", e.desc, e.wrapped.Error())
case 0b101:
    // wrapped=yes, desc=no, rootLevel=yes
    return fmt.Sprintf("a/an %s error occurred: %s", e.categoryWithDefault(), e.wrapped.Error())
case 0b110:
    // wrapped=no, desc=yes, rootLevel=yes
    return fmt.Sprintf("a/an %s error occurred: %s", e.categoryWithDefault(), e.desc)
case 0b111:
    // wrapped=yes, desc=yes, rootLevel=yes
    return fmt.Sprintf("a/an %s error occurred: %s: %s", e.categoryWithDefault(), e.desc, e.wrapped.Error())
default:
    // 0b000: no wrapped, no desc, not root
    // 0b100: no wrapped, no desc, root
    return "an unknown error occurred"
}
```

pkg/utils/errors/errors_test.go
michaelawyu: ~NVM~

pkg/utils/errors/errors.go
michaelawyu: Thanks 🙏 Let me make the edit. -> Done

pkg/utils/errors/errors_test.go
michaelawyu: On second thought this might be better -> let me make the edit.

weng271190436 (approved):

britaniar (approved):

PR #685, opened 2026-04-30
feat: make update run stuck threshold configureable

### LINKED ISSUE #640: [Feature] Make updateRunStuckThreshold configurable
**Is your feature request related to a problem? Please describe.**
StagedUpdateRun waits 5 minutes before making the update run as stuck for a resource taking a while to apply.

**Describe the solution you'd like**
Make updateRunStuckThreshold configurable.
https://github.com/kubefleet-dev/kubefleet/blob/b548d9ba15e23dcad9f4be3403d75c4e9353ef91/pkg/controllers/updaterun/execution.go#L51-L53

britaniar:

Fixes #640 

I have: made the update run configurable by adding a flag that will be passed to the update run reconciler.

- UT
- Integration Test


pkg/controllers/updaterun/execution.go
```diff
@@ -47,10 +47,6 @@ var (
 	// stageUpdatingWaitTime is the time to wait before rechecking the stage update status.
 	// Put it as a variable for convenient testing.
 	stageUpdatingWaitTime = 60 * time.Second
-
-	// updateRunStuckThreshold is the time to wait on a single cluster update before marking update run as stuck.
-	// TODO(wantjian): make this configurable
-	updateRunStuckThreshold = 5 * time.Minute
```
Arvindthiru, line 53: This looks it's an API level change where we can allow users to configure the stuckThreshold for each updateRun

pkg/controllers/updaterun/execution.go
```diff
@@ -39,6 +39,11 @@ import (
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/controller"
 )
 
+const (
+	// defaultStuckThreshold is the default duration after which a cluster update is considered stuck.
+	defaultStuckThreshold = 5 * time.Minute
```
Arvindthiru: Why are we not adding a marker for default value in the CRD ?

apis/placement/v1beta1/stageupdate_types.go
```diff
@@ -202,6 +202,14 @@ type UpdateRunSpec struct {
 	// +kubebuilder:default=Initialize
 	// +kubebuilder:validation:Enum=Initialize;Run;Stop
 	State State `json:"state,omitempty"`
+
+	// StuckThreshold is the duration after which a cluster update is considered stuck if no progress is made.
```
Arvindthiru:
```suggestion
	// StuckThreshold is the duration after which a cluster update is considered stuck if propagated resources are either not applied or available.
```

pkg/controllers/updaterun/controller.go
```diff
@@ -567,3 +567,13 @@ func updateAllStageStatusConditionsGeneration(stageStatus *placementv1beta1.Stag
 		}
 	}
 }
+
+// getStuckThreshold returns the stuck threshold for the given update run.
+// It first checks if the update run has a stuck threshold specified in its spec,
+// and falls back to the default threshold if not specified.
+func getStuckThreshold(updateRun placementv1beta1.UpdateRunObj) time.Duration {
+	if spec := updateRun.GetUpdateRunSpec(); spec != nil && spec.StuckThreshold != nil {
```
Arvindthiru, line 575: If the default CRD marker is applied, spec being nil, StuckThreshold being nil is something we can avoid and hence we avoid this access function altogether ?

pkg/controllers/updaterun/initialization.go
```diff
@@ -635,6 +639,22 @@ func (r *Reconciler) getResourceSnapshotObjs(ctx context.Context, placement plac
...
 
+// validateStuckThreshold validates the stuck threshold specified in the update run spec.
+// Returns an error if the threshold is zero or negative.
+func validateStuckThreshold(updateRun placementv1beta1.UpdateRunObj) error {
+	spec := updateRun.GetUpdateRunSpec()
+	if spec == nil || spec.StuckThreshold == nil {
+		return nil
+	}
+	threshold := spec.StuckThreshold.Duration
+	if threshold <= 0 {
```
Arvindthiru: We won't need this check if follow the CRD marker for waitTime in StageTask https://github.com/kubefleet-dev/kubefleet/blob/main/apis/placement/v1beta1/stageupdate_types.go#L350

pkg/controllers/updaterun/execution.go
ytimocin: Shouldn't this be in a types.go?

pkg/controllers/updaterun/initialization.go
```diff
@@ -635,6 +639,22 @@ func (r *Reconciler) getResourceSnapshotObjs(ctx context.Context, placement plac
...
 }
 
+// validateStuckThreshold validates the stuck threshold specified in the update run spec.
+// Returns an error if the threshold is zero or negative.
+func validateStuckThreshold(updateRun placementv1beta1.UpdateRunObj) error {
+	spec := updateRun.GetUpdateRunSpec()
+	if spec == nil || spec.StuckThreshold == nil {
+		return nil
+	}
+	threshold := spec.StuckThreshold.Duration
```
ytimocin: nit: this can be in-lined. do we actually need threshold variable?

test/apis/placement/v1beta1/api_validation_integration_test.go
```diff
@@ -1464,6 +1493,24 @@ var _ = Describe("Test placement v1beta1 API validation", func() {
 			Expect(statusErr.ErrStatus.Message).Should(MatchRegexp("metadata.name max length is 63"))
 		})
 
+		It("Should deny creation of ClusterStagedUpdateRun stuckThreshold field that is not s, m, or h", func() {
```
Arvindthiru, line 1497: can we add a test testing the negative duration being disallowed

pkg/controllers/updaterun/controller.go
britaniar: We cannot do this if want it to be immutable like the rest of the spec.

pkg/controllers/updaterun/execution.go
britaniar: Default is applied on every write operation so everytime it is created/updated. If we want this field to be immutable like every other field then we cannot use default.

PR #686, opened 2026-05-04
fix: specify which object failed in override error message

ytimocin:

When an override snapshot fails to apply on a selected resource, the binding's `Overridden=False` condition message now identifies **which** override snapshot and **which** target object failed, instead of surfacing only the underlying patch error.

Before:
> `Failed to apply the override rules on the resources: failed to process the request due to a client error: add operation does not apply: doc is missing key: /invalid`

After:
> `Failed to apply the override rules on the resources: ClusterResourceOverrideSnapshot "cro-2" failed to apply on Namespace "app": add operation does not apply: doc is missing key: /invalid`

**Implementation:** the wrap is done once at the outer `applyOverrides` call site so the existing controller-level message trim at `controller.go:188-194` works correctly without doubled sentinel prefixes. `applyOverrideRules` and `applyJSONPatchOverride` no longer call `controller.NewUserError` themselves — they return raw errors, and `applyOverrides` does the sentinel-tagging once with the snapshot + target context. `errors.Is(err, controller.ErrUserError)` still holds end-to-end.

Boy Scout cleanups limited to the files touched by this fix:
- `override.go` — `isClusterScopeResource` → `isClusterScopedResource`; capitalised log messages; `cluster.ObjectMeta.Labels` → `cluster.Labels`; doc comments on three unexported functions; redundant `// do nothing` dropped.
- `override_test.go` — `unmarshl` → `unmarshal` (3 sites); `clusterole` → `clusterRole`; corrected the wrong function name in a test failure diagnostic (`TestReplaceClusterLabelKeyVariables` was logging `applyJSONPatchOverride()`); `expected/expectErr` → `want/wantErr` per project style; test renamed `TestApplyOverrides_namespacedScopeResource` → `_namespaceScopedResource`; explanatory comment added for the inverted-looking `IsClusterScopedResource: false` `FakeManager` flag.
- `controller_integration_test.go` — double-space typos in `By(...)` strings.

I have:

- `make test` passes (`pkg/controllers/workgenerator` runs 86 Ginkgo specs plus unit tests; `pkg/utils/condition` and `pkg/controllers/placement` unaffected and green).
- **Unit coverage:**
  - `TestApplyOverrides_clusterScopedResource` and `TestApplyOverrides_namespaceScopedResource` gained a `wantErrSubstr` field. Four failure cases now assert on the new message format (snapshot identity + target identity), including two cases that previously had no `wantErrSubstr` and would have surfaced an empty snapshot name post-change.
  - New case for the `IsClusterMatched` error path through `applyOverrideRules` (an invalid `MatchExpressions` `Operator` makes `metav1.LabelSelectorAsSelector` fail). This path was previously unexercised in `override_test.go`.
- **Integration coverage:**
  - The existing `Bound ClusterResourceBinding … invalid override` context now asserts via `MatchRegexp` on the new message structure.
  - New sibling context `Bound ClusterResourceBinding … invalid resource override` mirrors the CRO failure path for `ResourceOverrideSnapshot`. New fixture `invalidResourceOverrideSnapshot` (`ro-2`) added in `suite_test.go`.

- **Single-call wrap is intentional.** `applyOverrideRules` and `applyJSONPatchOverride` no longer call `controller.NewUserError`. Re-wrapping an already-tagged error created a doubled `"failed to process the request due to a client error: "` prefix that the trim at `controller.go:188-194` couldn't strip cleanly. Doing the wrap once at `applyOverrides` keeps the `%w: <suffix>` shape the trim expects.
- **Trim untouched.** The pre-existing trim at `controller.go:188-194` and its `//TODO: check if it's user error and set a different failed reason` are unchanged; cleaning that up is a separate refactor.
- **Scope discipline.** Two follow-on improvements were considered and explicitly deferred to keep this PR focused:
  - Firing a Kubernetes `Event` on `Overridden=False` so `kubectl describe binding` shows the same enriched message.
  - Enriching the CRP/RP `Overridden=False` rollup condition message with the failing cluster names instead of just a count.

### DIFF
--- a/pkg/controllers/workgenerator/override.go
+++ b/pkg/controllers/workgenerator/override.go
@@ -35,6 +35,11 @@ import (
 	"github.com/kubefleet-dev/kubefleet/pkg/utils/overrider"
 )
 
+// fetchClusterResourceOverrideSnapshots returns the ClusterResourceOverrideSnapshots referenced
+// by the given binding, keyed by the ResourceIdentifier each snapshot's selectors target. The
+// returned map preserves the order in which the binding lists the snapshots, so the caller can
+// apply them in a deterministic sequence.
+//
 // TODO: combine the following two functions into one, as they are very similar.
 func (r *Reconciler) fetchClusterResourceOverrideSnapshots(ctx context.Context, resourceBinding placementv1beta1.BindingObj) (map[placementv1beta1.ResourceIdentifier][]*placementv1beta1.ClusterResourceOverrideSnapshot, error) {
 	croMap := make(map[placementv1beta1.ResourceIdentifier][]*placementv1beta1.ClusterResourceOverrideSnapshot)
@@ -70,6 +75,10 @@ func (r *Reconciler) fetchClusterResourceOverrideSnapshots(ctx context.Context, 
 	return croMap, nil
 }
 
+// fetchResourceOverrideSnapshots returns the ResourceOverrideSnapshots referenced by the given
+// binding, keyed by the ResourceIdentifier each snapshot's selectors target. The returned map
+// preserves the order in which the binding lists the snapshots, so the caller can apply them in
+// a deterministic sequence.
 func (r *Reconciler) fetchResourceOverrideSnapshots(ctx context.Context, resourceBinding placementv1beta1.BindingObj) (map[placementv1beta1.ResourceIdentifier][]*placementv1beta1.ResourceOverrideSnapshot, error) {
 	roMap := make(map[placementv1beta1.ResourceIdentifier][]*placementv1beta1.ResourceOverrideSnapshot)
 
@@ -127,11 +136,11 @@ func (r *Reconciler) applyOverrides(resource *placementv1beta1.ResourceContent, 
 		Kind:    gvk.Kind,
 		Name:    uResource.GetName(),
 	}
-	isClusterScopeResource := r.InformerManager.IsClusterScopedResources(gvk)
+	isClusterScopedResource := r.InformerManager.IsClusterScopedResources(gvk)
 
 	// For the namespace scoped resource, it could be selected by the namespace itself.
-	// use the namespace as the key
-	if !isClusterScopeResource {
+	// Use the namespace as the key.
+	if !isClusterScopedResource {
 		key = placementv1beta1.ResourceIdentifier{
 			Group:   utils.NamespaceMetaGVK.Group,
 			Version: utils.NamespaceMetaGVK.Version,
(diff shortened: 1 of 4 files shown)


pkg/controllers/workgenerator/override.go
```diff
@@ -172,20 +182,36 @@ func (r *Reconciler) applyOverrides(resource *placementv1beta1.ResourceContent,
...
 	}
 	return resource.Raw == nil, nil
 }
 
+// formatOverrideTarget returns a human-readable identifier for the resource being overridden,
+// suitable for inclusion in a user-facing error message (e.g. `Deployment "my-app" in namespace "default"`).
+func formatOverrideTarget(target *unstructured.Unstructured) string {
+	kind := target.GetObjectKind().GroupVersionKind().Kind
+	if ns := target.GetNamespace(); ns != "" {
+		return fmt.Sprintf("%s %q in namespace %q", kind, target.GetName(), ns)
```
britaniar, line 201: does this have the quotations around the target name and ns like you have in the example?
`Deployment "my-app" in namespace "default"`

pkg/controllers/workgenerator/override.go
```diff
@@ -172,20 +182,36 @@ func (r *Reconciler) applyOverrides(resource *placementv1beta1.ResourceContent,
...
+// clears the resource and ends rule application; otherwise the rule's JSON patches are applied
+// in order. The first rule that fails returns its raw error to the caller, which is responsible
+// for tagging the failure as a user error.
 func applyOverrideRules(resource *placementv1beta1.ResourceContent, cluster *clusterv1beta1.MemberCluster, rules []placementv1beta1.OverrideRule) error {
 	for _, rule := range rules {
 		matched, err := overrider.IsClusterMatched(cluster, rule)
 		if err != nil {
 			klog.ErrorS(controller.NewUnexpectedBehaviorError(err), "Found an invalid override rule")
-			return controller.NewUserError(err) // should not happen though and should be rejected by the webhook
+			return err // should not happen though and should be rejected by the webhook
```
britaniar: nit: I am a little hesitant about this because a user (with enough privileges aka cluster-admin) could delete the webhook and then this would not be caught.

pkg/controllers/workgenerator/override.go
ytimocin: Yes, it does.

pkg/controllers/workgenerator/override.go
ytimocin: The wrap just moved from the inner helper to the outer call site — **controller.NewUserError(fmt.Errorf(...))** still tags **ErrUserError**, so **errors.Is** holds even if the webhook is bypassed. It had to move so the message could include the **snapshot + target** identity, which the inner helper doesn't know.

britaniar (approved):

PR #691, opened 2026-05-05
fix: close override snapshot transition race on update

ytimocin:

When a `ClusterResourceOverride` or `ResourceOverride` is updated, the controller previously flipped `is-latest-snapshot=false` on the old snapshot **before** creating the new one. The rollout snapshot watcher only handles `Create`/`Generic` events (no `UpdateFunc`), so a crash between flip and create left rollout un-enqueued until the next resync (6 h hub-agent default) or parent-CR mutation, and `FetchAllMatchingOverridesForResourceSnapshot` would silently drop the affected override.

This PR reorders the controller to **create-first, demote-after**, with two additional safety steps so transient duplicate `latest=true` is benign:

- New snapshot is `Create`d with `latest=true` first; the previous snapshot is flipped to `false` afterwards.
- On `AlreadyExists`, the existing snapshot is fetched via `mgr.GetAPIReader()` (uncached) and its `OverrideHash` verified before proceeding. Mismatches surface as a `Warning` event on the parent CRO/RO and an `ExpectedBehaviorError` (no stack trace, requeues for transient causes like etcd restore).
- A new `cleanupStaleLatestSiblings` helper flips any older sibling that still carries `latest=true`. Runs on both the hash-match and new-snapshot paths.
- `FetchAllMatchingOverridesForResourceSnapshot` deduplicates at read time, picking the highest `OverrideIndexLabel` per parent (CRO key: tracking label; RO key: namespace + tracking label). Malformed labels log loudly and skip rather than blocking the entire rollout hot path.
- `IsNotFound` on the demote `Update` is tolerated so the sibling audit still runs after a concurrent prune or parent deletion.

Fixes #

I have:

- New unit tests in `pkg/utils/overrider/overrider_test.go` cover the generic dedup helper (CRO and RO, including the namespace-collision and malformed-label cases) and a `FetchAllMatchingOverridesForResourceSnapshot` integration case with two `latest=true` snapshots for the same parent.
- New integration tests in `pkg/controllers/overrider/common_test.go` cover the sibling-audit helper (steady-state no-op, post-crash recovery) and the `AlreadyExists` recovery branch (matching hash → demote proceeds; mismatched hash → `ExpectedBehaviorError` + Warning event, no demote).
- All existing override controller integration tests (envtest) and `pkg/utils/overrider` unit tests pass.

- `Placement == nil` rollout watcher skip (`pkg/controllers/rollout/controller.go:849, :882`) and the same flip-then-create antipattern in `pkg/utils/controller/resource_snapshot_resolver.go` are left out of scope; both deserve their own follow-up issues.
- Long-term we should consider migrating to a `status.latestSnapshotName` pointer on the parent CRO/RO so labels stop being load-bearing for "current" semantics; that would be a separate, larger change.

### DIFF
--- a/pkg/controllers/overrider/clusterresource_controller.go
+++ b/pkg/controllers/overrider/clusterresource_controller.go
@@ -23,9 +23,11 @@ import (
 	"strconv"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -42,7 +44,7 @@ type ClusterResourceReconciler struct {
 	Reconciler
 }
 
-// Reconcile triggers a single  reconcile round when scheduling policy has changed.
+// Reconcile triggers a single reconcile round when the override has changed.
 func (r *ClusterResourceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	name := req.NamespacedName
 	clusterOverride := placementv1beta1.ClusterResourceOverride{}
@@ -99,29 +101,24 @@ func (r *ClusterResourceReconciler) ensureClusterResourceOverrideSnapshot(ctx co
 		return err
 	}
 
-	latestSnapshotIndex := -1 //so index start at 0
+	latestSnapshotIndex := -1 // so index starts at 0
+	var latestSnapshot *placementv1beta1.ClusterResourceOverrideSnapshot
 	if len(snapshotList.Items) != 0 {
-		// convert the last unstructured snapshot to the typed object
-		latestSnapshot := &placementv1beta1.ClusterResourceOverrideSnapshot{}
+		// Convert the last (highest-index) unstructured snapshot to the typed object.
+		latestSnapshot = &placementv1beta1.ClusterResourceOverrideSnapshot{}
 		if err = runtime.DefaultUnstructuredConverter.FromUnstructured(snapshotList.Items[len(snapshotList.Items)-1].Object, latestSnapshot); err != nil {
 			klog.ErrorS(err, "Invalid overrideSnapshot", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(&snapshotList.Items[len(snapshotList.Items)-1]))
 			return controller.NewUnexpectedBehaviorError(err)
 		}
 		if string(latestSnapshot.Spec.OverrideHash) == overrideSpecHash {
-			// the content has not changed, so we don't need to create a new snapshot.
-			return r.ensureSnapshotLatest(ctx, latestSnapshot)
-		}
-		// mark the last policy snapshot as inactive if it is different from what we have now.
-		if latestSnapshot.Labels[placementv1beta1.IsLatestSnapshotLabel] == strconv.FormatBool(true) {
-			// set the latest label to false first to make sure there is only one or none active policy snapshot
-			latestSnapshot.Labels[placementv1beta1.IsLatestSnapshotLabel] = strconv.FormatBool(false)
-			if err := r.Client.Update(ctx, latestSnapshot); err != nil {
-				klog.ErrorS(err, "Failed to set the isLatestSnapshot label to false", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(latestSnapshot))
-				return controller.NewUpdateIgnoreConflictError(err)
+			// The content has not changed; no new snapshot is needed.
+			// Ensure the highest-index snapshot is marked latest, then audit siblings to clean
+			// up any duplicate latest=true left by a prior partial run.
+			if err := r.ensureSnapshotLatest(ctx, latestSnapshot); err != nil {
+				return err
 			}
-			klog.V(2).InfoS("Marked the latest overrideSnapshot as inactive", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(latestSnapshot))
+			return r.cleanupStaleLatestSiblings(ctx, snapshotList)
 		}
-		// we need to figure out the last snapshot index.
 		latestSnapshotIndex, err = labels.ExtractIndex(latestSnapshot, placementv1beta1.OverrideIndexLabel)
 		if err != nil {
 			klog.ErrorS(err, "Failed to parse the override index label", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(latestSnapshot))
@@ -129,7 +126,10 @@ func (r *ClusterResourceReconciler) ensureClusterResourceOverrideSnapshot(ctx co
 		}
 	}
 
-	// Need to create new snapshot when 1) there is no snapshots or 2) the latest snapshot hash != current one.
+	// Create the new snapshot (latest=true) BEFORE flipping the old one to false. This avoids a
+	// zero-latest window: if we crash between operations, at worst we leave two snapshots with
+	// latest=true, which is resolved by read-time dedup and cleaned up by cleanupStaleLatestSiblings
+	// on the next reconcile that hits the hash-match path above.
 	latestSnapshotIndex++
 	newSnapshot := &placementv1beta1.ClusterResourceOverrideSnapshot{
 		ObjectMeta: metav1.ObjectMeta{
@@ -149,15 +149,70 @@ func (r *ClusterResourceReconciler) ensureClusterResourceOverrideSnapshot(ctx co
 		},
 	}
 	if err := r.Client.Create(ctx, newSnapshot); err != nil {
-		klog.ErrorS(err, "Failed to create new overrideSnapshot", "newOverrideSnapshot", klog.KObj(newSnapshot))
-		return controller.NewAPIServerError(false, err)
+		if !errors.IsAlreadyExists(err) {
+			klog.ErrorS(err, "Failed to create new overrideSnapshot", "newOverrideSnapshot", klog.KObj(newSnapshot))
+			return controller.NewAPIServerError(false, err)
+		}
+		// AlreadyExists here normally means a prior reconcile succeeded the Create but failed to
+		// flip the old snapshot. The deterministic name {ParentName}-{index} plus the hash check
+		// above mean the existing object should have identical content. Verify before proceeding,
+		// because etcd restore from backup, manual edits, or a future hash-function change could
+		// leave an existing object whose hash no longer matches the current spec.
+		//
+		// Read through the uncached client: the cached read can lag the server's write to the
+		// just-Created object and return NotFound, forcing a needless requeue.
+		existing := &placementv1beta1.ClusterResourceOverrideSnapshot{}
+		if getErr := r.UncachedReader.Get(ctx, types.NamespacedName{Name: newSnapshot.Name}, existing); getErr != nil {
+			klog.ErrorS(getErr, "Failed to get existing overrideSnapshot for hash verification", "clusterResourceOverride", croKObj, "newOverrideSnapshot", klog.KObj(newSnapshot))
+			return controller.NewAPIServerError(false, getErr)
+		}
+		if string(existing.Spec.OverrideHash) != overrideSpecHash {
+			mismatchErr := fmt.Errorf("existing overrideSnapshot %s has hash %q, want %q", newSnapshot.Name, string(existing.Spec.OverrideHash), overrideSpecHash)
+			klog.ErrorS(mismatchErr, "Existing overrideSnapshot has different content than expected; will requeue and retry", "clusterResourceOverride", croKObj)
+			// Surface to the operator via an Event on the parent CRO so it shows up in
+			// `kubectl describe`. Without this, a permanent mismatch (manual edit, hash-fn break)
+			// would requeue forever with only log noise to indicate the problem.
+			if r.recorder != nil {
+				r.recorder.Eventf(cro, corev1.EventTypeWarning, "OverrideSnapshotHashMismatch",
+					"existing snapshot %s has a different hash than the current spec; retrying. If this persists, the snapshot needs operator inspection (etcd restore from backup, manual edit, or hash-function change).", newSnapshot.Name)
+			}
+			// Treat as expected-behavior: the most likely real-world trigger is etcd restore from
+			// backup, where retry resolves once the parent CRO converges. A truly unrecoverable
+			// state (manual edit, hash-fn break) will keep producing this log on every retry,
+			// which is enough for operators to notice without flooding stack traces.
+			return controller.NewExpectedBehaviorError(mismatchErr)
+		}
+		klog.V(2).InfoS("Snapshot already exists with matching content; recovering from prior partial reconcile", "clusterResourceOverride", croKObj, "newOverrideSnapshot", klog.KObj(newSnapshot))
+	} else {
+		klog.V(2).InfoS("Created new overrideSnapshot", "clusterResourceOverride", croKObj, "newOverrideSnapshot", klog.KObj(newSnapshot))
 	}
-	klog.V(2).InfoS("Created new overrideSnapshot", "clusterResourceOverride", croKObj, "newOverrideSnapshot", klog.KObj(newSnapshot))
-	return nil
+
+	// Demote the previous latest snapshot. Tolerate IsNotFound from a concurrent prune or parent
+	// deletion: in that case there is nothing to demote, but we still want to run the sibling
+	// audit below to clean up any stale latest=true left by earlier crashes.
+	if latestSnapshot != nil && latestSnapshot.Labels[placementv1beta1.IsLatestSnapshotLabel] == strconv.FormatBool(true) {
+		latestSnapshot.Labels[placementv1beta1.IsLatestSnapshotLabel] = strconv.FormatBool(false)
+		if err := r.Client.Update(ctx, latestSnapshot); err != nil {
+			if errors.IsNotFound(err) {
+				klog.V(2).InfoS("Old overrideSnapshot already gone; skipping demotion", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(latestSnapshot))
+			} else {
+				klog.ErrorS(err, "Failed to set the isLatestSnapshot label to false on previous snapshot", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(latestSnapshot))
+				return controller.NewUpdateIgnoreConflictError(err)
+			}
+		} else {
+			klog.V(2).InfoS("Marked previous overrideSnapshot as inactive", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(latestSnapshot))
+		}
+	}
+
+	// Audit older siblings for stale latest=true left by prior crashes. cleanupStaleLatestSiblings
+	// preserves the highest-index item in snapshotList (the previous latest, which we just demoted
+	// in-memory and on the server) and flips any older sibling still carrying latest=true.
+	return r.cleanupStaleLatestSiblings(ctx, snapshotList)
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *ClusterResourceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.recorder = mgr.GetEventRecorderFor("clusterresourceoverride-controller")
 	return ctrl.NewControllerManagedBy(mgr).
 		Named("clusterresourceoverride-controller").
 		For(&placementv1beta1.ClusterResourceOverride{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
--- a/pkg/controllers/overrider/clusterresource_controller.go
+++ b/pkg/controllers/overrider/clusterresource_controller.go
@@ -101,19 +101,16 @@ func (r *ClusterResourceReconciler) ensureClusterResourceOverrideSnapshot(ctx co
 		return err
 	}
 
-	latestSnapshotIndex := -1 // so index starts at 0
+	latestSnapshotIndex := -1 // index starts at 0
 	var latestSnapshot *placementv1beta1.ClusterResourceOverrideSnapshot
 	if len(snapshotList.Items) != 0 {
-		// Convert the last (highest-index) unstructured snapshot to the typed object.
 		latestSnapshot = &placementv1beta1.ClusterResourceOverrideSnapshot{}
 		if err = runtime.DefaultUnstructuredConverter.FromUnstructured(snapshotList.Items[len(snapshotList.Items)-1].Object, latestSnapshot); err != nil {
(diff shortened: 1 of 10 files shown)


ytimocin:
## E2E validation on Kind clusters

Ran four targeted edge-case tests against this branch on a fresh `make setup-clusters` (hub + 3 members).

### Setup
- `make setup-clusters MEMBER_CLUSTER_COUNT=3` (fresh build, hub-agent rev `cc4657694`)
- `make join-members`
- 4 Kind clusters on Apple Silicon (Mac), envtest already covers the in-process unit/integration paths

### Edge cases

| # | Test | Verifies | Result |
|---|------|----------|--------|
| 1 | Hub-agent restart mid-reconcile | Recovery from a crash during snapshot churn (the original bug scenario) | ✅ PASS |
| 2 | 10 concurrent CRO updates | Workqueue contention doesn't break invariants | ✅ PASS |
| 3 | Inject duplicate `latest=true` sibling | `cleanupStaleLatestSiblings` audit fires on next reconcile | ✅ PASS |
| 4 | RO same name in two namespaces | `(namespace, OverrideTrackingLabel)` dedup key isolates parents | ✅ PASS |

### Sample results

**Test 1** — after force-killing all hub-agent pods (`--grace-period=0 --force`) during 3 rapid CRO updates:
```
NAME                       LATEST   INDEX
edge-restart-cro-53101-0   false    0
edge-restart-cro-53101-1   false    1
edge-restart-cro-53101-2   true     2
```
Exactly one `latest=true` at the highest index — the create-first ordering recovered cleanly.

**Test 2** — 10 patches submitted in parallel via `&`:
```
edge-concurrent-cro-53313-1  false  1
edge-concurrent-cro-53313-10 true   10   ← single latest
edge-concurrent-cro-53313-2  false  2
…
edge-concurrent-cro-53313-9  false  9
```

**Test 3** — manually re-flipped snapshot 0 to `latest=true` to simulate a post-crash duplicate, then patched the CRO. Hub-agent log:
```
overrider/clusterresource_controller.go:187  "Created new overrideSnapshot"  newOverrideSnapshot="edge-dup-cro-53784-2"
overrider/clusterresource_controller.go:203  "Marked previous overrideSnapshot as inactive"  overrideSnapshot="edge-dup-cro-53784-1"
overrider/common.go:198                       "Flipped stale latest sibling to false"        overrideSnapshot="edge-dup-cro-53784-0"
```
The new audit helper fired exactly once on the injected duplicate.

**Test 4** — two `ResourceOverride`s named `edge-ro-shared-…` in two different namespaces. Updating one produced index-1 in that namespace only; the other namespace's RO snapshot stayed at index 0.

### Hub-agent log scan across all four runs

- `Created new overrideSnapshot`: 19 occurrences (create-first ordering live)
- `Marked previous overrideSnapshot as inactive`: 14 occurrences (demote-after live)
- `Flipped stale latest sibling to false`: 1 occurrence (audit helper from test 3)
- `OverrideSnapshotHashMismatch` events: 0
- `recovering from prior partial reconcile`: 0
- `Failed to create new overrideSnapshot`: 0
- `Failed to set the isLatestSnapshot label to false`: 0
- panics: 0

### Side observation

The override controller uses `predicate.GenerationChangedPredicate{}`, so reconcile only fires on **spec** changes — not annotation/label edits. This means a duplicate `latest=true` left behind by a crash will persist until either (a) the next CRO/RO spec change, or (b) the informer's full-list resync (6 h hub-agent default) emits a synthetic Create event. Read-time dedup in `FetchAllMatchingOverridesForResourceSnapshot` masks this in the meantime, so callers always see the correct authoritative snapshot — just worth noting that the on-disk cleanup is gated on the next reconcile, not wall-clock immediate.

### Baseline `make e2e-tests` caveat

The full suite under `-p` (9 parallel workers vs 4 Kind clusters on a single Mac) produced ~99 failures distributed uniformly across all placement test types. Failures were 10-second `Eventually()` timeouts — environmental, not regressions. No override-controller errors in hub-agent logs during the run. Should not affect CI/dedicated runners with proper resource isolation.

pkg/controllers/overrider/resource_controller.go
```diff
@@ -150,15 +147,58 @@ func (r *ResourceReconciler) ensureResourceOverrideSnapshot(ctx context.Context,
...
 	if err = r.Client.Create(ctx, newSnapshot); err != nil {
-		klog.ErrorS(err, "Failed to create new overrideSnapshot", "newOverrideSnapshot", klog.KObj(newSnapshot))
-		return controller.NewAPIServerError(false, err)
+		if !errors.IsAlreadyExists(err) {
+			klog.ErrorS(err, "Failed to create new overrideSnapshot", "newOverrideSnapshot", klog.KObj(newSnapshot))
+			return controller.NewAPIServerError(false, err)
+		}
+		// AlreadyExists here normally means a prior reconcile succeeded the Create but failed to
+		// flip the old snapshot. Verify the existing object's hash before proceeding so we don't
+		// silently promote stale content (etcd restore from backup, manual edit, future hash bug).
```
weng271190436: Have a question: why would "etcd restore from backup" result in AlreadyExists path?

What does "future hash bug" mean? Does it mean hash function change?

pkg/controllers/overrider/resource_controller.go
```diff
@@ -150,15 +147,58 @@ func (r *ResourceReconciler) ensureResourceOverrideSnapshot(ctx context.Context,
...
+		// Read through the uncached client: the cached read can lag the just-Created object.
+		existing := &placementv1beta1.ResourceOverrideSnapshot{}
+		if getErr := r.UncachedReader.Get(ctx, types.NamespacedName{Name: newSnapshot.Name, Namespace: newSnapshot.Namespace}, existing); getErr != nil {
+			klog.ErrorS(getErr, "Failed to get existing overrideSnapshot for hash verification", "resourceOverride", roKObj, "newOverrideSnapshot", klog.KObj(newSnapshot))
+			return controller.NewAPIServerError(false, getErr)
+		}
+		if string(existing.Spec.OverrideHash) != overrideSpecHash {
```
weng271190436, line 156: Maybe we can recompute overrideHash of existing to avoid situations where someone modifies the overrideHash?

pkg/controllers/overrider/resource_controller.go
```diff
@@ -150,15 +147,58 @@ func (r *ResourceReconciler) ensureResourceOverrideSnapshot(ctx context.Context,
...
+			mismatchErr := fmt.Errorf("existing overrideSnapshot %s/%s has hash %q, want %q", existing.Namespace, existing.Name, string(existing.Spec.OverrideHash), overrideSpecHash)
+			klog.ErrorS(mismatchErr, "Existing overrideSnapshot has different content than expected; will requeue and retry", "resourceOverride", roKObj)
+			// Surface to the operator via an Event on the parent RO; see CRO controller for rationale.
```
weng271190436: One idea: shall we delete the mismatched existing snapshot and recreate?

My reason:
I think override snapshot is "owned" by the the CRO/RO. A reconciler, by definition, should fix observed states which deviated from desired state. So it is conceptually correct to reconcile the snapshot, even if the snapshot is being used somehow. And even if the mismatched snapshot is being used somehow, I would argue that the current snapshot that is being applied is wrong (i.e. doesn't match desired state) anyway. And it should be reconciled.

pkg/controllers/overrider/resource_controller.go
```diff
@@ -82,54 +84,49 @@ func (r *ResourceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
-		if latestSnapshot.Labels[placementv1beta1.IsLatestSnapshotLabel] == strconv.FormatBool(true) {
-			// set the latest label to false first to make sure there is only one or none active policy snapshot
-			latestSnapshot.Labels[placementv1beta1.IsLatestSnapshotLabel] = strconv.FormatBool(false)
-			if err := r.Client.Update(ctx, latestSnapshot); err != nil {
-				klog.ErrorS(err, "Failed to set the isLatestSnapshot label to false", "ResourceOverride", croKObj, "overrideSnapshot", klog.KObj(latestSnapshot))
-				return controller.NewUpdateIgnoreConflictError(err)
+			// The content has not changed; no new snapshot is needed. Ensure the highest-index
+			// snapshot is marked latest, then audit siblings to clean up any duplicate latest=true
+			// left by a prior partial run.
+			if err := r.ensureSnapshotLatest(ctx, latestSnapshot); err != nil {
```
weng271190436, line 113: When would we need to mark the snapshot as latest?

If we always create a new snapshot with is-latest = true, the latest snapshot already has is-latest label set to true even if a crash happened right after a new snapshot was created?

Want to understand the scenario that you have in mind so that we can think about how to handle.

pkg/utils/overrider/overrider.go
```diff
@@ -248,6 +256,80 @@ func isClusterMatched(cluster *clusterv1beta1.MemberCluster, policy *placementv1
...
+			winnerIndex[key] = index
+		}
+	}
+	deduped := make([]T, 0, len(winners))
+	for _, idx := range winners {
+		deduped = append(deduped, items[idx])
+	}
+	// Map iteration order is non-deterministic; sort the result by namespace+name so callers
+	// (and tests) see a stable order. Cluster-scoped snapshots have an empty namespace, which
+	// sorts naturally before namespaced ones.
```
weng271190436: I think this comment is misleading "Cluster-scoped snapshots have an empty namespace, which sorts naturally before namespaced ones." because it is not possible for dedupLatestSnapshots to handle mixed-scope items? They would need different "keyOf func(PT) string". And you can only pass in one keyOf func per dedupLatestSnapshots call.

pkg/controllers/overrider/resource_controller.go
ytimocin: For the 1st question, I was trying to be extra defensive by thinking about a partial restore. And, for the 2nd question, yes, I was thinking about a hash function change in between versions.

pkg/controllers/overrider/resource_controller.go
ytimocin: **API Server** applies defaults on **Create**, and, that is why, the hashes would be different between **Create** and **re-hash of a read-back object**.

The defaults come from `kubebuilder:default` markers on the API types — that's what makes the original and read-back objects differ.

pkg/controllers/overrider/resource_controller.go
ytimocin: Good idea! The hash-mismatch branch now deletes the existing snapshot before returning the error — same change in both CRO and RO controllers, with the integration tests updated to assert the deletion.

pkg/controllers/overrider/resource_controller.go
ytimocin: **ensureSnapshotLatest** was already here before this PR — I just kept calling it from the hash-match path as defensive insurance against out-of-band label edits. It's a no-op in steady state. Want me to drop it as part of this PR, or leave it alone since it's pre-existing?

pkg/utils/overrider/overrider.go
ytimocin: Comment updated. Thanks for catching that.

pkg/controllers/overrider/common_test.go
```diff
@@ -225,6 +226,146 @@ var _ = Describe("Test ClusterResourceOverride common logic", func() {
...
+			err := aeReconciler.ensureClusterResourceOverrideSnapshot(ctx, aeCRO, 10)
+			Expect(err).Should(HaveOccurred(), "mismatched hash should not silently succeed")
+			Expect(errors.Is(err, controller.ErrExpectedBehavior)).Should(BeTrue(),
+				"error should wrap ErrExpectedBehavior so retry runs without stack-trace floods, got %v", err)
+
+			By("Verifying snapshot 0 was NOT demoted (we abort before the demote step)")
+			final0 := &placementv1beta1.ClusterResourceOverrideSnapshot{}
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: snapshot0.Name}, final0)).Should(Succeed())
+			Expect(final0.Labels[placementv1beta1.IsLatestSnapshotLabel]).Should(Equal("true"))
+		})
```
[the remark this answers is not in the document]
ytimocin: Addressed in 016efa2f.

pkg/utils/overrider/overrider_test.go
```diff
@@ -1491,10 +1650,17 @@ func TestFetchAllMatchingOverridesForResourceSnapshot(t *testing.T) {
 			for i := range tc.snapshots {
 				objects = append(objects, tc.snapshots[i])
 			}
+			// Production-realistic labels: every override snapshot is created with
+			// OverrideTrackingLabel and OverrideIndexLabel by the override controllers. The
+			// fixtures only declare the latest-snapshot label they care about, so we inject the
+			// rest here. Without these the read-time dedup path correctly errors on malformed
+			// labels.
```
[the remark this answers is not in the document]
ytimocin: Addressed in 016efa2f.

pkg/controllers/overrider/clusterresource_controller.go
```diff
@@ -99,37 +101,30 @@ func (r *ClusterResourceReconciler) ensureClusterResourceOverrideSnapshot(ctx co
 		return err
 	}
 
-	latestSnapshotIndex := -1 //so index start at 0
+	latestSnapshotIndex := -1 // index starts at 0
+	var latestSnapshot *placementv1beta1.ClusterResourceOverrideSnapshot
 	if len(snapshotList.Items) != 0 {
-		// convert the last unstructured snapshot to the typed object
-		latestSnapshot := &placementv1beta1.ClusterResourceOverrideSnapshot{}
+		latestSnapshot = &placementv1beta1.ClusterResourceOverrideSnapshot{}
```
michaelawyu, line 107: Hi Yetkin! Thanks so much for this. It is a very interesting PR and I know that such fixes take a long while and a lot of efforts to complete. It is just that, I couldn't help feeling that given the nature of the bug, i.e., rollout controller failing to catch the state change where a resource snapshot is marked as latest again, it can just be easily fixed by tweaking the handler funcs in the rollout controller a bit:

michaelawyu: The system just need an UpdateFunc that enqueues placement if it sees that for a snapshot update event, `(e.ObjectOld.GetLabels()[isLatestLabelName] != e.ObjectNew.GetLabels()[isLatestLabelName]) && e.ObjectNew.GetLabels()[isLatestLabelName] == "true"` (very roughly written).

pkg/controllers/overrider/clusterresource_controller.go
```diff
@@ -99,37 +101,30 @@ func (r *ClusterResourceReconciler) ensureClusterResourceOverrideSnapshot(ctx co
...
-		}
-		// mark the last policy snapshot as inactive if it is different from what we have now.
-		if latestSnapshot.Labels[placementv1beta1.IsLatestSnapshotLabel] == strconv.FormatBool(true) {
-			// set the latest label to false first to make sure there is only one or none active policy snapshot
-			latestSnapshot.Labels[placementv1beta1.IsLatestSnapshotLabel] = strconv.FormatBool(false)
-			if err := r.Client.Update(ctx, latestSnapshot); err != nil {
-				klog.ErrorS(err, "Failed to set the isLatestSnapshot label to false", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(latestSnapshot))
-				return controller.NewUpdateIgnoreConflictError(err)
+			// Hash matches: re-mark this snapshot latest (idempotent) and audit siblings.
+			if err := r.ensureSnapshotLatest(ctx, latestSnapshot); err != nil {
```
michaelawyu, line 114: This part is a bit interesting I think. If we do switch to the create-then-demote model, this part is a bit redundant (it is either always a no-op or always fail due to conflicts). And it cannot be reliably used to correct unexpected cases either (e.g., a latest label is set to false unexpectedly), as it would also trigger the issue we want to fix, i.e., the rollout controller missing the update.

pkg/controllers/overrider/clusterresource_controller.go
```diff
@@ -149,15 +144,64 @@ func (r *ClusterResourceReconciler) ensureClusterResourceOverrideSnapshot(ctx co
...
+		// differ from the input, so a recompute would never match.
+		if string(existing.Spec.OverrideHash) != overrideSpecHash {
+			mismatchErr := fmt.Errorf("existing overrideSnapshot %s has stored hash %q, want %q", newSnapshot.Name, string(existing.Spec.OverrideHash), overrideSpecHash)
+			klog.ErrorS(mismatchErr, "Existing overrideSnapshot has different content than the current spec; deleting it so the next reconcile can recreate with correct content", "clusterResourceOverride", croKObj)
+			// Event surfaces the recovery in `kubectl describe`.
+			r.recorder.Eventf(cro, corev1.EventTypeWarning, "OverrideSnapshotHashMismatch",
+				"existing snapshot %s has different content than the current spec; deleting it for the next reconcile to recreate. If this persists, investigate the snapshot for manual edits or a hash-function change between controller versions.", newSnapshot.Name)
+			// Drive observed → desired: delete the mismatched snapshot so the next reconcile
+			// recreates it with the correct content.
+			if delErr := r.Client.Delete(ctx, existing); delErr != nil && !errors.IsNotFound(delErr) {
```
michaelawyu, line 171: This deletion part is a bit problematic I fear.

Aside from the semantic concerns (the snapshots should have monotonically, continuous indexes), the snapshot we would like to delete might have already been referenced in some bindings.

michaelawyu: This might lead to the work generator getting stuck in a loop of errors when processing some bindings as it can no longer find a referenced snapshot. As our rollout (mostly rolling update) checks for Applied/Availability info, it might also lead a rollout to be stuck.

michaelawyu (commented):
Hi Yetkin! I added some comments; would be really happy to discuss about this issue further.

pkg/utils/overrider/overrider.go
```diff
@@ -248,6 +256,79 @@ func isClusterMatched(cluster *clusterv1beta1.MemberCluster, policy *placementv1
...
+// (namespace, OverrideTrackingLabel) composite because RO names can collide across namespaces.
+//
+// A snapshot whose tracking label is missing/empty or whose OverrideIndexLabel is unparsable is
+// logged as an error and skipped: this fetch runs in the rollout hot path, so failing the entire
+// fetch over a single corrupt snapshot would block override application for every placement.
+// The error log is loud enough for operators to notice; the affected override loses one update
+// cycle until the snapshot's labels are repaired.
+//
+// PT is *T constrained to client.Object so we can extract metadata and labels without copying.
+func dedupLatestSnapshots[T any, PT interface {
```
michaelawyu, line 271: Hi Yetkin! This part is particularly interesting I think. 

One of the major benefits that the demote then create pattern yields is that, it guarantees that at any given point of watches (with the cache lagging or not), the consumer (whoever needs to read/get notified on the appearance of a latest snapshot) will either see:

a) there is no latest snapshot -> which means the system observes an intermediate state and can simply wait for the next event; or
b) there is exactly one latest snapshot.

So the consumer can just do a list op with a label selector, and decide what to do based on the number of items returned. No additional sorting/ranking is needed.

michaelawyu: e.g,

0 (latest) - | -> 0 (demoted) - | -> 1 (latest) - | -> 1 (demoted) - | -> 2 (latest) - | -> ...

| = time point where the consumer reads data

michaelawyu: But with the create then demote pattern, 


e.g,

0 (latest) - | -> 1 (latest) - | -> 0 (demoted) - | -> 2 (latest) - | -> 1 (demoted) - | -> ...

Roughly (very roughly) 50% of the time, the consumer will see that there are multiple (2) latest snapshots. And it must do the sorting/ranking to determine the actual latest snapshot (or it can just back off until the situation is resolved by the snapshotting controller). The correctness part, as Yetkin you have acutely catches in this part of the code, can still be guaranteed, but it kinds of defeats the purpose of having the latest label. As the source of truth always lies in the monotonically increasing indexes, the consumer can simply just list every snapshot relevant and do the sorting/ranking every time to find the latest snapshot -> it's more costly, but more straightforward.

pkg/controllers/overrider/clusterresource_controller.go
```diff
@@ -149,15 +144,64 @@ func (r *ClusterResourceReconciler) ensureClusterResourceOverrideSnapshot(ctx co
...
+		// uncached reader — the cache can lag the just-Created object.
+		existing := &placementv1beta1.ClusterResourceOverrideSnapshot{}
+		if getErr := r.UncachedReader.Get(ctx, types.NamespacedName{Name: newSnapshot.Name}, existing); getErr != nil {
+			klog.ErrorS(getErr, "Failed to get existing overrideSnapshot for hash verification", "clusterResourceOverride", croKObj, "newOverrideSnapshot", klog.KObj(newSnapshot))
+			return controller.NewAPIServerError(false, getErr)
+		}
+		// Compare stored OverrideHash, not a re-hash of existing.Spec.OverrideSpec: API-server
+		// defaulting on Create (selectionScope, scope, overrideType) makes the read-back spec
+		// differ from the input, so a recompute would never match.
+		if string(existing.Spec.OverrideHash) != overrideSpecHash {
```
michaelawyu, line 163: This part is particularly tricky I assume (the flow itself, not specific to this PR)...

Generally speaking, if the user makes different edits each time (A -> B -> C -> ...) and we do not consider hash collision an issue, then this branch will never run and the issue of a past snapshot being marked as latest again will never occur.

The problem will only happen when the user does an A -> B -> A fashion of edits:

michaelawyu: In the case of a demote then create flow, there are two possibilities:

case 1 => 

there is a snapshot 1 with hash A -> the user edits overrides that has hash B -> the controller is triggered -> the controller has snapshot 1 demoted, but failed to create snapshot 2 with hash B -> the user edits overrides back to the state of hash A -> the controller is triggered again and marked snapshot 1 as latest

case 1 is harmless, as it still aligns with our eventually consistent model; the system is just behaving as if the edit of hash B never happened. There is no rollout needed either.

case 2 =>

there is a snapshot 1 with hash A -> the user edits overrides that has hash B -> the controller is triggered -> the controller has snapshot 1 demoted, and created snapshot 2 with hash B -> the user edits overrides back to the state of hash A -> the controller is triggered again, but due to cache lagging, it just sees snapshot 1 in its demoted state, snapshot 2 hasn't appeared yet -> the controller marks snapshot 1 as latest again -> the system is trapped in an erred state with two latest snapshots

case 2 is harmful; the consumer side needs to handle this by doing its own sorting/ranking, and the producer side needs to fix the situation in later reconciliations.

michaelawyu: And this is not a demote then create flow specific problem -> create then demote can have this as well. And this PR has fixed this by adding additional logic to clean up the history and setting the consumer to de-dup as needed.

But these occurrences are rare in nature -> if we bring object generation into the hash calculation the chances would get even more slim (to the brim of being impossible). Instead of having the cleanup logic and de-duppiing run often, I wonder if it would actually be easier to just:

a) on the consumer side, list with latest label selector, and back off if there are multiple latest snapshots -> wait for the producer side to fix things
b) on the producer side, hash with object generation embedded, and in very rare cases where multiple latest snapshots do appear, do the cleanup.

michaelawyu: Important to note that any change we make must guarantee that the new flow won't trigger rollouts on existing placements, so the hashing part might not be easily editable.

pkg/controllers/overrider/clusterresource_controller.go
michaelawyu: (There is actually a bit more to this bug -> please see some of the other comments, happy to discuss about this a bit more and make sure that I understand it correctly)

ytimocin:
We will get back to this later. cc/ @michaelawyu

PR #703, opened 2026-05-11
fix: add securityContext to hub-agent and member-agent deployment

jwtty:

Hardening the securityContexts of hub-agent and member-agent deployments.
The webhook cert mount path needs a mount too as we now have `readOnlyRootFileSystem=true`.

Fixes #

I have:

### DIFF
--- a/charts/member-agent/templates/deployment.yaml
+++ b/charts/member-agent/templates/deployment.yaml
@@ -16,10 +16,23 @@ spec:
     spec:
       restartPolicy: Always
       serviceAccountName: {{ include "member-agent.fullname" . }}-sa
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65532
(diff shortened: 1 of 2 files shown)


charts/member-agent/templates/deployment.yaml
```diff
@@ -16,10 +16,23 @@ spec:
     spec:
       restartPolicy: Always
       serviceAccountName: {{ include "member-agent.fullname" . }}-sa
+      securityContext:
```
ytimocin, line 19: I know we set the user in the Dockerfiles we have but what is the difference between setting others here and a Dockerfile?

jwtty: It's like a double-enforcement. In case the image is changed, the deployment can still enforece the non-root user.

britaniar (approved):

PR #706, opened 2026-05-13
fix: address a RO directory issue in templates

michaelawyu:

This PR addresses an issue which will cause agent to fail upon startup when using self-signed certs with the newly improved security context setup.

I have:

- [NA] Associated this change with a known KubeFleet Issue (Bug, Feature, etc). -> Related to #703 .
- [*] Run `make reviewable` to ensure this PR is ready for review.

NA

NA

### DIFF
--- a/charts/hub-agent/templates/deployment.yaml
+++ b/charts/hub-agent/templates/deployment.yaml
@@ -26,6 +26,15 @@ spec:
         fsGroup: 65532
         seccompProfile:
           type: RuntimeDefault
+      initContainers:
+        {{- if .Values.crdInstaller.enabled }}
+        - name: crd-installer
+          image: "{{ .Values.crdInstaller.image.repository }}:{{ .Values.crdInstaller.image.tag }}"
+          imagePullPolicy: {{ .Values.crdInstaller.image.pullPolicy }}
+          args:
+            - --mode=hub
+            - --v={{ .Values.crdInstaller.logVerbosity }}
+        {{- end }}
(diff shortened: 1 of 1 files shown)


charts/hub-agent/templates/deployment.yaml
```diff
@@ -90,13 +90,21 @@ spec:
                 fieldPath: metadata.namespace
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          {{- if .Values.useCertManager }}
           volumeMounts:
           - name: webhook-cert
             # This path must match FleetWebhookCertDir in pkg/webhook/webhook.go
             mountPath: /tmp/k8s-webhook-server/serving-certs
-            {{- if .Values.useCertManager }}
             readOnly: true
```
Arvindthiru, line 98: I'm assuming since we certManager which generates certs in this flow we make this volume read only and when we don't since we clear the directory and write the certs, we can't have the readOnly field down below ?

Arvindthiru (approved):

charts/hub-agent/templates/deployment.yaml
michaelawyu: Yeah, it seems that the cause of this issue is that, with self-signed certs, we would always need write access. The `readOnly` field was set to `false` (the default) originally; the tricky part is about the mount path 😞

michaelawyu:
Merging this to unblock progress -> if there's any concern, please let me know.

Issue #707, opened 2026-05-13
[Feature] Use validating admission policies for applicable validating logic


michaelawyu:
**Is your feature request related to a problem? Please describe.**
KubeFleet should switch to validating admission policies for easier validation. It addresses many of the issues webhooks might have, incl. increased overhead/latencies, and limited reliability.

**Describe the solution you'd like**
Move applicable validating logic to VAPs.

**Describe alternatives you've considered**
NA

**Additional context**
NA

PR #708, opened 2026-05-13
feat: migrating to validating admission policies for our validating logic (1/)

michaelawyu:

This PR is the first of many following PRs to migrate to validating admission policies for applicable validation logic.

Specifically, this PR:

* adds a admission policy manager for managing admission policies when the agent starts up. This manager is best used for environments where policies cannot be easily deployed/refreshed as part of the agent Helm charts.
* adds two policy generator: one for denying writes to pods + replica sets in non-reserved namespaces, the other one for denying writes to service accounts and creating token requests in reserved namespaces.

Releated to #707.

I have:

- [*] Associated this change with a known KubeFleet Issue (Bug, Feature, etc).
- [*] Run `make reviewable` to ensure this PR is ready for review.

- [*] Integration tests


michaelawyu:
Please note that for this PR the manager is not enabled in the hub agent yet.

To control the PR size, additional tests will be submitted as a separate PR.

pkg/admissionpolicymanager/svcaccountsntokenreqs.go
```diff
@@ -0,0 +1,200 @@
...
+		Spec: admissionregistrationv1.ValidatingAdmissionPolicySpec{
+			FailurePolicy: ptr.To(admissionregistrationv1.Fail),
+			MatchConstraints: &admissionregistrationv1.MatchResources{
+				ResourceRules: []admissionregistrationv1.NamedRuleWithOperations{
+					{
+						RuleWithOperations: admissionregistrationv1.RuleWithOperations{
+							Operations: []admissionregistrationv1.OperationType{
+								admissionregistrationv1.Create,
+								admissionregistrationv1.Update,
+								admissionregistrationv1.Delete,
```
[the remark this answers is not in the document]
michaelawyu: Fixed.

pkg/admissionpolicymanager/svcaccountsntokenreqs.go
```diff
@@ -0,0 +1,200 @@
...
+
+	celExprAcc := strings.Join(celExprAccSegs, " || ")
+
+	celExprNSSegs := []string{}
+	for _, prefix := range g.ReservedNamespacePrefixes {
+		celExprNSSegs = append(celExprNSSegs, fmt.Sprintf(`object.metadata.namespace.startsWith("%s")`, prefix))
+	}
+	celExprNS := strings.Join(celExprNSSegs, " || ")
+
+	celExpr := fmt.Sprintf("!(%s) || (%s)", celExprNS, celExprAcc)
```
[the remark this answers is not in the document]
michaelawyu: Fixed.

pkg/admissionpolicymanager/podsnreplicasets.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// Policies generates a ValidatingAdmissionPolicy that denies creation of pods and
+// replicasets in non-reserved namespaces.
+func (g *PodsAndReplicaSetsValidatingAdmissionPolicyGenerator) Policies() []*admissionregistrationv1.ValidatingAdmissionPolicy {
+	celExprSegs := []string{}
+	for _, prefix := range g.ReservedNamespacePrefixes {
+		celExprSegs = append(celExprSegs, fmt.Sprintf(`object.metadata.namespace.startsWith("%s")`, prefix))
+	}
+	celExpr := strings.Join(celExprSegs, " || ")
```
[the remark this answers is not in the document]
michaelawyu: Added a note for this. It is assumed that Validate() has been called before Policies() is called.

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,332 @@
...
+				})
+				if err != nil {
+					return errors.NewAPIServerError(err,
+						"failed to create/update validating admission policy",
+						false,
+						"op", opRes, "policyName", policy.Name, "policyGenerator", gen.Name())
+				}
+				return nil
+			})
+			if err != nil {
```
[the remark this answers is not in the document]
michaelawyu: Fixed.

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,332 @@
...
+)
+
+var (
+	// A list of all available policy generators.
+	AllGenerators = sets.Set[string]{
+		PodsAndReplicaSetsVAPGeneratorName:          {},
+		SvcAccountsAndTokenRequestsVAPGeneratorName: {},
+	}
+)
+
```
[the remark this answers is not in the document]
michaelawyu: It's supposed to be used only internally but I guess there is no hurt adding an extra layer. Fixed.

pkg/admissionpolicymanager/manager_integration_test.go
```diff
@@ -0,0 +1,265 @@
...
+			); diff != "" {
+				return fmt.Errorf("policy bindings mismatch (-got +want):\n%s", diff)
+			}
+			return nil
+		}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Received unexpected list of policy bindings or an error has occurred")
+	})
+
+	// Note: in the integration test environment, it appears that validating admission policies have no effect or are
+	// not being enforced as expected. As a result the effects are validated using E2E tests instead.
+})
```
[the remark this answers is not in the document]
michaelawyu: N/A -> as the first PR the integration hasn't been done yet and E2E is not yet possible.

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,332 @@
...
+	}, func() error {
+		if err := m.Client.List(ctx, existingPolicyBindingList, managedByAndPartOfKubeFleetLabelSelector); err != nil {
+			return errors.NewAPIServerError(err, "failed to list all validating admission policy bindings managed by KubeFleet", false)
+		}
+		return nil
+	})
+	if err != nil {
+		// No need to wrap this for another time. The inner error already contains sufficient context about the failure.
+		return err
+	}
```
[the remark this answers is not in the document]
michaelawyu: It is assumed that the manager should manage all VAPs; but for the clarity I will add a label.

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,332 @@
...
+)
+
+const (
+	// The following labels are added to all policies created by the policy manager,
+	// so that the agent can track the lifecycle of created policies across different runs
+	// and act accordingly.
+	VAPManagedByKubeFleetLabelKey   = "app.kubernetes.io/managed-by"
+	VAPManagedByKubeFleetLabelValue = "fleet"
+	VAPPartOfKubeFleetLabelKey      = "app.kubernetes.io/part-of"
+	VAPPartOfKubeFleetLabelValue    = "fleet"
```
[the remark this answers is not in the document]
michaelawyu: Done. Cannot use kubefleet due to the presence of the managed solution. Used fleet for simplicity reasons. The managed by value has been updated to use `fleet-hub-agent`.

pkg/utils/common.go
```diff
@@ -49,13 +49,13 @@ import (
...
-	NamespaceNameFormat        = fleetMemberNamespacePrefix + "%s"
-	RoleNameFormat             = fleetPrefix + "role-%s"
-	RoleBindingNameFormat      = fleetPrefix + "rolebinding-%s"
+	KubePrefix                 = "kube-"
+	FleetPrefix                = "fleet-"
+	FleetMemberNamespacePrefix = FleetPrefix + "member-"
+	FleetSystemNamespace       = FleetPrefix + "system"
+	NamespaceNameFormat        = FleetMemberNamespacePrefix + "%s"
+	RoleNameFormat             = FleetPrefix + "role-%s"
+	RoleBindingNameFormat      = FleetPrefix + "rolebinding-%s"
```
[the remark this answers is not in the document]
michaelawyu: Done.

pkg/admissionpolicymanager/svcaccountsntokenreqs.go
```diff
@@ -0,0 +1,200 @@
...
+
```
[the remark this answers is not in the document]
michaelawyu: Note: this restriction is added in consistency with our existing webhook setup. `system:masters` is supposed to have unlimited access to all resources BTW.

pkg/admissionpolicymanager/podsnreplicasets.go
```diff
@@ -0,0 +1,145 @@
...
+						},
+					},
+				},
+			},
+			Validations: []admissionregistrationv1.Validation{
+				{
+					Expression: celExpr,
+					// The error message has been set to match with the checks in some of the E2E tests
+					// in our existing release pipeline.
+					Message: "creating pods and replicas is disallowed in the fleet hub cluster",
```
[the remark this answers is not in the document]
michaelawyu: N/A: as mentioned, the error msg is explicitly kept in the same style as some of the existing checks for releasing purposes.

pkg/admissionpolicymanager/podsnreplicasets.go
```diff
@@ -0,0 +1,145 @@
...
+	PodsAndReplicaSetsVAPGeneratorName = "DenyPodsAndReplicaSetsOutsideReservedNamespaces"
+)
+
+const (
+	podsAndReplicaSetsVAPPolicyName        = "deny-pods-and-replicasets-outside-reserved-namespaces"
+	podsAndReplicaSetsVAPPolicyBindingName = "deny-pods-and-replicasets-outside-reserved-namespaces-binding"
+)
+
+// reservedNamespacePrefixRegExp matches valid namespace prefix characters (DNS label subset).
+var reservedNamespacePrefixRegExp = regexp.MustCompile(`^[a-z0-9-]+$`)
```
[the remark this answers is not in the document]
michaelawyu: Done. No preference on this part -> to me it's Regular Expression, hence RegExp 😂

pkg/admissionpolicymanager/manager_integration_test.go
```diff
@@ -0,0 +1,265 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
```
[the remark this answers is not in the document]
michaelawyu: Done.

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+	}
+
+	if len(enabledPolicyNameSet) != len(enabledPolicyGenerators) {
+		configuredPolicyNames := make([]string, 0, len(enabledPolicyGenerators))
+		for name := range enabledPolicyGenerators {
+			configuredPolicyNames = append(configuredPolicyNames, name)
+		}
+		return nil, errors.NewUserError(nil, "some enabled policy generators are not configured properly", "enabledPolicies", enabledPolicyNames, "configuredPolicies", configuredPolicyNames)
+	}
```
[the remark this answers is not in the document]
michaelawyu: Ah... The log has already includes the enabled names and configured names. Doing a set diff really isn't necessary.

pkg/admissionpolicymanager/podsnreplicasets.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return nil
+}
+
+// Policies generates a ValidatingAdmissionPolicy that denies creation of pods and
+// replicasets in non-reserved namespaces.
+func (g *PodsAndReplicaSetsValidatingAdmissionPolicyGenerator) Policies() []*admissionregistrationv1.ValidatingAdmissionPolicy {
+	celExprSegs := []string{}
+	for _, prefix := range g.ReservedNamespacePrefixes {
+		celExprSegs = append(celExprSegs, fmt.Sprintf(`object.metadata.namespace.startsWith("%s")`, prefix))
```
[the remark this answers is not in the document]
michaelawyu: Done.

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,366 @@
...
+	// A list of all available policy generators.
+	allGenerators = sets.Set[string]{
+		PodsAndReplicaSetsVAPGeneratorName:          {},
+		SvcAccountsAndTokenRequestsVAPGeneratorName: {},
+	}
+)
+
+// AllGenerators returns a copy of all available policy generators.
+func AllGenerators() sets.Set[string] {
+	return allGenerators.Clone()
```
weng271190436, line 71: Can you get all generator names by reflecting over DefaultPolicyGeneratorConfigs the same way preparePolicyGenerators do?

Keeping them in sync can prevent people adding new ValidatingAdmissionPolicyGenerator but forget to add to one of allGenerators or DefaultPolicyGeneratorConfigs

pkg/admissionpolicymanager/podsnreplicasets.go
```diff
@@ -0,0 +1,147 @@
...
+	PodsAndReplicaSetsVAPGeneratorName = "DenyPodsAndReplicaSetsOutsideReservedNamespaces"
+)
+
+const (
+	podsAndReplicaSetsVAPPolicyName        = "deny-pods-and-replicasets-outside-reserved-namespaces"
+	podsAndReplicaSetsVAPPolicyBindingName = "deny-pods-and-replicasets-outside-reserved-namespaces-binding"
+)
+
+// reservedNamespacePrefixRegexp matches valid namespace prefix characters (DNS label subset).
+var reservedNamespacePrefixRegexp = regexp.MustCompile(`^[a-z0-9-]+$`)
```
weng271190436: This should be in common.go (or another common file that doesn't belong to any specific VAP Generator) because it is used by both podsnreplicasets.go and svcaccountsntokenreqs.go?

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,366 @@
...
+		Duration: 1 * time.Second,
+		Factor:   2.0,
+		Jitter:   0.1,
+	}
+)
+
+type ValidatingAdmissionPolicyGenerator interface {
+	Name() string
+	Validate() error
+	Policies() []*admissionregistrationv1.ValidatingAdmissionPolicy
```
weng271190436: Instead of independent Policies and PolicyBindings members, can we do

```go
type PolicyWithBindings struct {
    Policy   admissionregistrationv1.ValidatingAdmissionPolicy
    Bindings []admissionregistrationv1.ValidatingAdmissionPolicyBinding
}

type ValidatingAdmissionPolicyGenerator interface {
	Name() string
	Validate() error
    PoliciesWithBindings() []PolicyWithBindings
}
```

?

The concern is that right now the relationship between Policy and Bindings is not explicit and is more error-proned (e.g. someone created a Policy but no Binding; someone deleted a Policy but forgot to delete the corresponding Bindings)

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,366 @@
...
+		policies := gen.Policies()
+		policyBindings := gen.PolicyBindings()
+
+		for _, policy := range policies {
+			// Add the managed by and part of labels to the policy, so that the agent can track the
+			// lifecycle of created policies across different runs and act accordingly.
+			if policy.Labels == nil {
+				policy.Labels = make(map[string]string)
+			}
+			policy.Labels[VAPManagedByKubeFleetLabelKey] = VAPManagedByKubeFleetLabelValue
```
weng271190436: If you define a map like this on top of file

```go
var AdmissionPolicyManagerLabels = map[string]string{
    VAPManagedByKubeFleetLabelKey:  VAPManagedByKubeFleetLabelValue,
    VAPPartOfKubeFleetLabelKey:     VAPPartOfKubeFleetLabelValue,
    VAPComponentKubeFleetLabelKey:  VAPComponentAdmissionPolicyManagerLabelValue,
}
```

this can be replaced with

```go
maps.Copy(policy.Labels, AdmissionPolicyManagerLabels)
```

can do similar things to 2 other places in this file since these 3 labels and values are always used together for all of AdmissionPolicyManager usages

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,366 @@
...
+			klog.V(2).InfoS("Successfully created or updated validating admission policy", "policyName", policy.Name, "policyGenerator", gen.Name())
+		}
+
+		for _, policyBinding := range policyBindings {
+			// Add the managed by and part of labels to the policy binding, so that the agent can track the
+			// lifecycle of created policy bindings across different runs and act accordingly.
+			if policyBinding.Labels == nil {
+				policyBinding.Labels = make(map[string]string)
+			}
+			policyBinding.Labels[VAPManagedByKubeFleetLabelKey] = VAPManagedByKubeFleetLabelValue
```
weng271190436: If you define a map like this on top of file

```go
var AdmissionPolicyManagerLabels = map[string]string{
    VAPManagedByKubeFleetLabelKey:  VAPManagedByKubeFleetLabelValue,
    VAPPartOfKubeFleetLabelKey:     VAPPartOfKubeFleetLabelValue,
    VAPComponentKubeFleetLabelKey:  VAPComponentAdmissionPolicyManagerLabelValue,
}
```

this can be replaced with

```go
maps.Copy(policyBinding.Labels, AdmissionPolicyManagerLabels)
```

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	return createdOrUpdatedPolicyNames, createdOrUpdatedPolicyBindingNames, nil
+}
+
+func (m *PolicyManager) garbageCollectUnusedPoliciesAndBindings(ctx context.Context, createdOrUpdatedPolicyNames sets.Set[string], createdOrUpdatedPolicyBindingNames sets.Set[string]) error {
+	// List all existing policies and policy bindings created by the manager.
+	existingPolicyList := &admissionregistrationv1.ValidatingAdmissionPolicyList{}
+	existingPolicyBindingList := &admissionregistrationv1.ValidatingAdmissionPolicyBindingList{}
+	managedByAndPartOfKubeFleetLabelSelector := client.MatchingLabels{
```
weng271190436:
```go
managedByAndPartOfKubeFleetLabelSelector := client.MatchingLabels(AdmissionPolicyManagerLabels)
```

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,366 @@
...
+			policy.Labels[VAPPartOfKubeFleetLabelKey] = VAPPartOfKubeFleetLabelValue
+			policy.Labels[VAPComponentKubeFleetLabelKey] = VAPComponentAdmissionPolicyManagerLabelValue
+
+			policyToCreateOrUpdate := &admissionregistrationv1.ValidatingAdmissionPolicy{
+				ObjectMeta: policy.ObjectMeta,
+			}
+			err := retry.OnError(policyRWOpBackoff, func(err error) bool {
+				// Retry on any error expect for context cancellation. Note that nil errors are not passed to this function,
+				// and AlreadyExists errors will not occur.
+				if ctx.Err() != nil {
```
weng271190436:
```go
func retryOnAnyErrorUnlessCancelled(ctx context.Context) func(error) bool {
    return func(err error) bool {
        return ctx.Err() == nil
    }
}
```

and then replace 6 of these usages with

```go
retry.OnError(policyRWOpBackoff, retryOnAnyErrorUnlessCancelled(ctx), func() error {
    // ...
})
```

?

pkg/admissionpolicymanager/manager.go
```diff
@@ -0,0 +1,366 @@
...
+// policies and their bindings for all enabled generators, and returns the names of
+// created or updated policies and policy bindings.
+func (m *PolicyManager) createOrUpdatePoliciesAndBindingsForEnabledGenerators(ctx context.Context) (sets.Set[string], sets.Set[string], error) {
+	createdOrUpdatedPolicyNames := sets.New[string]()
+	createdOrUpdatedPolicyBindingNames := sets.New[string]()
+
+	for _, gen := range m.enabledPolicyGenerators {
+		// As a sanity check, do one more round of validation.
+		//
+		// Normally this check would never fail as the generators have been validated before
```
weng271190436, line 173: This comment seems to be suggesting that in manager initialization (i.e. `New` method), the Validate methods were called. But they were not called in `New`?

pkg/admissionpolicymanager/podsnreplicasets.go
```diff
@@ -0,0 +1,147 @@
...
+				ResourceRules: []admissionregistrationv1.NamedRuleWithOperations{
+					{
+						RuleWithOperations: admissionregistrationv1.RuleWithOperations{
+							Operations: []admissionregistrationv1.OperationType{
+								admissionregistrationv1.Create,
+							},
+							Rule: admissionregistrationv1.Rule{
+								APIGroups:   []string{""},
+								APIVersions: []string{"v1"},
+								Resources:   []string{"pods"},
```
weng271190436, line 99: We have constants for "pods" and "replicasets" in webhook.go. Shall we start sharing them across webhook package and this new package by pulling them out into a common package? And eventually webhook.go will be deleted in one PR together with other webhook logic? Want to make sure that we still have these as constants after replacing webhook by VAP.

pkg/admissionpolicymanager/svcaccountsntokenreqs.go
```diff
@@ -0,0 +1,203 @@
...
+	// Check if any of the prefixes includes illegal characters.
+	for _, prefix := range g.ReservedNamespacePrefixes {
+		if !reservedNamespacePrefixRegexp.MatchString(prefix) {
+			return errors.NewUserError(nil, "prefix contains illegal characters; only lowercase alphanumeric characters and hyphens are allowed", "prefix", prefix)
+		}
+	}
+	// Check if any whitelisted username or user group contains characters that are
+	// illegal in a CEL string literal (\ and ").
+	for _, username := range g.WhitelistedUsernames {
+		if strings.ContainsAny(username, `"\`) {
```
weng271190436: These 2 illegal characters are applicable to all CELs, so a similar check needs to be done for many more future VAP generators? Maybe extract into a function so that it is more discoverable for future generators?

```go
// ValidateCELStringLiterals checks that none of the values contain characters
// that would break or inject into a CEL double-quoted string literal.
// The fieldName is used in the error message for context.
func ValidateCELStringLiterals(values []string, fieldName string) error {
    for _, v := range values {
        if strings.ContainsAny(v, `"\`) {
            return errors.NewUserError(nil,
                fmt.Sprintf("%s contains characters illegal in a CEL string literal", fieldName),
                fieldName, v)
        }
    }
    return nil
}
```

weng271190436: So these 2 become:

```go
    if err := ValidateCELStringLiterals(g.WhitelistedUsernames, "whitelisted username"); err != nil {
        return err
    }
    if err := ValidateCELStringLiterals(g.WhitelistedUserGroups, "whitelisted user group"); err != nil {
        return err
    }
```

pkg/admissionpolicymanager/svcaccountsntokenreqs.go
```diff
@@ -0,0 +1,203 @@
...
+// and creation of token requests in reserved namespaces, except for requests from certain
+// whitelisted users and user groups.
+//
+// For simplicity reasons, the code here assumes that the generator has been validated before Policies() is called.
+func (g *ServiceAccountsAndTokenRequestsValidatingAdmissionPolicyGenerator) Policies() []*admissionregistrationv1.ValidatingAdmissionPolicy {
+	celExprAccSegs := []string{}
+
+	// Exempt whitelisted users from this admission policy.
+	for _, username := range g.WhitelistedUsernames {
+		celExprAccSegs = append(celExprAccSegs, fmt.Sprintf(`request.userInfo.username == "%s"`, username))
```
weng271190436, line 106: Have a concern that this CEL construction code is not very readable. Might be difficult for someone to spot a mistake if there is any.

Wondering if it is worth investing in a celbuilder.go like this?

```go
// CELExpr represents a CEL expression fragment.
type CELExpr string

// String returns the CEL expression as a string.
func (e CELExpr) String() string {
    return string(e)
}

// UsernameEquals returns a CEL expression matching a specific username.
func UsernameEquals(username string) CELExpr {
    return CELExpr(fmt.Sprintf(`request.userInfo.username == "%s"`, username))
}

// InGroup returns a CEL expression matching membership in a group.
func InGroup(group string) CELExpr {
    return CELExpr(fmt.Sprintf(`"%s" in request.userInfo.groups`, group))
}

// NamespaceStartsWith returns a CEL expression matching a namespace prefix.
func NamespaceStartsWith(prefix string) CELExpr {
    return CELExpr(fmt.Sprintf(`request.namespace.startsWith("%s")`, prefix))
}

// Or joins expressions with ||.
func Or(exprs ...CELExpr) CELExpr {
    parts := make([]string, len(exprs))
    for i, e := range exprs {
        parts[i] = string(e)
    }
    return CELExpr(strings.Join(parts, " || "))
}

// Not negates an expression.
func Not(expr CELExpr) CELExpr {
    return CELExpr(fmt.Sprintf("!(%s)", expr))
}
```

then we can build and use CEL like this:

```go
func (g *SvcAccountsAndTokenRequestsVAPGenerator) Policies() []*admissionregistrationv1.ValidatingAdmissionPolicy {
    // Namespace condition
    var nsClauses []CELExpr
    for _, prefix := range g.ReservedNamespacePrefixes {
        nsClauses = append(nsClauses, NamespaceStartsWith(prefix))
    }

    // User exemptions
    var exemptions []CELExpr
    for _, u := range g.WhitelistedUsernames {
        exemptions = append(exemptions, UsernameEquals(u))
    }
    for _, grp := range g.WhitelistedUserGroups {
        exemptions = append(exemptions, InGroup(grp))
    }
    exemptions = append(exemptions,
        UsernameEquals(kubeSchedulerUserName),
        UsernameEquals(kubeControllerManagerUserName),
        InGroup(kubeNodeUserGroup),
        InGroup(adminUserGroup),
    )

    // Allow if: not reserved namespace OR user is exempt
    celExpr := Or(Not(Or(nsClauses...)), Or(exemptions...))

    policy := &admissionregistrationv1.ValidatingAdmissionPolicy{
        // ...
        Spec: admissionregistrationv1.ValidatingAdmissionPolicySpec{
            Validations: []admissionregistrationv1.Validation{
                {
                    Expression: celExpr.String(),
                    Message:    "writing service accounts in reserved namespaces or requesting tokens from such service accounts is disallowed",
                    Reason:     ptr.To(metav1.StatusReasonForbidden),
                },
            },
        },
    }
    return []*admissionregistrationv1.ValidatingAdmissionPolicy{policy}
}
```

weng271190436: podsnreplicasets.go can also do similar:

```go
func (g *PodsAndReplicaSetsVAPGenerator) Policies() []*admissionregistrationv1.ValidatingAdmissionPolicy {
    var nsClauses []CELExpr
    for _, prefix := range g.ReservedNamespacePrefixes {
        nsClauses = append(nsClauses, NamespaceStartsWith(prefix))
    }

    celExpr := Not(Or(nsClauses...))

    policy := &admissionregistrationv1.ValidatingAdmissionPolicy{
        // ...
        Spec: admissionregistrationv1.ValidatingAdmissionPolicySpec{
            Validations: []admissionregistrationv1.Validation{
                {
                    Expression: celExpr.String(),
                    Message:    "creating pods and replicas is disallowed in the fleet hub cluster",
                    Reason:     ptr.To(metav1.StatusReasonForbidden),
                },
            },
        },
    }
    return []*admissionregistrationv1.ValidatingAdmissionPolicy{policy}
}
```

pkg/admissionpolicymanager/manager.go
michaelawyu: Done ✔️

pkg/admissionpolicymanager/podsnreplicasets.go
michaelawyu: Done ✔️

pkg/admissionpolicymanager/manager.go
michaelawyu: Yeah, I guess... I do not particularly have a strong opinion on this; to me personally it is really an internal implementation that are more guarded by various checks and tests rather than structural limitations, not to say that even with this limitation it still does not guarantee any form of correctness (one can still write misconfigurations). But it is true that it doesn't hurt either.

pkg/admissionpolicymanager/manager.go
michaelawyu: Ah, let me add a utility function for this. `maps.Copy()` does not handle nils I think.

pkg/admissionpolicymanager/manager.go
michaelawyu: Done.

pkg/admissionpolicymanager/manager.go
michaelawyu: Done.

pkg/admissionpolicymanager/manager.go
michaelawyu: Hi Wei! A bit context on this: the validation will be done on the CLI handling code (as with other CLI args, since we added CLI validation earlier), which is why they are not called in `New()` again.

pkg/admissionpolicymanager/podsnreplicasets.go
michaelawyu: Hi Wei! For this one, I didn't set the code to share variables as the plan is to drop certain webhooks after VAPs become fully available (not all of them can be replaced by VAPs I fear). At this moment they are both present for release safety reasons.

pkg/admissionpolicymanager/svcaccountsntokenreqs.go
michaelawyu: Hi Wei! ~For this one, I have some reservations; it's only saving one line of code at the expense of another indirect all (with field names being passed in) -> I fear that it might not be much beneficial. I've moved the illegal chars to a const though.~

NVM, let me move this to a utility function.

pkg/admissionpolicymanager/svcaccountsntokenreqs.go
michaelawyu: Hi Wei! Thanks for this 🙏 TBH I was thinking a bit about roughly the same thing when the code was gen'd, but the concern back then was that, to add proper support, we would need to add operators like `Or`, `And`, `Not`, plus brackets. In the end it makes things more complicated but does not really help with the readability much.

For example, the one we use with service accounts and tokens would become:

```
Or(Not(Or(ExprA, ExprB, ...)), Bracket(Or(ExprX, ExprY, ...)))
```

and it feels a bit even more difficult to reason about?

michaelawyu: Considering this is part of the internal code (i.e., not public-facing API), I feel that it might be better not to over-complicate things (not to say that the current organization is perfect, that's for sure)?

Plus, LLMs are pretty good at generating the truth tables for CEL expressions; it might be easier to just let it validate. I will add some UTs in future PRs to validate the CEL expressions generated just to be extra safe.

pkg/admissionpolicymanager/podsnreplicasets.go
weng271190436: If we are going to keep webhook.go forever I think it is important to share the constants across webhook and admissionpolicymanager so that they are not being kept in 2 places?

pkg/admissionpolicymanager/svcaccountsntokenreqs.go
weng271190436: I still think CEL expression wrappers make it easier for human to read. Regarding the LLM argument, my concern is that, when LLM generated code caused issue, humans are still being held accountable (not just the person who created the PR, but also the PR reviewer).

```go
    var nsClauses []CELExpr
    for _, prefix := range g.ReservedNamespacePrefixes {
        nsClauses = append(nsClauses, NamespaceStartsWith(prefix))
    }

    // User exemptions
    var exemptions []CELExpr
    for _, u := range g.WhitelistedUsernames {
        exemptions = append(exemptions, UsernameEquals(u))
    }
    for _, grp := range g.WhitelistedUserGroups {
        exemptions = append(exemptions, InGroup(grp))
    }
    exemptions = append(exemptions,
        UsernameEquals(kubeSchedulerUserName),
        UsernameEquals(kubeControllerManagerUserName),
        InGroup(kubeNodeUserGroup),
        InGroup(adminUserGroup),
    )

    // Allow if: not reserved namespace OR user is exempt
    celExpr := Or(Not(Or(nsClauses...)), Or(exemptions...))
```

pkg/admissionpolicymanager/manager.go
weng271190436: You can actually do this to guarantee correctness such that the bindings are always tied to the policy?

```go
// NewPolicyWithBindings creates a PolicyWithBindings and automatically sets
// each binding's spec.policyName to match the policy's name.
func NewPolicyWithBindings(
    policy *admissionregistrationv1.ValidatingAdmissionPolicy,
    bindings []*admissionregistrationv1.ValidatingAdmissionPolicyBinding,
) PolicyWithBindings {
    for _, b := range bindings {
        b.Spec.PolicyName = policy.Name
    }
    return PolicyWithBindings{
        Policy:   policy,
        Bindings: bindings,
    }
}
```

pkg/admissionpolicymanager/podsnreplicasets.go
michaelawyu: Hi Wei! Sorry for any misunderstanding; what I meant was that, webhook.go will stay, but not all the webhooks within it. The pod/replicaset webhooks can be dropped safely eventually as VAPs can replace the logic completely; on the other hand, we have webhooks that cross-refs objects, which are not yet ready to migrate to VAP.

pkg/admissionpolicymanager/manager.go
michaelawyu: Bindings have other fields that depend on semantics (it has its own resource selectors, filters, etc.), right? Ultimately any check we do here won't be able to guarantee that it will work (especially considering that this is part of the internal logic that functions statically); it is still the UTs/ITs/E2Es that guard against any misconfigurations. My concern is that, if we are going to check them there anyway (and it's a word by word, diff-based check), is it really needed to add a very superficial check in the logic?

I am not saying that I am not a fan of defensive programming - for a dynamic system, a bit of check can help. But, for the internal, static logic part of the code, I have my reservations.

pkg/admissionpolicymanager/svcaccountsntokenreqs.go
michaelawyu: Ah, I 100% agree with the LLM responsibility part; it is always us to bear the consequences of changes, not the machine. What I was trying to say, however, is that, for this part of the logic, LLM can easily generate the end result plus its truth table; and we can easily build UTs based on the truth table to make sure that the expression is correct both in form and in semantics.

For the builder part, to do things truly correctly and elegantly, we need to start building a CEL expression tree. But the ultimate question to me, at this moment, is that, is it really worth it? We will be adding many layers of indirections and a composing logic of its own for a part of the code that is a) internal, b) largely static, c) only run once at startup, d) with limited complexity (a few clauses at max.) and e) not likely to change much over time.

I understand that this is a very subjective discussion at the moment; to unblock progress (as we really need to move to VAPs sooner), let me add a TO-DO to the code; as soon as we know that VAP setup is working (it is not connected to the main entrypoint yet) and releases seem fine, I will send another PR to switch to tree-building. Would this be alright in your opinion?

weng271190436: Sure can add TODO and come back to this. Thank you.

Also want to explain why I keep emphasizing on readability. When one writes a piece of code, it always makes sense to the person writing it because the person knows the piece of code very well. When a person reads it for the first time, the perspective is very different and appropriate abstraction can really help.

I am suggesting this because I spent significant time trying to understand this part of the code when reading it. So I suggested this because I think it helps with understanding.

```go
    exemptions = append(exemptions,
        UsernameEquals(kubeSchedulerUserName),
        UsernameEquals(kubeControllerManagerUserName),
        InGroup(kubeNodeUserGroup),
        InGroup(adminUserGroup),
    )

    // Allow if: not reserved namespace OR user is exempt
    celExpr := Or(Not(Or(nsClauses...)), Or(exemptions...))
```

And readability is not just for someone reading this code 2 months from now because they want to change the validation logic slightly or add a new VAP generator. The most immediate benefit is that it helps review. Sometimes people open a PR, read a little bit, find it hard to understand, and they will just give up.

And well-defined structure helps. To a reader, there is a big difference between custom `strings.Join` and `Or` because or is a well-understood logical operator while `strings.Join` can be used for a lot of different things. By appealing to well-understood concepts, the code is self-documenting and easy to understand.

michaelawyu: Thanks, Wei! Yeah, added the TO-DO comment (added to the package level for better visibility). Let me refactor this to a CEL expression tree based solution in a later PR after VAPs are known to be working.

pkg/admissionpolicymanager/manager.go
weng271190436: It is not about defensive programming. I actually don't like some defensive programming which is usually error handling things that can hardly happen.

I am advocating for structuring the code logically so that it helps understanding. The reason I made this comment was that when I first read this piece of code, one of the first thoughts that came to my mind was "Does the Policies and Bindings under the same VAP generator have to be related to each other in a certain way? Or are they just arbitrary collections of Policies and Bindings." And I thought about it a little bit and figured that we mostly likely do not want to have arbitrary collections of Policies and Bindings. But one Policy can have multiple Bindings. So I suggested this interface.

And the intention is that the interface itself is logical and self-documenting so that each time this piece of code is read or each time a new generator gets reviewed, people don't have to spend more time thinking about the relationship between the Policies and Bindings.

weng271190436 (approved):

michaelawyu:
Merging this PR to unblock progress -> if there's any concern, please let me know.

PR #717, opened 2026-05-18
fix: fix subsequent failure metrics and revert some changes

britaniar:

When an UpdateRun fails:
1. First reconciliation: reconcileErr contains the actual error (wrapped with ErrUserError or not), metric emitted with correct failureType
2. Subsequent reconciliations: The failed state is detected via conditions (lines 127-129 or 147-149), reconcileErr is nil, metric emitted with failureType=none

Essentially if an UpdateRun was already in a failed state and when the hub agent restarts to trigger a reconciliation, we have an early check that return nil and the reconcile error which will lead to a false metric output.

Fixes #

I have:

- Integration Test
- Locally


pkg/controllers/updaterun/metrics.go
```diff
@@ -37,49 +38,74 @@ func deleteUpdateRunMetrics(updateRun placementv1beta1.UpdateRunObj) {
...
+//   - "internal_error" for terminal failure conditions or stuck conditions that require investigation
+//
+// The failure type is derived from the condition message to ensure consistency across
+// reconciliations. This handles subsequent reconciliations where the original error
+// is not available but the failure type needs to be preserved in the condition message.
+//
+// Note: A "stuck" update run is classified as "internal_error" because it may never resolve
+// and effectively represents a failure state that needs investigation.
+func determineFailureType(cond *metav1.Condition) hubmetrics.UpdateRunFailureType {
+	if cond == nil || cond.Status != metav1.ConditionFalse {
```
jwtty, line 55: What about "Unknown" status?

pkg/controllers/updaterun/metrics.go
```diff
@@ -37,49 +38,74 @@ func deleteUpdateRunMetrics(updateRun placementv1beta1.UpdateRunObj) {
...
+	}
+
+	// Stuck is always classified as internal error
+	if cond.Reason == condition.UpdateRunStuckReason {
+		return hubmetrics.UpdateRunFailureTypeInternalError
+	}
+
+	// Check if it's a terminal failure condition
+	if isFailureReason(cond.Reason) {
+		if strings.Contains(cond.Message, controller.ErrUserError.Error()) {
```
jwtty, line 66: nit: add a comment for the exact error message string?

pkg/controllers/updaterun/metrics_test.go
```diff
@@ -17,50 +17,185 @@ limitations under the License.
...
 		},
 		{
-			name: "update run is stuck - internal error",
-			err:  errors.New("updateRun is stuck waiting for 1 cluster(s) in stage stage1 to finish updating, please check placement status for potential errors"),
+			name: "initialized condition is false with UpdateRunInitializeFailed reason and user error in message",
+			cond: &metav1.Condition{
+				Type:    string(v1beta1.StagedUpdateRunConditionInitialized),
+				Status:  metav1.ConditionFalse,
+				Reason:  condition.UpdateRunInitializeFailedReason,
+				Message: fmt.Sprintf("failed to initialize: %s: CRP not found", controller.ErrUserError.Error()),
```
jwtty: Use `controller.NewUserError` to wrap the error

pkg/controllers/updaterun/metrics_test.go
```diff
@@ -17,50 +17,185 @@ limitations under the License.
...
+			},
+			want: hubmetrics.UpdateRunFailureTypeUserError,
+		},
+		{
+			name: "initialized condition is false with UpdateRunInitializeFailed reason but no user error in message",
```
jwtty, line 124: Change to progressing? and wrap the error with `controller.NewUnexpectedBehaviorError`

pkg/controllers/updaterun/initialization.go
```diff
@@ -91,7 +91,7 @@ func (r *Reconciler) validatePlacement(ctx context.Context, updateRun placementv
 		if apierrors.IsNotFound(err) {
 			placementNotFoundErr := controller.NewUserError(fmt.Errorf("parent placement not found"))
 			klog.ErrorS(err, "Failed to get placement", "placement", placementKey, "updateRun", updateRunRef)
-			return nil, types.NamespacedName{}, fmt.Errorf("%w: %w", errValidationFailed, placementNotFoundErr)
+			return nil, types.NamespacedName{}, fmt.Errorf("%w: %s", errValidationFailed, placementNotFoundErr.Error())
```
jwtty, line 94: Why do we need these changes?

pkg/controllers/updaterun/controller.go
```diff
@@ -109,17 +109,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		return runtime.Result{}, err
 	}
 
-	// Track errors for metrics emission. The error is used to determine the failure type
-	// (user_error vs internal_error) in the emitted metrics.
-	var reconcileErr error
 	// Emit the update run status metric based on status conditions in the updateRun.
-	// Use a closure to capture reconcileErr by reference, so it reflects any updates made during reconciliation.
-	defer func() { emitUpdateRunStatusMetric(updateRun, reconcileErr) }()
+	// The failure type is derived from the condition message, not from the reconcile error.
```
jwtty: No need to add this comment. Looks weird without the context.

pkg/controllers/updaterun/initialization.go
britaniar: This was the original code. Just reverting back since we are no longer relying on the actual error and just the condition message.

pkg/controllers/updaterun/metrics.go
britaniar: Uknown status is considered part of `cond.Status != metav1.ConditionFalse`, which only happens when "stopping". The so it will be `none` failure type

pkg/controllers/updaterun/initialization.go
jwtty: I think we can keep these. Originally only 1 `%w` was allowed in `fmt.Errorf()` iirc.

jwtty (approved):

Issue #693, opened 2026-05-05
Epic: Revamp the release process


ytimocin:
Tracking epic for the phased release-process revamp. Subtasks will be filed as separate issues from this epic as work begins; this issue is the umbrella — check items off as sub-issues close.

## Phase 1 — v0.4.0: Policy, Hygiene, Correctness

- [ ] Strict SemVer + `-rc` pre-release tags (validation landed in #713, hardened in #821); cut first `release-0.Y` at first RC — waits on `v0.4.0-rc.1`
- [x] `cherry-pick/0.Y` label automation with explicit squash / conflict / force-push policy (#777)
- [ ] Define cadence (proposal: ~3-month minors, last 2 supported); closes Q1 in Phase 1, not deferred to Phase 3 — required to make `SECURITY.md` enforceable and Phase 2b's gate definable
- [ ] Support window in `SECURITY.md` (decision required: last 3 minors vs N, N−1)
- [x] `VERSIONING.md` covering agent skew, Work/AppliedWork schema, binding contract, heartbeat stability (#741)
- [x] Delete vestigial `crdbases/` from both charts (#743)
- [x] Auto-generated release notes via `.github/release.yml` (uses existing PR-title prefixes) (#713)
- [x] Add `linux/arm64` to `make push` (#745; follow-ups #851, #858)
- [x] Replace `imagetools create` retag with single dual-tag `buildx` invocation (#745)
- [x] Standalone CRD tarball as a Release asset (#743)
- [ ] Publish the real chart version to the GitHub Pages index — every release so far republished `0.1.0` (#822, in review; redundant if #823 merges first)

## Phase 2a — v0.5.0: Atomic Orchestrator + Signing + SBOMs

- [ ] Collapse `release.yml` + `chart.yml` into one orchestrating workflow with explicit `needs:` (#823, in review — end-to-end dry run green on a fork)
- [ ] Cosign keyless signing per producer job (digest-pinned, GitHub OIDC) (#824, in review, stacked on #823)
- [ ] BuildKit `--sbom=true` SPDX in image index (#824)
- [ ] `cosign attest --type=spdxjson` Sigstore-native attestation — deliberately left out of #824 pending maintainer input (see comment below)
- [ ] Signed `SHA256SUMS` for the CRD tarball (#824 — signs the `.sha256` as a blob, uploads the Sigstore bundle as a third asset)
[...]

PR #745, opened 2026-06-27
feat: build multi-arch (amd64/arm64) images and publish both tags in a single buildx

ytimocin:
## What

Make the `hub-agent`, `member-agent`, and `refresh-token` images multi-arch (`linux/amd64` + `linux/arm64`) and collapse the image publish path into a single dual-tag `docker buildx build`, removing the separate `docker buildx imagetools create` retag step.

Addresses two Phase 1 items from #693:
- Add `linux/arm64` to `make push`.
- Replace the `imagetools create` retag with a single dual-tag `buildx` invocation.

## Changes

**Makefile**
- Add `PLATFORMS` (defaults to the host platform so local single-arch builds — e.g. loading into kind — keep working) and `RELEASE_PLATFORMS` (`linux/amd64,linux/arm64`); `make push` now builds a multi-arch manifest.
- New `setup-qemu` target registers QEMU emulation for any multi-platform build **regardless of whether the buildx builder already exists**. The previous recipe registered QEMU only inside the builder-creation branch, so on a runner where the builder already existed emulation was skipped and the foreign-arch build failed with `exec format error`. Single-platform builds skip QEMU (native, not needed).
- `IMAGE_EXTRA_TAG` adds an additional tag within the same `buildx` invocation, so a stable release publishes both the full and short tags from one build.

**Dockerfiles (`docker/*.Dockerfile`)**
- Use BuildKit-provided `TARGETOS`/`TARGETARCH` for per-arch native builds. `CGO_ENABLED=1` + `GOEXPERIMENT=systemcrypto` links against the target architecture's OpenSSL, so pure cross-compilation isn't possible — each arch is built natively under emulation.
- Add `# syntax=docker/dockerfile:1` and `COPY --link`.
- Use package-path `go build` (e.g. `./cmd/hubagent/`) so a future second source file in a package isn't silently dropped.
- The pinned distroless base digest is a multi-arch image index, so BuildKit resolves the matching base layer per target arch.

**release.yml**
- A single dual-tag `make push` replaces the old `make push` + `imagetools create` retag. Stable releases publish both `vX.Y.Z` and the short `X.Y.Z`; RC tags publish only the full form (the short-tag namespace stays reserved for stable releases consumers pin to).
- The verify step now asserts each published tag is a real multi-arch manifest (via `imagetools inspect`) instead of only echoing tag strings.

## Testing

- Dry-runs confirm `make push IMAGE_EXTRA_TAG=<v>` forwards through the recursive make and emits two `--tag` flags per image; single-platform default emits one.
- End-to-end build of `refresh-token` to a local registry: one `buildx` push produced both tags at the **same digest**, each a manifest list containing `linux/amd64` + `linux/arm64`; the verify-step logic passed.
- All three `cmd` main packages compile via the package-path form.
- Confirmed the distroless base digest resolves to a multi-arch index.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,9 +61,9 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       # A single multi-arch `docker buildx build` per image publishes the full
-      # tag (e.g. ":v0.4.0") and, for stable releases, the short alias
-      # (e.g. ":0.4.0") in the same push - no separate retag step. RC images are
-      # published under the long form (":v0.4.0-rc.1") for testers only: the
+      # tag (e.g. "v0.4.0") and, for stable releases, the short alias
+      # (e.g. "0.4.0") in the same push - no separate retag step. RC images are
+      # published under the long form ("v0.4.0-rc.1") for testers only: the
       # short-tag namespace is deliberately reserved for stable releases that
       # consumers can safely pin to, so RC tags get no short alias.
       - name: Build and push images with tag ${{ env.TAG }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,39 +60,47 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      # A single multi-arch `docker buildx build` per image publishes the full
+      # tag (e.g. ":v0.4.0") and, for stable releases, the short alias
+      # (e.g. ":0.4.0") in the same push - no separate retag step. RC images are
+      # published under the long form (":v0.4.0-rc.1") for testers only: the
+      # short-tag namespace is deliberately reserved for stable releases that
+      # consumers can safely pin to, so RC tags get no short alias.
       - name: Build and push images with tag ${{ env.TAG }}
-        run: |
-          make push
-
-      # The short-tag (e.g. ":0.4.0") aliases the long form for stable releases
-      # only. RC images are published under the long form (":v0.4.0-rc.1") for
-      # testers, but we deliberately do NOT alias them to a short tag - the
-      # short-tag namespace is reserved for stable releases that consumers can
-      # safely pin to (and "imagetools create" with an RC alias would publish
-      # "0.4.0-rc.1" into that namespace, muddying it).
-      - name: Tag and push images without v prefix
-        if: ${{ !contains(needs.export-registry.outputs.tag, '-rc.') }}
         env:
           VERSION: ${{ needs.export-registry.outputs.version }}
         run: |
           set -euo pipefail
-          for IMAGE in ${{ env.HUB_AGENT_IMAGE_NAME }} ${{ env.MEMBER_AGENT_IMAGE_NAME }} ${{ env.REFRESH_TOKEN_IMAGE_NAME }}; do
-            docker buildx imagetools create \
-              --tag "${{ env.REGISTRY }}/${IMAGE}:${VERSION}" \
-              "${{ env.REGISTRY }}/${IMAGE}:${{ env.TAG }}"
-          done
+          if [[ "${TAG}" == *-rc.* ]]; then
+            make push
+          else
+            make push IMAGE_EXTRA_TAG="${VERSION}"
+          fi
 
-      - name: Verify images
+      # Confirm every published tag is a real multi-arch manifest list rather
+      # than a silently degraded single-arch image: a build that pushed only one
+      # architecture would otherwise go unnoticed until a consumer on the other
+      # architecture failed to pull. Stable releases also carry the short alias.
+      - name: Verify images are multi-arch
         env:
           VERSION: ${{ needs.export-registry.outputs.version }}
         run: |
           set -euo pipefail
-          echo "✅ Published images:"
+          tags="${TAG}"
+          if [[ "${TAG}" != *-rc.* ]]; then
+            tags="${tags} ${VERSION}"
+          fi
+          echo "✅ Verifying published images:"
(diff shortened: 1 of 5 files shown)


ytimocin:
Both Copilot comments addressed in e3572593:
- **Builder selection**: `docker-buildx-builder` now runs `buildx use` + `inspect --bootstrap` unconditionally, not only on creation — an existing-but-unselected builder can no longer route builds to a docker-driver builder without multi-platform support.
- **Image pinning**: both privileged binfmt registration containers are now pinned by tag **and digest**, and both now come from Microsoft-vetted MCR mirrors (no direct third-party pulls): `mcr.microsoft.com/mirror/docker/multiarch/qemu-user-static:7.2.0-1@sha256:cb0dff99...` and `mcr.microsoft.com/mirror/docker/tonistiigi/binfmt:qemu-v9.2.2-52@sha256:1b804311...`. The binfmt mirror digest was cross-verified byte-identical to the upstream Docker Hub tag it mirrors.

Re-verified locally after the change: full `linux/amd64,linux/arm64` build to an OCI tarball, index carries both platform manifests, and the `hubagent` ELF binaries in each variant are genuinely x86-64 / AArch64.

Makefile
```diff
@@ -295,64 +317,75 @@ crd-verify: ## Verify the chart CRD directories cover every CRD in config/crd/ba
...
-		docker buildx inspect $(BUILDX_BUILDER_NAME) --bootstrap; \
+	@case "$(PLATFORMS)" in \
+		*,*) \
+			echo "Multi-platform build ($(PLATFORMS)); registering QEMU emulation"; \
+			if [ "$(TARGET_ARCH)" = "amd64" ] ; then \
+				docker run --rm --privileged $(QEMU_IMAGE) --reset -p yes; \
+			else \
+				docker run --rm --privileged $(BINFMT_IMAGE) --install all; \
+			fi ;; \
+		*) echo "Single-platform build ($(PLATFORMS)); skipping QEMU setup" ;; \
```
michaelawyu, line 341: Hi Yetkin! Just a nit: would there still be cases where we need to build, say ARM64 image for AMD64 platforms?

.github/workflows/release.yml
```diff
@@ -60,39 +60,47 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      # A single multi-arch `docker buildx build` per image publishes the full
+      # tag (e.g. ":v0.4.0") and, for stable releases, the short alias
```
michaelawyu, line 64: Hi Yetkin! Just a nit on the comments (`:v0.4.0` -> `v0.4.0`).

michaelawyu (commented):
Just two nits. Otherwise LGTM ;)

Makefile
ytimocin: Good catch — yes: a single-platform *cross* build (e.g. `make docker-build-hub-agent PLATFORMS=linux/arm64` on an amd64 host) is a real case, and the previous guard (register only when `PLATFORMS` contains a comma) would have skipped QEMU there and failed with `exec format error`, since the builder stage runs per-target under emulation (CGO + systemcrypto). Fixed in 25ab2454: `setup-qemu` now skips registration only when `PLATFORMS` is exactly the native `$(TARGET_OS)/$(TARGET_ARCH)`, and registers emulation for everything else (multi-platform or cross single-platform).

.github/workflows/release.yml
ytimocin: Done in 25ab2454 — dropped the leading `:` from the tag examples in the comment.

Makefile
```diff
@@ -295,25 +308,34 @@ crd-verify: ## Verify the chart CRD directories cover every CRD in config/crd/ba
...
-		if [ "$(TARGET_ARCH)" = "amd64" ] ; then \
-			echo "The target is an x86_64 platform; setting up emulation for other known architectures"; \
-			docker run --rm --privileged mcr.microsoft.com/mirror/docker/multiarch/qemu-user-static:$(QEMU_VERSION) --reset -p yes; \
-		else \
-			echo "Setting up emulation for known architectures"; \
-			docker run --rm --privileged tonistiigi/binfmt --install all; \
-		fi ;\
 		docker buildx create --driver-opt image=mcr.microsoft.com/oss/v2/moby/buildkit:$(BUILDKIT_VERSION) --name $(BUILDX_BUILDER_NAME) --use; \
 		docker buildx inspect $(BUILDX_BUILDER_NAME) --bootstrap; \
 	fi
```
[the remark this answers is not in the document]
ytimocin: Addressed earlier in the digest-pin commit (now 25ab2454 after rebase): `docker-buildx-builder` unconditionally runs `docker buildx use` + `inspect --bootstrap`, so the named builder is always selected even when it already exists.

Makefile
```diff
@@ -295,25 +308,34 @@ crd-verify: ## Verify the chart CRD directories cover every CRD in config/crd/ba
...
+setup-qemu: ## Register QEMU emulation for multi-architecture image builds
 	$(info Auto-detected system architecture: $(TARGET_ARCH))
+	@case "$(PLATFORMS)" in \
+		*,*) \
+			echo "Multi-platform build ($(PLATFORMS)); registering QEMU emulation"; \
+			if [ "$(TARGET_ARCH)" = "amd64" ] ; then \
+				docker run --rm --privileged mcr.microsoft.com/mirror/docker/multiarch/qemu-user-static:$(QEMU_VERSION) --reset -p yes; \
+			else \
+				docker run --rm --privileged tonistiigi/binfmt --install all; \
+			fi ;; \
```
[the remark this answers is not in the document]
ytimocin: Addressed: both privileged images are now digest-pinned and pulled from the Microsoft-vetted MCR mirror (`mcr.microsoft.com/mirror/docker/multiarch/qemu-user-static` and `mcr.microsoft.com/mirror/docker/tonistiigi/binfmt`), digests cross-verified against Docker Hub.

michaelawyu (approved):
LGTM ;)

Issue #714, opened 2026-05-14
ci: lint and test the GitHub Actions workflow files


ytimocin:
Spun out from #693 (Phase 1 follow-up). The repo has no static analysis or tests for the files under `.github/workflows/`, so workflow regressions only surface when a release tag is actually pushed. #713 picked up a partial-match bug in `setup-release.yml`'s SemVer regex during code review that automated tooling would have flagged; we should set up that tooling.

This issue covers (a) one concrete deliverable that's ready to land and (b) a short investigation to scope follow-up testing layers.

## Part A — Static analysis (concrete, do first)

Add a small new workflow `.github/workflows/workflow-lint.yml` that runs:

- [ ] **`actionlint`** against `.github/workflows/*.yml` — validates YAML syntax, action references, `${{ ... }}` expressions, `needs:` graph, secret/permission usage. Single pinned binary, no config needed for the common case.
- [ ] **`shellcheck`** on the `run: |` bash bodies (`actionlint` shells out to it automatically if it's on `$PATH`). Catches unquoted expansions, command-substitution pitfalls, regex escaping issues.

Trigger on `pull_request` + `push` with `paths: [".github/workflows/**", ".github/release.yml"]` so it only runs when relevant files change.

**Bug class it would have caught in recent PRs:**

- The pre-#713 regex `^v[0-9]+\.[0-9]+\.[0-9]+` (no end anchor) — `shellcheck` flags ambiguous `=~` patterns and `actionlint` will surface bash quoting issues, but neither catches the missing anchor. Caveat: regex semantics are not in scope for static analysis; that's why Part B exists.
- Unquoted `${{ inputs.tag }}` expansions, unset-var risks in shared scripts, etc.

**Acceptance criteria:**

- New `workflow-lint` job runs on every PR that touches `.github/workflows/**` or `.github/release.yml`.
- Job is green on `main` at PR-merge time (any existing findings either fixed inline or explicitly ignored with a justification comment).
- Action versions and the `actionlint` binary are pinned by SHA / version, consistent with the existing pin-by-SHA policy established in #573.

## Part B — Investigation: what else should we test?

Open question: beyond static analysis, what's the right second layer?

Investigate and document findings (post a comment summarizing tradeoffs and a recommendation; don't implement until we agree):

[...]

PR #747, opened 2026-07-01
ci: add actionlint + shellcheck workflow linting

haseebnizami:

Adds a lightweight Workflow Lint workflow to catch GitHub Actions issues during review instead of mid-release.

- A new `Workflow Lint` workflow that runs `[actionlint](https://github.com/rhysd/actionlint)` whenever a PR or push touches `.github/workflows/**` or `.github/release.yml`. actionlint validates workflow syntax, expression references, and event triggers, and automatically runs `shellcheck` (already on the `ubuntu-latest` image) over every `run:` block.

- `.github/actionlint.yaml` registers our self-hosted runner label (`oracle-vm-16cpu-64gb-x86-64`) so actionlint doesn't false-fail on `ci.yml`. This is the only config the linter needs.

Refs:  #714

I have:

- Ran `actionlint` v1.7.12 with `shellcheck` 0.11.0 locally against the whole `.github/workflows/` tree — it now exits cleanly with zero findings.
- Ran `make reviewable`.

Two of the fixes were more than mechanical, so I want to call them out explicitly. please push back if you'd prefer a different approach:

1. `release.yml` (SC2193): The `[[ "${{ env.TAG }}" != *-rc.* ]]` comparison tripped shellcheck because actionlint expands `${{ env.TAG }}` before shellcheck sees it. `TAG` is already a job-level `env:` var, so I switched it to reference the shell variable directly, same runtime behavior, and it also matches GitHub's guidance on not inlining `${{ }}` into scripts:
 ```
   [[ "${TAG}" != *-rc.* ]]
```

2. trivy.yml: this one changes trigger behavior, please review. The tags: ['v*.*.*'] filter was placed under the create event, where GitHub silently ignores it, so the scanner was firing on every branch and tag creation, not just version tags. I moved the filter under push, which matches the file's own "publish semver tags as releases" comment:
```
 on:
   push:
     branches:
       - main
     tags:
       - 'v*.*.*'
   workflow_dispatch: {}
```

Before: ran on push to main + any create event.

After: runs on push to main + push of a v*.*.* tag.

If you'd rather keep the create event (e.g. guard the job with if: startsWith(github.ref, 'refs/tags/v')) or split this trigger fix into its own PR, just say so and I'll adjust.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,3 @@
+self-hosted-runner:
+  labels:
+    - oracle-vm-16cpu-64gb-x86-64
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,9 +50,9 @@ jobs:
       
       - name: Prepare necessary environment variables
         run: |
-          echo "CGO_ENABLED=1" >> $GITHUB_ENV
+          echo "CGO_ENABLED=1" >> "$GITHUB_ENV"
           KUBEBUILDER_ASSETS=$(make --silent kubebuilder-assets-path)
-          echo "KUBEBUILDER_ASSETS="$KUBEBUILDER_ASSETS"" >> $GITHUB_ENV
+          echo "KUBEBUILDER_ASSETS=$KUBEBUILDER_ASSETS" >> "$GITHUB_ENV"
 
       # Certain tests that require special setup (e.g., those that should be run with Ginkgo CLI only) will
       # be skipped in this step.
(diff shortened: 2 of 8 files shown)


weng271190436 (dismissed):

britaniar:
Hi @haseebnizami, we can merge this once you resolve the conflicts. Thanks!

haseebnizami:
> Hi Muhammad Haseeb (@haseebnizami), we can merge this once you resolve the conflicts. Thanks!

@britaniar conflicts resolved!

britaniar (approved):

britaniar (approved):

PR #764, opened 2026-07-24
fix: correct misspelling in WorkSynchronized condition message

akshita317:
### Description of your changes

Fixes a misspelling and awkward pluralization in the `WorkSynchronized` success condition message. It read:

> Works(s) are succcesfully created or updated in N target cluster(s)' namespaces

"successfully" was spelled with three c's, and "Works(s)" is grammatically off. Corrected to:

> Work(s) are successfully created or updated in N target cluster(s)' namespaces

This matches the phrasing already used elsewhere in the same file (`condition.go`: `"Work(s) have been created or updated successfully ..."`). The message appears in both `TrueClusterResourcePlacementCondition` and `TrueResourcePlacementCondition`, so both are corrected.

Fixes #762

I have:

- [x] Associated this change with a known KubeFleet Issue (Bug, Feature, etc).
- [ ] Run `make reviewable` to ensure this PR is ready for review.

### How has this code been tested

`go test ./pkg/utils/condition/...` passes and `gofmt` reports no changes. The change is limited to two user-facing condition message strings with no logic change; grepping the repository confirmed the message text is not asserted in any test, so no tests needed updating.

(Note: I ran `gofmt`, `go vet`, and the affected package's tests rather than the full `make reviewable` lint/staticcheck sweep, since this is a message-only change — happy to run the full suite if you'd like.)

### Special notes for your reviewer

Message-only change, two identical lines corrected. No behavior or API change.

Issue #762, opened 2026-07-24
[BUG] Works success message mispelling.


sjwaight:
### **Describe the bug**
In the detailed results of `describe` the message returned by `WorkSynchronized` has the word "successfully" misspelt.

```
Message: Works(s) are **succcesfully** created or updated in 2 target cluster(s)' namespaces
```

This is mispelt in `condition.go`:

- [Line 343](https://github.com/kubefleet-dev/kubefleet/blob/42a1df03b2597ad55ca771c668d0db64aac1aae0/pkg/utils/condition/condition.go#L343).
- [Line 478](https://github.com/kubefleet-dev/kubefleet/blob/42a1df03b2597ad55ca771c668d0db64aac1aae0/pkg/utils/condition/condition.go#L487).

Consider fixing pluralization of Works as well.

PR #763, opened 2026-07-24
fix: correct WorkSynchronized success message

pollychen-lab:
## Summary
- fix the WorkSynchronized success message typo from `succcesfully` to `successfully`
- normalize `Works(s)` to `Work(s)` for consistency with the existing event message wording
- add coverage for both ClusterResourcePlacement and ResourcePlacement true WorkSynchronized condition messages

Fixes #762.

## Testing
- `PATH=/tmp/go1.25.12/bin:$PATH go test ./pkg/utils/condition`
- `git diff --check`

Also tried `PATH=/tmp/go1.25.12/bin:$PATH go test ./pkg/controllers/placement`, but the suite could not start because `/usr/local/kubebuilder/bin/etcd` is not installed in this environment.

PR #766, opened 2026-07-25
test: add coverage for True(Cluster)ResourcePlacementCondition

akshita317:

Adds unit-test coverage for two previously-untested helpers in `pkg/utils/condition`:

- `ResourceCondition.TrueClusterResourcePlacementCondition`
- `ResourceCondition.TrueResourcePlacementCondition`

Both build the set of "true" placement conditions and return the one for the receiver `ResourceCondition`. Neither had any unit test. The new table-driven tests assert the full returned `metav1.Condition` (type, status, reason, message, observed generation) for every `ResourceCondition` value — `RolloutStarted`, `Overridden`, `WorkSynchronized`, `Applied`, `Available`, and `DiffReported` — using `cmp.Diff`, matching the style already used in `condition_test.go`.

I have:

`go test ./pkg/utils/condition/...` passes, and `go vet` is clean. Tests are pure and offline (no cluster or network).

(I ran `gofmt`, `go vet`, and the package tests rather than the full `make reviewable` lint/staticcheck sweep for this test-only change — happy to run the full suite if preferred.)

Test-only change; no production code is modified.

One coupling worth flagging: the `WorkSynchronized` cases assert the message text **exactly as it exists on `main` today** (`"Works(s) are succcesfully ..."`). That string is a typo being corrected in #762 / #764 — when that fix merges, these two assertions will need to be updated to `"Work(s) are successfully ..."`. I've left a comment in the test noting this. Happy to reorder/rebase if you'd prefer these tests land after the typo fix.

### DIFF
--- a/pkg/utils/condition/condition_test.go
+++ b/pkg/utils/condition/condition_test.go
@@ -386,3 +386,179 @@ func TestTrueWorkSynchronizedConditionMessage(t *testing.T) {
 		})
 	}
 }
+
+func TestTrueClusterResourcePlacementCondition(t *testing.T) {
+	const (
+		generation   = int64(3)
+		clusterCount = 2
+	)
+	tests := []struct {
+		name      string
+		condition ResourceCondition
+		want      metav1.Condition
+	}{
+		{
+			name:      "rollout started",
+			condition: RolloutStartedCondition,
+			want: metav1.Condition{
+				Status:             metav1.ConditionTrue,
+				Type:               string(fleetv1beta1.ClusterResourcePlacementRolloutStartedConditionType),
+				Reason:             RolloutStartedReason,
+				Message:            "All 2 cluster(s) start rolling out the latest resource",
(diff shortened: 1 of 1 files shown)


britaniar (changes requested):
Hi [Akshita kumari (@akshita317)](https://github.com/akshita317), we merged a PR that touched this file. Could you rebase and resolve the merge conflict? Thanks!

akshita317:
Thanks @britaniar! Rebased onto latest main and resolved the conflict in `condition_test.go`. Since #763 landed the WorkSynchronized message fix, I also updated the two `work synchronized` expectations here to the corrected "Work(s) are successfully created or updated..." message so they match the source. `TestTrueWorkSynchronizedConditionMessage` from main is preserved alongside the new coverage. All tests in the package pass locally. Ready for another look. 🙏

britaniar (approved):
LGTM. Thank you!

PR #767, opened 2026-07-26
fix: surface unknown eviction conditions in draincluster

### LINKED ISSUE #659: kubectl-fleet draincluster: safeguard unknown eviction condition
## What
`draincluster` proceeds without checking whether eviction conditions are in an `Unknown` state. Add safeguards so the command waits for (or explicitly surfaces) unknown conditions instead of treating them as success.

## Location
- `tools/fleet/cmd/draincluster/draincluster.go:190`

akshita317:

Fixes #659.

The `draincluster` command waits for each eviction to reach a terminal state and then inspects its conditions, but it only handled the `True`/`False` cases of the `Valid` and `Executed` conditions. A **missing or `Unknown`** condition fell through:

- an `Unknown` `Executed` condition was neither `nil` nor `False`, so the loop treated the eviction as **successfully executed** and reported the drain as complete;
- a missing/`Unknown` `Valid` condition was likewise not surfaced.

This masked uncertain evictions as successful drains. This change implements the pre-existing `// TODO: add safeguards to check if eviction conditions are set to unknown.`

Now a missing or `Unknown` `Valid` or `Executed` condition marks the drain unsuccessful and logs that the drain could not be confirmed, instead of silently proceeding. The existing `True`/`False` handling is unchanged.

I have:

`go build ./tools/fleet/...`, `go vet`, and the existing `go test ./tools/fleet/cmd/draincluster/...` all pass.

Note on tests: the existing suite covers the drain *helper* functions (`fetchClusterResourcePlacementNamesToEvict`, `cordon`, etc.), but not the main drain loop, because it creates an eviction and waits for it to reach a terminal state — which the fake client cannot drive without a refactor. If you'd like, I'm happy to extract the eviction-condition evaluation into a small pure helper in a follow-up so this safeguard (and the surrounding True/False handling) can be unit-tested directly.

Behavior change is limited to previously-unhandled missing/`Unknown` conditions; `True`/`False` paths are untouched.


akshita317:
Thanks for the review! Good call — I've extracted the Valid/Executed evaluation into a pure `evaluateEviction` helper and added table-driven tests covering the True/False/Unknown/missing combinations (and the invalid-but-drained reasons). No behavior change.

Note: the failing `staticcheck` and `e2e-tests (resourceplacement)` checks look unrelated to this change — `staticcheck ./...` passes cleanly for me locally on the full repo, and the resourceplacement e2e doesn't exercise the draincluster CLI. Happy to rebase or dig in if either persists.

akshita317:
Thanks — both good catches, addressed in the latest push:

- **Missing `default`/`executed` case:** the drain switch now has an explicit `evictionResultExecuted` case plus a `default` that marks the drain unconfirmed, so a future/unexpected `evictionResult` can never fall through to a "successful" drain.
- **Non-`True` statuses:** `evaluateEviction` now treats only an explicit `True`/`False` status as definitive — `Unknown`, empty, or any unexpected status maps to `indeterminate`. Added a test case for an empty/invalid `Executed` status (it caught the original gap).

On the red `unit-and-integration-tests` check: it's an unrelated flaky integration test — `pkg/controllers/bindingwatcher` `TestAPIs` timed out after 10s on an `Eventually` enqueue check (`watcher_integration_test.go:751`). The `cmd/hubagent/options` and `cmd/memberagent/options` tests both passed (`ok`), and my changed package's tests pass locally. This push re-runs CI; happy to rebase if the flake persists.

akshita317:
Following up on this one, as it has been sitting on a red check since early August.

The failing `e2e-tests (default)` job does not exercise anything this PR changes — the diff is limited to `tools/fleet/cmd/draincluster/`, a standalone CLI, while the failure is in `test/e2e/webhook_test.go:1715` (CRP placement / webhook `AfterAll` teardown). That same job also fails intermittently on `main` (e.g. run 30952331356).

The branch is now ~27 commits behind `main`, so I am happy to rebase it to get a fresh CI run — just say the word. I can also squash the four commits into one if you would prefer that for the merge.

sjwaight:
@ytimocin as raiser of original issue can you please review. Thanks.

ytimocin (approved):
Verified the fix end-to-end: a drain-loop test (fake client, conditions stamped as `markEvictionInvalid` writes them) is red on `main`, green on this branch — the old `Reason`-vs-`Message` comparison made the invalid-but-drained path unreachable. One nit: the "missing or unknown condition" log is misleading for PickFixed/duplicate-CRB evictions, where `Valid` is explicitly `False`; log `validCondition.Message` instead.

PR #768, opened 2026-07-27
test: add coverage for pod and replicaset validating webhooks

akshita317:
## Description

`pkg/webhook/pod` and `pkg/webhook/replicaset` were the only two webhook packages with no unit tests. Both validators deny creation of their resource outside the reserved `fleet-` and `kube-` namespaces, and neither branch of `Handle` was covered.

This adds table-driven tests for both, following the existing `pkg/webhook/pdb` test.

## Coverage

Each handler is now tested for all four paths through `Handle`:

| Case | Expected |
| --- | --- |
| CREATE in a non-reserved namespace (`default`) | Denied, with the resource-specific denial message |
| CREATE in `fleet-system` | Allowed |
| CREATE in `kube-system` | Allowed |
| UPDATE / DELETE (non-CREATE) | Allowed without decoding |
| CREATE with an undecodable object | `Errored(http.StatusBadRequest, ...)` |

## One deliberate difference from the pdb test

The pdb test passes `cmpopts.IgnoreFields(metav1.Status{}, "Message")` to every comparison. That also discards the denial message, which is the main thing these webhooks produce, so the denial text goes unasserted.

Here the ignore is scoped to the decode-failure case only — the decoder's wording is implementation-defined, but the code and `allowed=false` are still compared. Every other case compares the full response, including the message. Happy to align with the pdb style instead if you'd prefer consistency across the webhook tests.

## Testing

- `go test ./pkg/webhook/pod/... ./pkg/webhook/replicaset/...` — all 12 subtests pass.
- `go vet` and `staticcheck` clean on both packages.
- Sanity-checked that the tests are not vacuous: inverting `!utils.IsReservedNamespace(...)` to `utils.IsReservedNamespace(...)` in the pod handler fails both the deny and the allow cases.

### DIFF
--- a/pkg/webhook/pod/pod_validating_webhook_test.go
+++ b/pkg/webhook/pod/pod_validating_webhook_test.go
@@ -1,0 +1,182 @@
+/*
+Copyright 2026 The KubeFleet Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pod
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	admissionv1 "k8s.io/api/admission/v1"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


michaelawyu:
Hi Akshita! Thanks for contributing to the KubeFleet project. If possible, would you mind signing the commits per our contributing guide so that the DCO checks can pass?

britaniar (approved):
Hi @akshita317, the changes look good. Could you sign the commits per our contributing guide? Thanks!
https://github.com/kubefleet-dev/kubefleet/pull/768/checks?check_run_id=90059167041

akshita317:
Thanks @britaniar and @michaelawyu, and sorry for the delay here.

Both commits are now signed off and the branch is force-pushed. The `Potential fix for pull request finding` commit was the unsigned one — it was created through the Copilot Autofix suggestion UI, which does not add a `Signed-off-by` line. I rebased with `--signoff` per the DCO bot's instructions; the tree is unchanged, only the commit messages differ.

akshita317:
The `unit-and-integration-tests` failure is in `pkg/controllers/bindingwatcher`, which this PR does not touch — the only two files changed are `pkg/webhook/pod/pod_validating_webhook_test.go` and `pkg/webhook/replicaset/replicaset_validating_webhook_test.go`.

The push that triggered this run was the `--signoff` rebase to fix DCO, so the tree is unchanged from the run that passed earlier; only the commit messages differ.

This job also fails intermittently on `main` (e.g. run 30832944139). Would you mind re-running it when you get a chance?

britaniar (approved):

PR #772, opened 2026-07-29
chore: upgrade Squad from v0.9.6-build.1 to v0.11.0

serbrech:
## What

Upgrades Squad in this repo from **v0.9.6-build.1 → v0.11.0** by running `squad upgrade` with the globally installed `@bradygaster/squad-cli@0.11.0` (already the latest published version).

All Squad-owned files were regenerated by the CLI; nothing was hand-authored.

## Changes

- Re-stamped `.github/agents/squad.agent.md` to v0.11.0
- Refreshed the 11 `squad-*` GitHub workflows and `.squad/templates/`
- Scaffolded built-in agents `Rai` and `fact-checker`
- Synced 19 skills to `.github/skills/`
- Added `.mcp.json` registering the `squad_state` MCP server
- Added `.squad/.cache/` to `.gitignore`

## Regressions the upgrade introduced (fixed here)

Regenerating the workflows reverted two fixes this repo already had on `main`:

1. **`actions/checkout` downgraded v7 → v4** in 12 places across 11 workflows. A silent dependency downgrade — nothing in the upgrade output flags it.
2. **`squad-heartbeat.yml` lost the quotes around `"$GITHUB_OUTPUT"`**, which fails the `actionlint + shellcheck` job with SC2086.

Both are restored in `fix: restore local workflow fixes clobbered by squad upgrade`.

Since these files are regenerated wholesale on every `squad upgrade`, local fixes get silently reverted. So squad-owned workflows are now excluded from actionlint via `.github/actionlint.yaml`, and the root causes are reported upstream: **bradygaster/squad#1556**.

The exclusion was verified with actionlint 1.7.12 + shellcheck 0.10.0 (the same versions CI uses): the SC2086 error in `squad-heartbeat.yml` is suppressed, while the identical error injected into a non-squad workflow still fails the run — so the ignore is correctly scoped and not over-matching. Note actionlint would never have caught the `checkout` downgrade anyway (not a lint rule), so this exclusion doesn't reduce coverage there.

## Two manual adjustments to the generated output

**1. `.mcp.json` pinned to `@latest` rather than `@insider`.**
The installer pinned the MCP server to `@bradygaster/squad-cli@insider`, but that dist-tag resolves to `0.10.0-insider.1` — *older* than the stable `0.11.0` we run:

```
preview: 0.8.17-preview
latest:  0.11.0
insider: 0.10.0-insider.1
```

Confirmed `squad state-mcp` exists in 0.11.0, so the insider fallback isn't needed.

**2. Deleted `squad.agent.md.local-backup`.**
The upgrade warned about "local customizations" and wrote a backup, but `Get-FileHash` confirmed it was byte-identical to `git show HEAD:.github/agents/squad.agent.md` — checksum drift, not real edits. Nothing lost.

## Team state preserved

`team.md`, `roster.md`, `decisions.md`, casting registry and agent histories are all untouched.

`.squad/config.json` still reads `stateBackend: "two-layer"` and is byte-identical to HEAD. Verified in the CLI source that this is guaranteed rather than incidental: `runUpgrade` only *reads* `.squad/config.json`, and the sole writer of `stateBackend` is `migrateStateBackend`, gated behind `--state-backend` (`cli-entry.js:445`) — not passed here.

Passing `--state-backend two-layer` would have been wrong, not just redundant: with `current === target` its only added step is `ensureOrphanBranch`, which checks *only* `refs/heads/squad-state` and ignores `refs/remotes/origin/squad-state`. It would have fabricated a parentless README-only commit diverging from the real remote state.

## Verification

`squad doctor` — **10 passed, 0 failed**, including *"git sync hooks installed — squad sync hooks present for 'two-layer' backend"*.

`squad sync --pull` succeeds, and the local state branch is correctly connected to real history:

```
refs/heads/squad-state          = 4e810fa8
refs/remotes/origin/squad-state = 4e810fa8
```

The one `squad doctor` warning (`copilot --version` failed) is pre-existing and environmental — it only affects optional `watch` capabilities.


michaelawyu (approved):
LGTM ;)

Issue #771, opened 2026-07-29
Consider hard-failing override selectors that match no resource


serbrech:
Split out of #607 to keep that fix reviewable.

## Background

Today an override selector that matches nothing is silently ignored. This is the underlying user complaint in #607 — the envelope bug was invisible precisely because ineffective overrides produce no signal.

The #607 fix will emit a warning event/log when an override selector matches neither a selected resource nor any resource inside an envelope, but it deliberately does not change placement failure semantics.

## Proposed work

Evaluate making no-match override selectors a hard failure, likely by setting `Overridden=False` on the binding, so users cannot miss ineffective override configuration.

## Why this is separate

Non-envelope overrides that match nothing are also silently ignored today, and have been for a long time. Hard-failing changes broader override behavior well beyond envelopes, so it needs its own compatibility review rather than riding along with the envelope fix.

Current maintainer lean is toward hard-failing, but the migration impact should be assessed first — in particular, whether any existing placements rely on overrides that legitimately match nothing for a subset of clusters.

Issue #770, opened 2026-07-29
Reject envelope wrapper override selectors and warn on bypassed selectors


serbrech:
Split out of #607 to keep that fix reviewable.

## Background

Override selectors that explicitly name `ClusterResourceEnvelope` or `ResourceEnvelope` target objects that never land on member clusters. The #607 fix makes those selectors ineffective by applying overrides only to the inner manifests extracted from the envelope, but it does not give users any feedback that such a selector is meaningless.

## Proposed work

- Reject new/updated `ClusterResourceOverride` selectors targeting `ClusterResourceEnvelope`, and `ResourceOverride` selectors targeting `ResourceEnvelope`, in the validating webhook.
- Keep already-persisted or webhook-bypassed selectors harmless at runtime, surfacing a warning event/log rather than failing.

A runtime guard is needed in addition to admission validation because:

- validating webhooks only run on create/update, so overrides persisted before this change survive untouched;
- webhooks can be disabled via `--enable-webhook=false`.

## Pointers

Selector validation lives in `pkg/utils/validator/clusterresourceoverride.go` and `pkg/utils/validator/resourceoverride.go`, called from the existing override webhook handlers. There is currently no precedent in these validators for rejecting a specific kind; `pkg/utils/validator/placement.go` has comparable namespace-kind special handling.

## Notes

No override selector targeting an envelope kind was found anywhere in `examples/`, `test/`, `test/e2e/`, `hack/`, or docs, so this is expected to be a low-impact addition.

PR #774, opened 2026-07-29
fix: apply overrides to resources inside envelope objects

### LINKED ISSUE #607: [BUG] Override rules silently skipped for resources inside envelope objects
## Summary

The work generator controller (`pkg/controllers/workgenerator/controller.go`, line ~499) applies override rules to selected resources **before** determining whether they are envelope objects or regular resources:

```go
for j := range selectedRes {
    selectedResource := selectedRes[j].DeepCopy()
    // Overrides applied to the raw selected resource (the envelope CR itself)
    resourceDeleted, overrideErr := r.applyOverrides(selectedResource, cluster, croMap, roMap)
    // Then processed — envelope resources get their own Work object
    newWork, simpleManifests, err = r.processOneSelectedResource(...)
}
```

There is an existing TODO:
```go
// TODO: apply the override rules on the envelope resources by applying them on the work instead of the selected resource
```

## Problem

For **envelope objects** (dedicated envelope CRs that wrap inner resources), `applyOverrides` targets the **envelope CR wrapper**, not the resources embedded inside it.

Example scenario:
- An envelope object contains a Deployment
- A `ResourceOverride` rule says "set replicas to 5 for cluster-west"
- The override matches against the envelope CR (which has no `replicas` field) → **silently does nothing**
- The Deployment inside the envelope keeps its original replica count

**Regular (non-envelope) resources are not affected** — overrides work correctly for them.

## Impact

Users combining envelope objects with override rules (`ClusterResourceOverride` / `ResourceOverride`) will have their overrides **silently ignored** for all resources inside envelopes. No error is surfaced.

## Proposal

Apply override rules at the **Work level** rather than the selected resource level. Since envelope objects are unpacked into Work manifests, overriding at the Work level would correctly target the inner resources:

1. For envelope resources, skip `applyOverrides` on the envelope CR
2. After `processOneSelectedResource` unpacks the envelope into a dedicated Work object, apply overrides to the individual manifests within that Work
3. Ensure override rule matching works against the inner resource GVK/name/namespace, not the envelope wrapper

## Scope

- `pkg/controllers/workgenerator/controller.go` — `syncAllWork` loop and `applyOverrides` call site
- Override matching logic may need adjustment to resolve against inner manifests
- Unit/integration tests for envelope + override combinations

serbrech:
## Summary

Override rules targeting resources wrapped in `ResourceEnvelope` or `ClusterResourceEnvelope` were silently ignored.

Fixes #607

## Root cause

Two coupled defects:

- **Selection** — `FetchAllMatchingOverridesForResourceSnapshot` built override candidates only from the resource snapshot's `SelectedResources`. For an envelope placement that is the wrapper alone; the envelope's `.data` was never opened, so overrides targeting inner resources never reached the binding.
- **Application** — `syncAllWork` applied overrides before envelope detection, so rules were evaluated against the wrapper instead of the resources it carries.

Because selection never produced inner-resource candidates, fixing the work generator alone would not have been sufficient.

## Changes

- Select override snapshots by inspecting the resources inside an envelope rather than the wrapper.
- Apply overrides after envelope extraction, per inner manifest.
- Ignore overrides targeting envelope wrappers.
- A `Delete` override on an inner resource drops only that manifest; the Work is still created when this empties the envelope.
- Override failures inside an envelope are reported on the `Overridden` condition, consistent with non-envelope resources.
- Added `examples/envelopes/override-inside-envelope.yaml`.

## Behavioral change

Overrides no longer apply to envelope wrapper objects, including `/data/...` JSONPatch paths.

Rationale: an override changes resources that land on a member cluster, and the wrapper never lands. A repo-wide search found no existing usage of wrapper-targeting override selectors.

## Override candidate collection is best-effort

Candidate collection skips an inner manifest that is invalid (wrong scope for the envelope type, or a namespace that differs from the envelope's) or whose content cannot be parsed. It logs at V(2) and continues with the remaining manifests instead of returning an error.

This matters because `FetchAllMatchingOverridesForResourceSnapshot` runs in the rollout controller (`pkg/controllers/rollout/controller.go`) and in update run initialization — both upstream of the work generator. Returning an error there aborts the rollout, leaving `RolloutStarted` as `Unknown` so the work generator never runs and never reports the real failure.

`pkg/controllers/workgenerator/envelope.go` already performs these same checks and is the authoritative validator; it returns a user error that surfaces as `SyncWorkFailed`. Duplicating that validation upstream, at fatal severity, masked it. This is also consistent with the existing treatment of override selectors that match nothing, which warn rather than fail.

## Note on the `overrideContext` struct

Applying overrides after envelope extraction meant threading the target cluster and the two override snapshot maps down through `processOneSelectedResource`, `createOrUpdateEnvelopeCRWorkObj`, and `applyOverridesToEnvelopeManifests` — pushing `processOneSelectedResource` from 10 to 13 parameters.

They are bundled into a small `overrideContext` struct, assembled once in `syncAllWork`, so this cost one parameter rather than three and future override inputs will not churn these signatures.

The two alternatives were considered and rejected:

- **Fetching the snapshots deeper in the stack** — the fetches are informer-cache reads, so there is no network I/O, but `processOneSelectedResource` runs inside a nested loop over resource snapshots × selected resources. Fetching there would turn O(overrides) into O(snapshots × resources × overrides) cache reads and deep copies per reconcile, on a path that runs for every binding on every member cluster.
- **Caching the maps on the `Reconciler`** — unsafe, not merely suboptimal. `MaxConcurrentReconciles` is greater than one, so a single `Reconciler` serves concurrent reconciles for different bindings, and these maps are per-binding.

`applyOverrides` and the rest of `override.go` are unchanged.

## Validation

- `go build ./...`, `go vet`, `gofmt`, `staticcheck`, and `golangci-lint` clean.
- Unit tests pass for `pkg/utils/overrider` and `pkg/controllers/workgenerator`.
- Work generator integration suite: 92/92 specs pass (87 pre-existing + 5 new).
- The new override-selection unit tests and 5 new integration specs were each confirmed to **fail on `main` and pass with this change**, so they genuinely cover the reported bug. The same check was applied to the candidate-collection tests: reverting only `overrider.go` makes all of them fail.
- Full CI green, including all four e2e shards (`default`, `custom`, `joinleave`, `resourceplacement`).

## Note on a change that was dropped

An earlier revision of this PR also tightened `upsertWork` to compare desired manifests and apply strategy rather than relying on override snapshot hashes alone, intended as a guard against stale Works.

That change caused a real regression: it made 15 e2e specs fail (all CRO/RO/envelope/override paths) with `Applied` pinned at `ApplyInProgress` and `Available` never reported. `Manifest` embeds `runtime.RawExtension`, and `equality.Semantic` has no custom equality for it, so the comparison was byte-for-byte. Overridden manifests are re-serialized by the JSON patch library, so their bytes never match the API-server-canonicalized bytes read back from the cluster. The short-circuit could therefore never fire, `upsertWork` updated on every reconcile, and the binding's `Applied` condition was continually reset before `setBindingStatus` could propagate real Work status.

It has been reverted so this PR stays scoped to #607. The hash-only short-circuit is sound in practice: manifest content is deterministic given the resource snapshot index and the override snapshot hashes, both of which are already compared.

## Follow-ups (deliberately out of scope)

- #770 — reject envelope wrapper override selectors via webhook, plus a runtime guard.
- #771 — consider hard-failing override selectors that match nothing.

---

Note for reviewers: this repo's `.github/release.yml` expects a single `release-note/*` label, but those labels do not currently exist in the repo. `release-note/fix` matches the title prefix; the wrapper behavior change may warrant `release-note/breaking` instead.

### DIFF
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -150,13 +150,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 		}
 	}
 
-	workUpdated := false
-	overrideSucceeded := false
+	// result must be allocated before listing the works so that if listing fails (and syncAllWork
+	// is never called), the existing status outcome is preserved: the list failure is reported on
+	// the Overridden condition and no Work is reported as changed.
+	result := &syncResult{overrideFailed: true}
 	// list all the corresponding works
 	works, syncErr := r.listAllWorksAssociated(ctx, resourceBinding)
 	if syncErr == nil {
-		// generate and apply the workUpdated works if we have all the works
-		overrideSucceeded, workUpdated, syncErr = r.syncAllWork(ctx, resourceBinding, works, &cluster)
+		// generate and apply the works if we have all the works.
+		// syncAllWork always returns a non-nil result, even on error.
+		result, syncErr = r.syncAllWork(ctx, resourceBinding, works, &cluster)
 	}
 	// Reset the conditions and failed/drifted/diffed placements.
 	for i := condition.OverriddenCondition; i < condition.TotalCondition; i++ {
@@ -165,7 +168,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 	resourceBinding.GetBindingStatus().FailedPlacements = nil
 	resourceBinding.GetBindingStatus().DriftedPlacements = nil
 	resourceBinding.GetBindingStatus().DiffedPlacements = nil
-	if overrideSucceeded {
+	if !result.overrideFailed {
 		overrideReason := condition.OverriddenSucceededReason
 		overrideMessage := "Successfully applied the override rules on the resources"
 		if len(resourceBinding.GetBindingSpec().ClusterResourceOverrideSnapshots) == 0 &&
@@ -191,7 +194,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 		if err := errors.Unwrap(syncErr); err != nil && len(err.Error()) > 2 {
 			errorMessage = errorMessage[len(err.Error())+2:]
 		}
-		if !overrideSucceeded {
+		if result.overrideFailed {
 			resourceBinding.SetConditions(metav1.Condition{
 				Status:             metav1.ConditionFalse,
 				Type:               string(fleetv1beta1.ResourceBindingOverridden),
@@ -217,7 +220,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 			Message:            "All of the works are synchronized to the latest",
 		})
 		switch {
-		case !workUpdated:
+		case !result.workUpdated:
 			// The Work object itself is unchanged; refresh the cluster resource binding status
 			// based on the status information reported on the Work object(s).
 			setBindingStatus(works, resourceBinding)
@@ -411,11 +414,20 @@ func (r *Reconciler) listAllWorksAssociated(ctx context.Context, resourceBinding
 	return currentWork, nil
 }
 
-// syncAllWork generates all the work for the resourceSnapshot and apply them to the corresponding target cluster.
-// it returns
-// 1: if we apply the overrides successfully
-// 2: if we actually made any changes on the hub cluster
-func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding fleetv1beta1.BindingObj, existingWorks map[string]*fleetv1beta1.Work, cluster *clusterv1beta1.MemberCluster) (bool, bool, error) {
+// syncResult captures the outcome of a syncAllWork call.
+type syncResult struct {
+	// overrideFailed reports whether a sync error should fail the binding's Overridden condition.
+	// It is false when syncAllWork succeeds.
+	overrideFailed bool
+	// workUpdated reports whether any Work object was actually changed on the hub cluster.
+	workUpdated bool
+}
+
+// syncAllWork generates all the work for the resourceSnapshot and applies them to the corresponding target cluster.
+//
+// The returned syncResult is always non-nil, including on error paths, so the caller can safely
+// inspect its fields regardless of the returned error. See syncResult for the meaning of each field.
+func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding fleetv1beta1.BindingObj, existingWorks map[string]*fleetv1beta1.Work, cluster *clusterv1beta1.MemberCluster) (*syncResult, error) {
 	updateAny := atomic.NewBool(false)
 	resourceBindingRef := klog.KObj(resourceBinding)
 
@@ -441,17 +453,17 @@ func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding fleetv1bet
 		})
 	}
 	if updateErr := errs.Wait(); updateErr != nil {
-		return false, false, updateErr
+		return &syncResult{overrideFailed: true}, updateErr
 	}
 
 	// the hash256 function can handle empty list https://go.dev/play/p/_4HW17fooXM
 	resourceOverrideSnapshotHash, err := resource.HashOf(resourceBinding.GetBindingSpec().ResourceOverrideSnapshots)
 	if err != nil {
-		return false, false, controller.NewUnexpectedBehaviorError(err)
+		return &syncResult{overrideFailed: true}, controller.NewUnexpectedBehaviorError(err)
 	}
 	clusterResourceOverrideSnapshotHash, err := resource.HashOf(resourceBinding.GetBindingSpec().ClusterResourceOverrideSnapshots)
 	if err != nil {
-		return false, false, controller.NewUnexpectedBehaviorError(err)
+		return &syncResult{overrideFailed: true}, controller.NewUnexpectedBehaviorError(err)
 	}
 	// TODO: check all work synced first before fetching the snapshots after we put ParentResourceOverrideSnapshotHashAnnotation and ParentClusterResourceOverrideSnapshotHashAnnotation in all the work objects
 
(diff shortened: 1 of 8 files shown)


pkg/controllers/workgenerator/envelope.go
serbrech, line 42: document what the new bool in the return arguments represents.

pkg/controllers/workgenerator/envelope.go
```diff
@@ -128,7 +139,46 @@ func (r *Reconciler) createOrUpdateEnvelopeCRWorkObj(
 		work = buildNewWorkForEnvelopeCR(workNamePrefix, binding, resourceSnapshot, envelopeReader, manifests, resourceOverrideSnapshotHash, clusterResourceOverrideSnapshotHash)
 	}
 
-	return work, nil
+	return work, false, nil
+}
+
+func (r *Reconciler) applyOverridesToEnvelopeManifests(
```
serbrech, line 156: add a doc comment on the func, in particular to clarify what the return arguments are.

pkg/controllers/workgenerator/envelope.go
serbrech: Done in `fd18c41c`.

Traced what the bool actually drives before writing this: it propagates up through `processOneSelectedResource` into `syncAllWork`, which negates it into `overrideSucceeded`, and `Reconcile` uses that to fail the binding's `Overridden` condition with `OverriddenFailedReason` rather than reporting the failure as a sync error.

```go
// createOrUpdateEnvelopeCRWorkObj creates or updates a work object for a given envelope CR.
//
// The returned bool reports an override failure (used to fail the binding's Overridden condition);
// it is only meaningful when the returned error is non-nil and is always false on success.
```

Called out the "only meaningful when the error is non-nil" part explicitly because `false` on its own does not mean overrides succeeded — it is also what every non-override failure path returns.

I documented `processOneSelectedResource` the same way. You did not comment on it, but it carries this same new bool and was equally undocumented, so leaving it out would have been inconsistent within one PR. Happy to drop that hunk if you would rather keep the diff to exactly what you flagged.

pkg/controllers/workgenerator/envelope.go
serbrech: Added in `fd18c41c` — this one had no doc comment at all.

```go
// applyOverridesToEnvelopeManifests applies the override rules to each manifest extracted from an
// envelope, dropping any manifest that a Delete override removes, and returns the surviving,
// overridden manifests.
//
// The returned bool reports whether applying the override rules failed, as opposed to a manifest
// parse failure; it is only meaningful when the returned error is non-nil, and lets the caller fail
// the binding's Overridden condition. On success it is always false, so false alone does not mean
// "overrides succeeded".
```

This is where the bool originates, so it is the fuller of the three — it is the only place the true/false distinction is actually made, and the distinction is specifically override failure versus manifest parse failure. The two functions that merely propagate it (`createOrUpdateEnvelopeCRWorkObj` and `processOneSelectedResource`) get a one-sentence version instead, to avoid repeating the same paragraph three times.

Also documented the `Delete` behaviour in the summary line, since dropping a manifest mid-loop is the non-obvious part of what this function does and it is otherwise only visible from the `continue`.

Comments only, no code change. gofmt, `go build`, `go vet`, `golangci-lint`, codespell, and the package unit tests all pass locally.

pkg/utils/overrider/overrider.go
```diff
@@ -130,9 +112,13 @@ func FetchAllMatchingOverridesForResourceSnapshot(
 			}
 			if possibleCROs[croKey] {
 				filteredCRO = append(filteredCRO, &croList.Items[i])
+				matched = true
 				break
 			}
 		}
+		if !matched {
+			klog.Warningf("ClusterResourceOverrideSnapshot %s has no selector that matches a selected resource or inner envelope resource in placement %s; selectors=%+v", klog.KObj(&croList.Items[i]), placementKey, croList.Items[i].Spec.OverrideSpec.ClusterResourceSelectors)
```
michaelawyu: Hi Stephane! Just a nit about this: since this func is called at the rollout controllers when rolling out a specific placement, I guess we could assume that it is quite common for an override to not have any matching selectors  for this given placement, e.g., 

override M targets deployment A
placement X places deployment A -> matched
placement Y places deployment B -> no match

Would it be better if we just use an info log instead of warning for this, as it's a normal occurrence?

michaelawyu: But I agree 100% about the inefficiency of this M:N matches. We do have a recommended limit of the number of overrides in the docs (100 cluster-scoped, 100 per namespace), but it still stands only as an expedient solution I guess.

pkg/controllers/workgenerator/controller.go
```diff
@@ -496,32 +504,24 @@ func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding fleetv1bet
...
-				ctx, selectedResource, resourceBinding, snapshot,
+			var overrideFailed bool
+			newWork, simpleManifests, overrideFailed, err = r.processOneSelectedResource(
+				ctx, selectedResource, overrideCtx, resourceBinding, snapshot,
 				workNamePrefix, resourceOverrideSnapshotHash, clusterResourceOverrideSnapshotHash,
 				activeWork, newWork, simpleManifests)
 			if err != nil {
 				klog.ErrorS(err, "Failed to process the selected resource", "snapshot", klog.KObj(snapshot), "selectedResourceIdx", j)
-				return true, false, err
+				return !overrideFailed, false, err
```
michaelawyu: Hi Stephane! There might be a minor misrepresentation issue here: say the `processOneSelectedResource` call failed at the a non-override related step, e.g. `UnmarshalJSON` cacll -> the method would return overrideFailed = true, which becomes overrideSucceeded in the caller (syncAllWork), and the work generator will populate a OverridenSuccessful condition, even though overriding hasn't been performed yet, and the failure would register as a syncErr.

This is a bit different from the existing behavior (due to the fact that overriding happens first on non-enveloped resources).

michaelawyu: (But to be honest, semantics wise overrding is a part of the synchronization process). I guess this is a more of design flaw in the status reporting design...

michaelawyu (commented):
Hi Stephane! I added a few comments, PTAL 🙏 Nothing particularly blocking I guess.

pkg/controllers/workgenerator/controller.go
serbrech: the func signature is:
// syncAllWork generates all the work for the resourceSnapshot and apply them to the corresponding target cluster.
// it returns
**// 1: if we apply the overrides successfully**
// 2: if we actually made any changes on the hub cluster

so, if overrideFailed = true, we return !overrideFailed (false) -> aka  "we did not apply the overrides successfully"

but this is just horrible to read, I'll encapsulare the return booleans into a type so that it's clear what we're returning.

serbrech: Implemented the readability refactor in `4d5eb4b2`.  `syncAllWork` now returns `(*syncResult, error)`:  ```go type syncResult struct {     // overrideFailed reports whether a sync error should fail the binding's Overridden condition.     // It is false when syncAllWork succeeds.     overrideFailed bool     // workUpdated reports whether any Work object was actually changed on the hub cluster.     workUpdated bool } ```  I deliberately used the **failure** direction rather than preserving the old `overrideSucceeded` name. The line under discussion is now:  ```go return &syncResult{overrideFailed: overrideFailed}, err ```  so there is no longer a double negative (`!overrideFailed` becoming an unnamed "success" bool). `Reconcile` keeps the result object intact and reads `result.overrideFailed` / `result.workUpdated` directly rather than unpacking it back into separate booleans.  This is status-attribution plumbing only; the existing behavior is unchanged, including the semantic oddity you called out for non-override failures before override application. I mapped all 12 old return triples to the new result and had an independent reviewer verify equivalence. Local build, vet, lint, codespell, unit tests, and the workgenerator envtest suite (92/92 specs) pass.

pkg/utils/overrider/overrider.go
serbrech: Good point, thanks Michael. A placement not selecting the resource an override targets is a normal case during rollout, so I downgraded both the CRO and RO non-match logs from warning to V(2) info to avoid log noise. Done in bed1f249.

michaelawyu (approved):
LGTM ;)

The work generator definitely needs some refactoring in the near future... Will create an issue and use the baseline to inform future dev work.

PR #781, opened 2026-08-03
interface: [FEP-0001] add API definition for the Placement Policy and Cluster Request API objects

### LINKED ISSUE #779: [FEP-0001] Add API definition: Placement Policy API objects
Add API definition for the Placement Policy API objects based on [FEP-0001](https://github.com/kubefleet-dev/enhancements/tree/main/enhancements/0001-placement-policy).

### LINKED ISSUE #780: [FEP-0001] Add API definition: Cluster Request API object
Add API definition for the Cluster Request API object based on [FEP-0001](https://github.com/kubefleet-dev/enhancements/tree/main/enhancements/0001-placement-policy).

michaelawyu:

This pull request introduces the new `PlacementPolicy`, `ClusterPlacementPolicy`, and `ClusterRequest` API objects per [FEP-0001](https://github.com/kubefleet-dev/enhancements/tree/main/enhancements/0001-placement-policy).

Fixes #779
Fixes #780 

I have:

NA

NA


apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
```diff
@@ -0,0 +1,93 @@
...
+	// +kubebuilder:validation:Optional
+	ProvisionedClusterName *string `json:"provisionedClusterName,omitempty"`
+
+	// The latest observed creation timestamp across all the member clusters. This field is used
+	// as a expedient solution to verify if a cluster request is still valid for consideration, i.e.,
+	// if the current latest observed cluster creation timestamp is later than this timestamp in the
+	// the status, a new member cluster must have been created after the cluster request was created,
+	// and thus the cluster request should be considered stale and can be ignored.
+	// +kubebuilder:validation:Optional
+	LatestObservedClusterCreationTimestamp *metav1.Time `json:"latestObservedClusterCreationTimestamp,omitempty"`
```
ytimocin: Thinking if this could be named `LastObservedClusterCreationTimestamp`...

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
```diff
@@ -0,0 +1,93 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	ClusterRequestCondTypeCompleted = "Completed"
+)
+
+// ClusterRequest is a KubeFleet API that represents a request for a member cluster to be provisioned.
+// It is created by KubeFleet when it fails to find a member cluster that can fulfill some scheduling
+// requirements as specified in a PlacementPolicy or ClusterPlacementPolicy object,
```
ytimocin: I am thinking of the following case:
* User creates a PlacementPolicy **w/o noticing that it points to a cluster that does not exist**
* Using that PlacementPolicy, the user sees that there is a new cluster in the Fleet

Should we add a warning in the PlacementPolicy step or is this case fine?

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
```diff
@@ -0,0 +1,93 @@
...
+type ClusterRequestSpec struct {
+	// The cluster selector terms that describe the requirements for a new member cluster.
+	//
+	// If not specified, any member cluster can satisfy the request.
+	//
+	// This field is immutable after creation.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="the clusterSelectorTerms field is immutable"
+	ClusterSelectorTerms []LabelAndClusterPropertySelectorTerm `json:"clusterSelectorTerms,omitempty"`
```
ytimocin: Do the labels in **Label**AndClusterPropertySelectorTerm belong to a Cluster? Are they Cluster labels? If so, should we rename `LabelAndClusterPropertySelectorTerm` to `ClusterLabelAndPropertySelectorTerm`?

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -0,0 +1,537 @@
...
+const (
+	PlacementPolicyCondTypeResourceCollected = "ResourceCollected"
+	PlacementPolicyCondTypeScheduled         = "Scheduled"
+	PlacementPolicyCondTypeSynchronized      = "Synchronized"
+	PlacementPolicyCondTypeAvailable         = "Available"
+)
+
+// The reasons for each condition type of PlacementPolicy and ClusterPlacementPolicy API objects.
+const (
+	PlacementPolicyResourceCollectedCondReasonResourceCollected            = "AllResourceCollected"
```
ytimocin:
```suggestion
	PlacementPolicyResourceCollectedCondReasonResourcesCollected            = "AllResourcesCollected"
```

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -0,0 +1,537 @@
...
+	// A list of cluster selectors that specifies the target clusters where KubeFleet should place
+	// the resources. A cluster selector consists of a list of label and cluster property selectors
+	// and a count; and for each cluster selector, KubeFleet will pick `count` number of clusters
+	// that match the given selectors for placing the resources.
+	//
+	// If not specified, KubeFleet will place the resources to all available member clusters.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=5
```
ytimocin: Would 5 be enough for users who could have thousands of clusters?

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -0,0 +1,537 @@
...
+	ClusterSelectors []ClusterSelector `json:"clusterSelectors,omitempty"`
+
+	// A list of resource selectors that specifies the resources that KubeFleet should place across
+	// the target clusters.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=20
+	ResourceSelectors []ObjectReference `json:"resourceSelectors,omitempty"`
```
ytimocin: I believe we could have some more consistency in the names. Like `ClusterSelectors` is an array of **ClusterSelector** but `ResourceSelectors` is an array of **ObjectReference**. Should **ObjectReference** be `ResourceSelector`?

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -0,0 +1,537 @@
...
+// +kubebuilder:validation:XValidation:rule="!has(self.minCount) || !has(self.count) || (type(self.count) == string && self.count == 'All') || (type(self.count) == int && self.minCount <= self.count) || (type(self.count) == string && self.count.matches('^[1-9][0-9]{0,2}$') && self.minCount <= int(self.count))",message="minCount must be less than or equal to count when count is not All"
+type ClusterSelector struct {
+	// A list of terms that form the selector. The terms are ORed, i.e., a cluster would match the selector
+	// if it matches any of the terms.
+	//
+	// If not specified, the selector will match all clusters.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxItems=5
+	Terms []LabelAndClusterPropertySelectorTerm `json:"terms,omitempty"`
```
ytimocin: This was **ClusterSelectorTerms** above but just **Terms** here. Is that to not have a clash of names?

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -0,0 +1,537 @@
...
+	// +kubebuilder:validation:Maximum=999
+	MinCount *int32 `json:"minCount,omitempty"`
+
+	// Whether to submit a cluster request if the cluster selector cannot be fulfilled.
+	//
+	// The default value is true. Note that this field takes effect if and only if cluster requests are enabled in KubeFleet.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
+	RequestClusterIfUnfulfilled *bool `json:"requestClusterIfUnfulfilled,omitempty"`
```
ytimocin: Should we just use `requestCluster`?

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -0,0 +1,537 @@
...
+	// A list of label key-value pairs that a cluster must have to match this selector term.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxProperties=5
+	MatchLabels map[string]string `json:"matchLabels,omitempty"`
+
+	// A list of label expressions that a cluster must all satisfy to match this selector term.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxItems=5
```
ytimocin: Could some users need more?

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -0,0 +1,537 @@
...
+	// Available options are:
+	// * ShareOwnership: KubeFleet registers itself as a co-owner of the resource.
+	// * ReportError: KubeFleet reports an error when a resource to be placed is already owned by other controllers.
+	//
+	// The default value is ReportError.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=ShareOwnership;ReportError
+	// +kubebuilder:default=ReportError
+	WhenOwnedByOthers WhenOwnedByOthersOption `json:"whenOwnedByOthers,omitempty"`
```
ytimocin, line 356: Could these be just `OwnedByOthers` or `IfOwnedByOthers`?

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
britaniar: I wonder if using just `RequestCluster` would be confusing to user. They could be expecting that this is requesting a cluster to be created despite clusters existing with the requirements.

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
```diff
@@ -0,0 +1,93 @@
...
+	// The specification of the cluster request.
+	// +kubebuilder:validation:Optional
+	Spec ClusterRequestSpec `json:"spec,omitempty"`
+
+	// The observed status of the cluster request.
+	// +kubebuilder:validation:Optional
+	Status ClusterRequestStatus `json:"status,omitempty"`
+}
+
+type ClusterRequestSpec struct {
```
britaniar, line 49: Missing `PlacementRef`:  https://github.com/kubefleet-dev/enhancements/blob/5f296e492f751a58e34aaf8587f3e6076bcad854/enhancements/0001-placement-policy/README.md?plain=1#L1082-L1083

and its definition.

```suggestion
type ClusterRequestSpec struct {

  // A reference to the placement policy that created this cluster request.
  PlacementPolicyRef *ObjectReference `json:"placementPolicyRef,omitempty"`
```

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -0,0 +1,537 @@
...
+
+// PlacementPolicy is the KubeFleet API that enables users to place resources within a namespace across
+// member clusters.
+//
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Namespaced,categories={kubefleet, kubefleet-placement}
+// +kubebuilder:storageversion
+// +kubebuilder:validation:XValidation:rule="!has(self.spec.resourceSelectors) || self.spec.resourceSelectors.all(s, size(s.namespace) == 0 || s.namespace == self.metadata.namespace)",message="PlacementPolicy is namespace-scoped and it may only select resources within the same namespace as the PlacementPolicy itself; leave the namespace field empty or set it to the PlacementPolicy namespace in the resource selectors"
```
britaniar: nit can be addressed later: add tests for all CEL validation

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
britaniar: I think it would be better to keep all the same. So, if we change this then the following should also be changed:
- WhenDrifted
- WhenAlreadyExists
- WhenPlacementDeleted
- WhenNamespaceDoesNotExist

Our previous placement (CRP/RP) uses when.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -0,0 +1,537 @@
...
+)
+
+type WhenNamespaceDoesNotExistOption string
+
+const (
+	WhenNamespaceDoesNotExistOptionCreateNamespace WhenNamespaceDoesNotExistOption = "CreateNamespace"
+	WhenNamespaceDoesNotExistOptionReportError     WhenNamespaceDoesNotExistOption = "ReportError"
+)
+
+type Toleration struct {
```
britaniar, line 489: nit: could add CEL to ensure the fields that need to be empty or set are.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
ytimocin: That is also correct.

britaniar: Another option to keep consistent with the other fields across our repo is `RequestClusterWhenUnfulfilled` (keep bool) or `WhenUnfulfilled` (actionable enum), which matches the `When*` naming convention used in `SyncStrategy` and across our repo.

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
michaelawyu: Hi Yetkin! Do you think `LastObservedMostRecentClusterCreationTimestamp` would be better?

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
michaelawyu: Hi Yetkin! Cluster requests are available only in supported environments and in the cluster selector we have included an option to skip cluster requests even if the cluster selector cannot be fulfilled (the `requestClusters` field). Would you prefer if we set this default in the cluster selector to false, i.e., by default no cluster requests will be sent even if cluster selectors cannot be fulfilled?

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
michaelawyu: Hi Yetkin! Yeah, they are cluster labels; and the name proposed is better. Thanks 🙏 Let me make the edit.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Hi Yetkin! Thanks 🙏 Will make the edit.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Hi Yetkin! Yeah, how about 20? We could bump this further if we get further requests from future users.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Hi Yetkin! Ah, my bad. Resource selectors actually need a struct of their own (similar to how RPs/CRPs currently handle them). Let me make the edit.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Hi Yetkin! The field is named `terms` here as the owner field is already called `clusterSelectors`. For the `clusterSelectorTerms` field in `ClusterRequests`, it was named so as other fields in the cluster selector (`count`, `minCount`, etc.) do not apply to cluster requests.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Hi Yetkin! Yeah, let me make the edit.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Hi Yetkin! In most cases we have observed so far, the users seem to tend to have simpler label matchers/expressions. How about we make it 10 at this moment? We could add more if such use cases arise.

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
michaelawyu: Thanks Britania 🙏 I must have missed that; my bad. Let me add the field back

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Thanks Britania! Yeah, we definitely need to do this; let me add a work item on Kanban. ([Link](https://github.com/kubefleet-dev/kubefleet/issues/793))

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Hi Britania! Yeah, thanks 🙏 Let me add that.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Hi Yetkin! I used the `When...` format just for consistency reasons; happy to discuss about it further if needed 🙏

Makefile
```diff
@@ -280,9 +280,11 @@ crd-package: ## Package the raw CRDs into a release tarball with a SHA-256 check
...
 
 .PHONY: crd-verify
-crd-verify: ## Verify the chart CRD directories cover every CRD in config/crd/bases
+## Verify the chart CRD directories cover every CRD in config/crd/bases
+## Note (chenyu1): kubefleet.dev CRDs are ignored for now until the implementation is completed.
+crd-verify:
 	@bases="$$(mktemp)"; charts="$$(mktemp)"; \
-	ls config/crd/bases/ | sort > "$$bases"; \
+	ls config/crd/bases/ | grep -v '^placement\.kubefleet\.dev' | sort > "$$bases"; \
 	{ ls charts/hub-agent/templates/crds/; ls charts/member-agent/templates/crds/; } | sort > "$$charts"; \
```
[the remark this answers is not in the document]
michaelawyu: N/A -> we will only add CRDs in the new group to the chart when the necessary prerequisite works are done.

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
```diff
@@ -0,0 +1,101 @@
...
+// ClusterRequest is a KubeFleet API that represents a request for a member cluster to be provisioned.
+// It is created by KubeFleet when it fails to find a member cluster that can fulfill some scheduling
+// requirements as specified in a PlacementPolicy or ClusterPlacementPolicy object.
+//
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories={kubefleet, kubefleet-placement}
+// +kubebuilder:storageversion
+type ClusterRequest struct {
```
serbrech, line 36: Just a consideration that came up from other side-discussions.
Openshift Hive system has a similar concept. They called the CR ClusterClaim: https://deepwiki.com/openshift/hive/7.3-clusterpool-and-clusterclaim-crds

Worth having a look.

britaniar:
Hi @michaelawyu looks like some of the integration tests are failing due to not being able to create the CRD. Please take a look. Seems like some of the CEL might not be able to execute. Can add validation in a follow up PR elsewhere.
image

michaelawyu:
> Hi michaelawyu (@michaelawyu) looks like some of the integration tests are failing due to not being able to create the CRD. Please take a look. Seems like some of the CEL might not be able to execute. Can add validation in a follow up PR elsewhere. 

Thanks, Britania! Let me fix that -> it appears that some of the checks have to be moved at the runtime VAP stage, will add a task on the board.

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
michaelawyu: Thanks, Stephane! Yeah, the current proposal assumes that we have a concept like cluster pools or cluster templates where clusters can be provisioned (activated) from, but we kept that part (semi-intentionally) vague in the API to move things forward -> will chat with Yetkin, Yoo Bin, and stakeholders to see what is the best way to move forward on this; will tweak the design (and add fields) as needed.

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
serbrech: this will create a ClusterRequest. whether the ClusterRequest is fulfilled is up to the obvserver of that CR, and to the gates in place for the ClusterRequest to be satisfied.
I think it's fine to create the ClusterRequest by default and move the decision on the provider.
The provider could be configured as "reject all" for example, if you never want to automatically create clusters?

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
serbrech: I like the Britania's suggestion to avoid the boolean

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
serbrech: I think it's usually best practice to not combine too many selectors.
you're often much better off defining a label that matches the grouping instead.

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// The name of the cluster that has been provisioned for this cluster request, if any.
+	// +kubebuilder:validation:Optional
+	ProvisionedClusterName *string `json:"provisionedClusterName,omitempty"`
+
+	// The last observed most recent creation timestamp across all the member clusters. This field is used
+	// as an expedient solution to verify if a cluster request is still valid for consideration, i.e.,
+	// if the currently observed most recent cluster creation timestamp is later than this timestamp in the
+	// status, a new member cluster must have been created after the cluster request was created,
+	// and thus the cluster request should be considered stale and can be ignored.
```
serbrech: > a new member cluster must have been created after the cluster request was created,
 and thus the cluster request should be considered stale and can be ignored

That's a bit unclear to me.
why is a cluster request stale as soon as another member joins?
if you have 2 cluster requests for 2 different apps, with different requirement, their fulfillment is independent and there is no staleness implied.

michaelawyu: Hi Stephane! When a new member cluster joins, it will trigger KubeFleet to reconcile all unfulfilled placement policies; and those with submitted cluster requests will either a) withdraw their cluster requests if the new cluster is picked, or b) refresh their cluster requests by updating this timestamp so that the requests will be considered again. 

Will edit the comment to be a bit more clear -> Edited.

michaelawyu: I added this field as it feels a bit like etcd's revision number (or a primary/secondary DB's transaction ID) read so far, as creation timestamps are also a monotonically increasing sequence inside one hub cluster; the placement policy, when submitting a cluster request, will constantly bump this number as new member clusters join to prove that it has caught up with all the changes so far on the memberships. And the platform/cloud provider code only needs to process cluster requests that have caught up, as an effort to contain herding effects/over-provisioning.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Let me change that to an enum as suggested. Thanks all 🙏 ->

two options are added: `WhenUnfulfilled` -> `RequestCluster` or `KeepSearching`.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
michaelawyu: Edit: I tweaked the limit to 10 instead to satisfy the CEL expression check complexity limit. We could bump this to a higher limit if needed (and move the CEL expr to a VAP).

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
ytimocin: Is this for a specific cluster? Then `LastObservedClusterCreationTimestamp` may be better. Why did we want to add `MostRecent` in `LastObservedMostRecentClusterCreationTimestamp`? Just asking to understand your reasoning better.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
ytimocin: Sounds good

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
ytimocin: sounds good

ytimocin: I was also talking about all of them but I am good with keeping how they are

apis/kubefleet.dev/placement/v1alpha1/clusterrequest_types.go
ytimocin: Cluster pools and templates sound like something we could use in future. Are they more like warm cluster pools?

Makefile
```diff
@@ -280,9 +280,9 @@ crd-package: ## Package the raw CRDs into a release tarball with a SHA-256 check
 	@echo "Packaged CRDs into $(CRD_PACKAGE_DIR)/$(CRD_PACKAGE_NAME).tgz"
 
 .PHONY: crd-verify
-crd-verify: ## Verify the chart CRD directories cover every CRD in config/crd/bases
+crd-verify: ## Verify the chart CRD directories cover every CRD in config/crd/bases; note (chenyu1): kubefleet.dev CRDs are ignored for now until the implementation is completed.
```
ytimocin, line 283: 👍
good catch

ytimocin (approved):
LGTM

michaelawyu:
Thanks, all 🙏 Will merge the PR now to unblock progress; if there is any concern or feedback, please let me know. I could submit additional follow-up PRs.

PR #796, opened 2026-08-05
fix: requeue transient policy snapshot lookup errors

### LINKED ISSUE #681: [BUG] Scheduler drops placement from queue on transient apiserver errors during policy snapshot lookup
### Describe the bug

In `pkg/scheduler/scheduler.go:189-197` (on `main`), the call site for `s.lookupLatestPolicySnapshot` calls `s.queue.Forget(placementKey); return` for any error. The helper currently returns three error categories:

- "no latest policy snapshot" — transient, plain `fmt.Errorf`
- "multiple active policy snapshots" — wrapped `ErrUnexpectedBehavior`
- `List` failure — `ErrAPIServerError` (and `ErrUnexpectedBehavior` for non-`StatusError` cache faults)

All three take the same `Forget`-without-requeue path. The "no snapshot" case is recoverable via the policy-snapshot watcher (it fires on Create / `IsLatestSnapshot` flip). But for `ErrAPIServerError` (transient apiserver/cache blip during `List`) the placement is dropped from the queue with no requeue scheduled, and only re-enters via unrelated event sources (placement update, member-cluster churn, binding deletion). In a quiet cluster the failure window is effectively unbounded — the user sees `SchedulingPolicyApplied=False` with no progress until something else touches the CRP/RP.

The two sibling error paths in the same file (`scheduler.go:146` after `FetchPlacementFromKey`, `scheduler.go:167` after `cleanUpAllBindingsFor`) branch on `errors.Is(err, controller.ErrUnexpectedBehavior)` for "no requeue" and call `s.queue.AddRateLimited(placementKey)` for everything else. Line 196 is the local outlier.

### Expected behavior

Transient apiserver errors should be requeued with backoff, matching the pattern at lines 146 and 167. Filing this to confirm whether the current `Forget`-on-all-errors is intentional.

### Additional context

- Pre-existing on `main`; not introduced by #667.
- Related: #643 / #667 (refactor extracting `LookupLatestPolicySnapshot`). The new sentinels (`ErrNoLatestPolicySnapshot`, `ErrMultipleActivePolicySnapshots`) introduced in #667 would make the fix a clean `errors.Is` branch.

jackwalkerlabs:

The scheduler treated every failure to look up the latest policy snapshot as watcher-recoverable. That is correct when no latest snapshot exists (and for the multiple-active-snapshot invariant), but a transient API-server or cache `List` failure has no guaranteed follow-up event, so `Forget` could drop a quiet placement indefinitely.

This change branches on the two dedicated non-requeue sentinels. Those cases still call `Forget`, while all other lookup failures call `AddRateLimited` and retry with backoff.

Fixes #681

I have:

- Red on unchanged `main`: `go test ./pkg/scheduler -run TestScheduleOncePolicySnapshotLookupErrors -count=1` failed because API-server and unexpected cache errors were forgotten instead of rate-limited.
- Green after the fix: the same four-case regression passes for no snapshot, API-server failure, unexpected cache failure, and multiple active snapshots.
- `KUBEBUILDER_ASSETS="$(make --silent kubebuilder-assets-path)" go test ./pkg/scheduler/... -count=1`
- `CGO_ENABLED=1 KUBEBUILDER_ASSETS="$(make --silent kubebuilder-assets-path)" go test ./pkg/scheduler -race -count=1`
- `make reviewable`
- `git diff --check`

There are no API or schema changes. The new scheduler-level test uses an intercepted fake client and a tracking queue so it can verify the exact `Forget` versus `AddRateLimited` decision without waiting on real backoff timers.

AI assistance disclosure: OpenAI Codex was used to help inspect the repository, implement the change, and run validation. I reviewed the complete diff and verified the red/green behavior locally.

### DIFF
--- a/pkg/scheduler/schedule_once_test.go
+++ b/pkg/scheduler/schedule_once_test.go
@@ -1,0 +1,146 @@
+/*
+Copyright 2026 The KubeFleet Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheduler
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+
+	fleetv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
+	"github.com/kubefleet-dev/kubefleet/pkg/scheduler/queue"
+)
+
+type trackingPlacementSchedulingQueue struct {
+	nextKey     queue.PlacementKey
+	rateLimited []queue.PlacementKey
+	forgotten   []queue.PlacementKey
+}
+
+func (q *trackingPlacementSchedulingQueue) Run() {}
+
+func (q *trackingPlacementSchedulingQueue) Close() {}
+
+func (q *trackingPlacementSchedulingQueue) CloseWithDrain() {}
+
+func (q *trackingPlacementSchedulingQueue) NextPlacementKey() (queue.PlacementKey, bool) {
+	return q.nextKey, false
+}
+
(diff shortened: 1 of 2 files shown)


michaelawyu:
Hi Jack! Thanks so much for the fix 🙏 I think it looks great.

pkg/scheduler/schedule_once_test.go
```diff
@@ -0,0 +1,146 @@
+/*
```
michaelawyu, line 1: Just a minor issue with the UTs: if possible, could this be merged into the current `scheduler_test.go` file?

pkg/scheduler/schedule_once_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+
+	fleetv1beta1 "github.com/kubefleet-dev/kubefleet/apis/placement/v1beta1"
+	"github.com/kubefleet-dev/kubefleet/pkg/scheduler/queue"
+)
+
+type trackingPlacementSchedulingQueue struct {
```
michaelawyu, line 37: Hi Jack! It might be easier to stub the rate limiter and use the simple queue instead. But that's not a blocker in any way.

michaelawyu (commented):
Some minor comments. Thanks again for submitting the fix.

Issue #791, opened 2026-08-03
[FEP-0001] Verify the cluster claim workflow


michaelawyu:
Verify the cluster claim workflow per specs in [FEP-0001](https://github.com/kubefleet-dev/enhancements/tree/main/enhancements/0001-placement-policy).

ytimocin:
## Interim verification report: cluster claim workflow (FEP-0001)

The claim controllers (#786, #788) haven't landed yet, so full verification isn't possible — as a first pass I verified the workflow *as specified*, end-to-end, via a spike (#811) with a fake provisioner and a withdrawer prototype built to the FEP's stated semantics. The happy path, withdrawal when another cluster satisfies the selector, the failure path, and staleness refresh all close. The spike also surfaced places where the spec is ambiguous or violates its own contract; each is pinned as a runnable spec in #811 so the discussion has concrete anchors:

1. **The withdrawal predicate fires before the cluster is usable.** The FEP withdraws a claim once a cluster "is joined to the fleet" — but a provisioner-created `MemberCluster` matches its selector at object creation, minutes before the scheduler's eligibility gate passes (`ClusterEligibilityChecker.IsEligible`: agent online, recent heartbeat, `Joined=True`, healthy) and before taints are considered (the claim spec carries no tolerations). Demonstrated in #811: with raw-match withdrawal the claim is deleted while `IsEligible` still returns false, so the placement stays unschedulable with no outstanding claim — and no concurrency-budget entry — to explain why. An eligibility-gated variant in the same spike holds the claim through the join window. Withdrawal, and the check gating new claims, should reuse the scheduler's predicate.

2. **No deletion protocol.** Withdrawal is a hard delete "regardless of the status of the cluster request," but real provisioning is a 5–15 minute cloud operation. Demonstrated in #811: a provisioner finalizer turns withdrawal into an indefinite Terminating state — and the contract doesn't say whether that claim still occupies the one-per-placement slot (starvation) or is replaced (double-provisioning), nor who owns a cluster whose claim was withdrawn mid-provision. CAPI `Machine` and Karpenter `NodeClaim` both treat `deletionTimestamp` as cancel-and-clean-up, never as instant disappearance.

[...]

Issue #788, opened 2026-08-03
[FEP-0001] Implement the scheduling framework for placement policies


michaelawyu:
Implement the scheduling framework for placement policies per specs in [FEP-0001](https://github.com/kubefleet-dev/enhancements/tree/main/enhancements/0001-placement-policy).

PR #820, opened 2026-08-13
feat: [FEP-0001] implement the placement policy controller

### LINKED ISSUE #786: [FEP-0001] Implement the Placement Policy controllers
Implement placement policy controllers per specs in [FEP-0001](https://github.com/kubefleet-dev/enhancements/tree/main/enhancements/0001-placement-policy).

ytimocin:
## Description

Closes #786.

Implements the placement policy controller for FEP-0001: it reconciles `PlacementPolicy` and `ClusterPlacementPolicy`, resolves their cluster selectors against the member cluster inventory, reports scheduling status, and manages the `ClusterClaim` lifecycle for selectors it cannot fulfill.

> [!NOTE]
> **Status: lands unwired.** Per review on #829, the experimental-API PRs merge as dormant code so existing users are unaffected until the feature is complete. This PR adds the controller, its unit and envtest coverage, and the API changes only. The hub agent flag, scheme registration, RBAC, the chart CRDs, and the kind e2e come in a later integration PR; the e2e is parked at `parked/fep0001-placement-policy-e2e`.

**What works**

- **Selector evaluation** — `matchLabels`, `matchLabelExpressions` (upstream label selector semantics, including absence handling), and `matchClusterPropertyExpressions` with both string and numeric (quantity) operators against reported properties and resource usage; `count`/`minCount` resolution including the `All` sentinel. Terms are ORed error-tolerantly: a term that cannot be evaluated does not veto a later term that matches on its own, and the error surfaces only when nothing matched.
- **Fulfillment counts only schedulable clusters** — a cluster satisfies a selector only once it passes the scheduler's own eligibility gate (`clustereligibilitychecker`: member agent online, heartbeating, joined) and its taints are tolerated. A provisioner-created cluster that has registered but not joined does not count, and does not withdraw a claim. This is the join-window issue raised on #791.
- **Status** — the `Scheduled` condition uses the API's existing reason constants (`FoundAllRequiredClusters` / `FailedToFindSomeRequiredClusters`), plus `desiredClusters`, `scheduledClusters`, and `activeClusterClaims`.
- **Cluster claims** — added for unfulfilled selectors that opt into `AddClusterClaim`, withdrawn once the selector is satisfied, with the freshness marker stamped at creation and advanced as the fleet grows.

**Design decisions worth reviewing**

- *Claim naming*: deterministic and namespace-qualified (selector index + hash of the policy's namespaced name). Claims are cluster-scoped while policy names are only unique per namespace, so a naive scheme collides; determinism also makes creation get-or-create, so a restarted reconciler converges instead of duplicating claims.
- *Claim cleanup*: a namespaced policy cannot own a cluster-scoped claim — garbage collection rejects cross-scope owners, though admission accepts them, which makes `ownerReferences` an inviting trap here. Ownership therefore uses labels (now exported API constants, since provisioners and `kubectl` users depend on them) and cleanup uses a policy finalizer, added before the first claim exists and released only after an **uncached** read confirms no claims remain — a read that goes by each claim's immutable `spec.placementPolicyRef` rather than the labels, since a label-stripped claim vanishing from a label-selected list would be orphaned permanently.
- *Terminating claims*: a claim held by a provisioner finalizer keeps its budget slot and its name, so a slow teardown can starve the slot but can never cause double-provisioning. A claim withdrawn within a reconcile round also keeps its slot until it is observed gone, so a differently-named claim for another selector cannot stand beside it; and a claim round that fails partway does not publish its partial count.
- *Watch volume*: member cluster events pass through a projection predicate that ignores heartbeat and observation timestamps; time-driven eligibility changes are covered by periodic requeues instead.

`pkg/v1/controllers/placementpolicy/IMPLEMENTATION-NOTES.md` records these decisions, the deliberate divergences from the FEP text (all in the direction of the #791 findings), and the API gaps found while implementing — including two pinned as `KNOWN GAP` specs:

1. `count: 0` and negative integers bypass CRD validation (the `XIntOrString` pattern only constrains the string form), so the controller rejects them at evaluation time. #829 now adds exactly that CEL rule — whichever PR merges second flips the `KNOWN GAP` spec pinning this.
2. Numeric operators (`Gt`, `Lt`, …) are accepted in `matchLabelExpressions` at admission; the API defers misuse to the scheduling phase.

**Not in scope** (deferred, and noted in the implementation notes): the FEP's fleet-wide concurrency limit and the eligible-keys allowlist, both prose-only config surfaces today; per-selector cluster *ranking*, which belongs with #788.

**Found by manual testing on a live cluster**

Two ways a policy could be silently starved of its claim, both from one root cause — a name legal for a Kubernetes object is not always legal where the controller put it:

- A policy name longer than 63 bytes overflowed the ownership *label value*, so claim creation failed API validation and the reconciler hot-looped while the policy reported no claims. Long names now contribute a prefix plus a hash to the label; the authoritative identity is read from `spec.placementPolicyRef` (required, immutable, unbounded), which the watch mappers now use.
- Truncating a long name for the generated *claim name* could land on a separator and produce a name the API server rejects. Both truncation points now trim trailing separators.

The name hash is 64 bits of SHA-256 rather than a 32-bit FNV sum, since a collision would let one policy select — and therefore withdraw — another's claim. Both cases have regression tests at the truncation boundaries.

Also added from review: metrics for scheduling status and outstanding claim count (dropped when the policy is deleted), and documentation that stopping the controller while claims are outstanding leaves those policies undeletable until it runs again.

**Split out**

- The `clusteraffinity` dash-parsing fix that rode here is now #900.

## How has this code been tested

- `make reviewable` (fmt, vet, lint, staticcheck, crd-verify, go mod tidy).
- Unit tests: selector matching, count resolution, structural validation, aggregation, condition mapping, toleration matching, claim naming and selection.
- Integration (envtest): scheduling status across the eligibility join window, taints and tolerations, `minCount`, `count: All`, the cluster-scoped API, and the full claim lifecycle — creation, budget capping, `KeepSearching`, freshness refresh, policy-deletion cleanup, selector-change replacement, and both provisioner-finalizer paths.
- E2E: a kind e2e for both the namespaced and cluster-scoped APIs is parked at `parked/fep0001-placement-policy-e2e` and rides with the integration PR, since it needs the flag and the chart.
- Manual testing against a kind hub, driving the controller directly: the join window (a matching cluster that has not joined does not satisfy a selector or withdraw a claim), claim add/withdraw, provisioner-finalizer teardown, both known API validation gaps, malformed cluster property data isolated to the offending cluster, out-of-band claim deletion, finalizer stripping, heartbeat staleness caught by the periodic requeue, and claim naming across namespaces and scopes. This is where the two naming bugs above came from.

## Special notes for your reviewer

This grew out of #791 (verifying the claim workflow) and the spike in #811. Where the FEP text and the verification findings disagree, the implementation follows the findings and documents why:

| FEP as written | Implemented | Why |
| --- | --- | --- |
| Withdraw when a matching cluster joins | Withdraw when a *schedulable* cluster matches | Otherwise the claim disappears minutes before the cluster is usable, and the placement is left unscheduled with nothing explaining why |
| Withdraw regardless of claim status, no finalizer contract | Provisioner finalizers honored; Terminating claims keep their slot | Hard-deleting mid-provision has no defined owner for the resulting cluster |
| No naming scheme | Deterministic, namespace-qualified | Cluster-scoped claims from same-named policies in different namespaces would collide |
| No cleanup contract | Policy finalizer + ownership labels | Cross-scope owner references cannot work; policy deletion would orphan claims forever |

Happy to split any of this out if it is easier to review in pieces — the commits are self-contained.


ytimocin:
**Live: policy self-heals when a satisfying cluster is removed**

On a live cluster running the FEP-0001 stack, with a `count: 2` `ClusterPlacementPolicy` satisfied by two eligible member clusters (`scheduled=2/2`, no claim):

- Deleting one satisfying `MemberCluster` → the controller re-evaluates on the delete event → `scheduled=1/2`, and an **`AddClusterClaim`** selector re-issues a claim for the shortfall; a **`KeepSearching`** selector drops to `1/2` with no claim. The policy `spec` is untouched (only status moves).
- The delete goes through a cleanup finalizer, so the cluster is `Terminating` briefly and is correctly excluded from the eligible count the moment `deletionTimestamp` is set.

Note for the claim-fulfillment contract (#831): the re-issued claim is named per `(policy, selector-index)` and carries **no reference to the removed cluster**, so a fulfiller can't distinguish "first shortfall" from "re-issued after a delete" — the re-issue-vs-remediate question. (Tested on the integrated #820 + #829 build.)

pkg/controllers/placementpolicy/selector.go
```diff
@@ -0,0 +1,400 @@
...
+	for i := range terms {
+		matched, err := matchesTerm(cluster, &terms[i])
+		if err != nil {
+			return false, err
+		}
+		if matched {
+			return true, nil
+		}
+	}
+	return false, nil
```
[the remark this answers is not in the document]
ytimocin: Addressed: `matchesAnyTerm` now continues past a term that errors, records the first error, and returns it only if no later term matched — a cluster that fails a numeric term but satisfies a later label-only term matches.

pkg/controllers/placementpolicy/claims.go
```diff
@@ -0,0 +1,338 @@
...
+		}
+		klog.V(2).InfoS("Withdrawing a cluster claim", "clusterClaim", claim.Name, "placementPolicy", klog.KObj(policy.Unwrap()))
+		if err := r.Delete(ctx, claim); err != nil && !errors.IsNotFound(err) {
+			return outstanding, err
+		}
+	}
+
+	if len(wantedByName) == 0 {
+		return outstanding, nil
+	}
```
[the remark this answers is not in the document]
ytimocin: Addressed: a claim with a non-zero `deletionTimestamp` is counted as outstanding and keeps its budget slot (and its deterministic name blocks re-creation) until a later watch-driven reconcile observes it gone.

pkg/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,233 @@
...
+		// claims are deliberately left untouched: their specs reflect the last valid
+		// selectors.
+		klog.ErrorS(err, "Failed to evaluate the cluster selectors", "placementPolicy", req.NamespacedName)
+		return runtime.Result{}, r.updateStatus(ctx, policy, nil, nil, invalidSelectorsCondition(policy.GetGeneration(), err))
+	}
+
+	activeClaims, claimErr := r.reconcileClaims(ctx, policy, outcomes, mostRecentClusterCreation)
+	// The scheduling status is written even when claim reconciliation failed partway; the
+	// claim error is returned afterwards so the round is retried.
+	if err := r.updateStatus(ctx, policy, outcomes, &activeClaims, scheduledCondition(policy.GetGeneration(), outcomes)); err != nil {
```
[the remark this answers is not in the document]
ytimocin: Addressed: the claim count is published only from a completed round; when `reconcileClaims` returns an error the reported count is `nil`, so the last completed value stands in status and metrics until the retry corrects it.

pkg/controllers/placementpolicy/claims.go
```diff
@@ -0,0 +1,338 @@
...
+// read from the API server directly, not the informer cache: a claim created moments before
+// the policy deletion might not have reached the cache yet, and releasing the finalizer on a
+// stale zero would orphan it permanently (nothing else ever looks at claims of a gone policy).
+func (r *Reconciler) cleanupClaims(ctx context.Context, policy policyObject) error {
+	if !controllerutil.ContainsFinalizer(policy.Unwrap(), claimCleanupFinalizer) {
+		return nil
+	}
+
+	claims := &kfplacementv1alpha1.ClusterClaimList{}
+	if err := r.uncachedReader.List(ctx, claims, claimOwnershipLabels(policy)); err != nil {
```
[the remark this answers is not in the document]
ytimocin: Addressed: ownership is keyed on the immutable `spec.placementPolicyRef`; cleanup and the finalizer release list claims by that reference rather than by the mutable labels, and the labels are self-healed only as a convenience for external consumers.

apis/kubefleet.dev/placement/v1alpha1/clusterclaim_types.go
```diff
@@ -22,6 +22,22 @@ import (
...
+	// placement — associates a claim with its policy. The namespace label is empty for claims
+	// added by a ClusterPlacementPolicy.
+	//
+	// Object names may be longer than the 63-byte limit on label values: when a policy's name
+	// does not fit, the label carries a shortened, collision-resistant form of it instead, and
+	// selecting on the label with the policy's own name then matches nothing. The authoritative
+	// identity is always spec.placementPolicyRef, so a consumer that must handle policies with
+	// names longer than 63 bytes should list cluster claims and match on that field rather than
+	// select on this label.
+	ClusterClaimPolicyNameLabel      = "placement.kubefleet.dev/policy-name"
```
michaelawyu: Hi Yetkin! Just a naming nit: would `ClusterClaimPlacementPolicyNameLabel` (`placement.kubefleet.dev/placement-policy-name`) and `ClusterClaimPlacementPolicyNamespaceLabel` (`placement.kubefleet.dev/placement-policy-namespace`) be better?

pkg/v1/controllers/placementpolicy/eligibility.go
```diff
@@ -0,0 +1,30 @@
+/*
```
michaelawyu, line 1: HI Yetkin! Just a nit about the filename 😂: since it only features a namespace, shall we call it `eligibiility.go` instead?

pkg/v1/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// fetchPolicy retrieves the policy object for the given request; requests without a namespace
+// concern the cluster-scoped ClusterPlacementPolicy API. Errors, including not-found ones, are
+// returned as is for the caller to inspect.
+func (r *Reconciler) fetchPolicy(ctx context.Context, req runtime.Request) (kfplacementv1alpha1.PlacementPolicyAccessor, error) {
+	if req.Namespace == "" {
+		policy := &kfplacementv1alpha1.ClusterPlacementPolicy{}
+		if err := r.Get(ctx, req.NamespacedName, policy); err != nil {
+			return nil, err
```
michaelawyu: Hi Yetkin! For the error handling part, would you prefer using the error utility we added earlier? Or course if you perfer adding the categorization logic later, that would be great as well.

pkg/v1/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,240 @@
...
+		latency := time.Since(startTime).Milliseconds()
+		klog.V(2).InfoS("Placement policy reconciliation ends", "placementPolicy", req.NamespacedName, "latency", latency)
+	}()
+
+	policy, err := r.fetchPolicy(ctx, req)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.V(4).InfoS("Ignoring not-found placement policy", "placementPolicy", req.NamespacedName)
+			// The policy is gone for good; stop reporting metrics for it.
+			forgetPolicyMetrics(req.Namespace, req.Name)
```
michaelawyu, line 94: Hi Yetkin! Just a nit: for this one, putting it in the Deleted event handler can save a GET trip I think. Of course the current implementation has nothing wrong and this is not a blocker of any sort.

pkg/v1/controllers/placementpolicy/claims.go
```diff
@@ -0,0 +1,506 @@
...
+		return nil
+	}
+
+	// The release decision cannot rest on the ownership labels: labels are mutable, and a claim
+	// whose labels were stripped would vanish from a label-selected list, releasing the
+	// finalizer over a claim that still exists -- permanently orphaned, since nothing else ever
+	// looks at claims of a gone policy. The claim's spec.placementPolicyRef is immutable, so
+	// every claim is listed and ownership is decided by the reference.
+	claims := &kfplacementv1alpha1.ClusterClaimList{}
+	if err := r.uncachedReader.List(ctx, claims); err != nil {
```
michaelawyu, line 376: Hi Yetkin! For this one, since the claims have owner labels added, it might be better to have a label matcher in the LIST call to filter the results on the API server end.

pkg/v1/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,240 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+
+	clusterv1beta1 "github.com/kubefleet-dev/kubefleet/apis/cluster/v1beta1"
+	kfplacementv1alpha1 "github.com/kubefleet-dev/kubefleet/apis/kubefleet.dev/placement/v1alpha1"
+	"github.com/kubefleet-dev/kubefleet/pkg/scheduler/clustereligibilitychecker"
+)
+
+const (
```
michaelawyu, line 43: Hi Yetkin! For performance reasons, we moved away from period requeues for the CRP/RP APIs; I wonder if we should do the same for PlacementPolicy APIs as well. With the work queue being exposed it is possible to have watchers (on `MemberClusters`) to auto-queue PlacementPolicies upon cluster-side changes.

michaelawyu: Of course it might be better to address this in separate PR, which is totally OK to me as well.

pkg/v1/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,240 @@
...
+		return runtime.Result{}, err
+	}
+	var mostRecentClusterCreation metav1.Time
+	for i := range memberClusters.Items {
+		if ts := memberClusters.Items[i].CreationTimestamp; ts.After(mostRecentClusterCreation.Time) {
+			mostRecentClusterCreation = ts
+		}
+	}
+
+	outcomes, err := evaluateSelectors(policy.GetSpec(), memberClusters.Items, r.eligibility)
```
michaelawyu, line 118: Hi Yetkin! I realized that since the scheduling framework part might not have been integrated with the controller here, the logic here might not be ready yet. Would you prefer if I wait until that part of the integration is completed?

michaelawyu: I am only asking cause at this moment the control loop is just inspecting how many clusters one selector can select and if the count matches with the desired number, without tracking any scheduling decisions.

michaelawyu: But if there is part that I miss/misunderstand, please let me know.

michaelawyu (commented):
Hi Yetkin! I added some comments, PTAL. If this is not the right opportunity to review this PR (i.e., some of the integration might not be ready), I am really sorry.

PR #823, opened 2026-08-13
ci: collapse release and chart workflows into one atomic orchestrator

ytimocin:
`release.yml` and `chart.yml` ran independently with no ordering between them. A chart publish could fail, or never run, while the GitHub Release was already public — advertising charts nobody could pull — and images could be published after the charts referencing them.

One workflow now owns the release. The GitHub Release is created as a draft before any artifact exists, every producer uploads into that draft, and `publish-release` flips it visible only once every producer that was supposed to run has succeeded. Chart jobs wait on the images they reference. RC gating moves from a workflow-level tag filter to a single `prerelease` output, so it is derived once instead of respelled per job.

```text
setup                       validate the tag; derive registry, version, prerelease
 └── create-draft-release    create (or reuse) the GitHub Release as a draft
      ├── publish-images     multi-arch push, then verify both platforms
      │    ├── publish-charts-oci     stable only
      │    └── publish-charts-pages   stable only
      └── publish-crds       package the CRDs, upload to the draft
publish-release              needs ALL of the above; verifies assets, then publishes
```

The `publish-release` condition is the only subtle part:

| Case | Result |
| --- | --- |
| Stable, all producers green | publish |
| RC, chart jobs skipped by design | publish |
| Stable, chart jobs skipped for any reason | do not publish |
| Run cancelled at any point | do not publish |

It overrides the implicit `success()` on `needs` because the chart jobs are legitimately skipped for RCs, and uses `!cancelled()` rather than `always()` so Cancel after the producers finish does not still publish.

**Reviewing this:** `release.yml` is effectively a new file — read it whole rather than as a diff. The two jobs with real branching logic live in `hack/release/` so they get shellcheck coverage and are testable; `hack/release/test-release-scripts.sh` exercises them against a stubbed `gh` (no test framework, no new dependencies) and `workflow-lint.yml` runs it. The assertion descriptions read as sentences — skim those and skip the plumbing.

**Known trade-off.** `publish-charts-pages` still uses `stefanprodan/helm-gh-pages`, a third-party Docker action holding `contents: write` whose base image is a floating tag resolved inside its own Dockerfile, so the SHA pin does not pin what executes. This PR increases its criticality: a chart failure previously left the release published anyway, now it blocks the release. I think that is the right direction — failing closed with the release still a draft beats publishing a release with missing charts — and `harden-runner` in audit mode records its egress meanwhile. Happy to replace it here instead of deferring if maintainers prefer.

**Executed end to end on a fork** (`ytimocin/kubefleet`, this exact commit tagged `v0.98.0`): [run 33680771871](https://github.com/ytimocin/kubefleet/actions/runs/33680771871), 9m25s, all seven jobs green, release flipped from draft to published last. Checked independently afterwards, not by the workflow's own steps:

- `ghcr.io/ytimocin/kubefleet/{hub-agent,member-agent,refresh-token}:{v0.98.0,0.98.0}` are OCI image indexes with `linux/amd64` + `linux/arm64`; both tags resolve to the same digest.
- The CRD bundle and its `.sha256` are attached to the release and the checksum verifies.
- OCI charts `hub-agent`/`member-agent` `0.98.0` pull with `appVersion: v0.98.0`; the gh-pages index gained a `0.98.0` entry for each chart (this is the bug #822 fixes, so #822 becomes redundant once this merges).

The RC path was exercised on the stacked #824 (`v0.99.0-rc.1`): chart jobs skipped, `publish-release` still ran and published it flagged as a pre-release, no `0.99.0-rc.1` short alias and no RC chart in either channel. Before running, validation was actionlint, shellcheck, the script suite, and a dry run of the tag/asset logic against real GitHub draft releases.

Part of #693.


serbrech:
lgtm

weng271190436 (approved):

PR #824, opened 2026-08-13
ci: sign release artifacts with cosign and attach SPDX SBOMs

ytimocin:
> **Stacked on #823.** Based on `ci/release-orchestrator` so the diff shows only this change; retarget to `main` once #823 merges.

Images and the CRD bundle now carry Sigstore signatures, keyless via the release workflow’s GitHub OIDC identity — no long-lived signing key to hold or rotate. Each image also carries a per-platform SPDX SBOM attached to its index, readable without pulling the image.

Three decisions worth reviewing:

- **Signed by digest, not by tag.** buildx records the digest it pushed via `--metadata-file` and the signing script reads it from there. Signing a tag would sign whatever that tag resolves to when cosign runs, which is not necessarily what the run built. `--recursive` also covers the per-platform manifests inside the index, so a verifier that already resolved to an arch-specific digest still finds a signature.
- **The work list comes from the build-metadata directory,** not a hand-maintained list of image names that could drift from what is actually built and publish an unsigned image while the job still exits 0. The complementary invariant — that everything a release should contain was in fact built — stays with the verification step, which runs before anything is signed.
- **The verification identity is bound to `release.yml`’s `job_workflow_ref`,** not to the repository, so another workflow that later gains `id-token: write` cannot mint signatures that pass. No ref constraint, so both tag pushes and `workflow_dispatch` verify. Any OIDC trust policy added later must be scoped the same way, never `repo:...:*`.

Both signing steps verify what they just produced and fail the release if it does not check out — a signature nobody can verify is worse than none, because `RELEASING.md` tells users to rely on it. The CRD bundle’s checksum is signed as a blob and its bundle uploaded as a third release asset, which `publish-release` now requires. Consumers that enumerate release assets should know about that addition.

SBOM and signing are one PR because signing by digest needs the `--metadata-file` plumbing the SBOM change adds; splitting them would land plumbing with no consumer.

**Executed end to end on a fork** (`ytimocin/kubefleet`, this exact commit): stable `v0.99.0` in [run 33680770499](https://github.com/ytimocin/kubefleet/actions/runs/33680770499) and RC `v0.99.0-rc.1` in [run 33680770294](https://github.com/ytimocin/kubefleet/actions/runs/33680770294), both green. Verified afterwards from a laptop with a stock `cosign` v3.1.3, not by the workflow's own verify steps:

- All three `v0.99.0` images verify with `--certificate-identity-regexp` set to the pattern `RELEASING.md` documents (with the fork's repo substituted), and also with the exact `--certificate-identity https://github.com/ytimocin/kubefleet/.github/workflows/release.yml@refs/tags/v0.99.0` — so the Fulcio SAN is the `job_workflow_ref` form the identity pattern assumes, which was the open question above.
- Each image index carries an SPDX SBOM for both `linux/amd64` and `linux/arm64` (77–97 packages each).
- `kubefleet-crds-v0.99.0.tgz.sha256.bundle` is attached and `cosign verify-blob` passes against the `.sha256`, whose checksum matches the tarball.
- RC run: `publish-release` published the release as a pre-release with the three CRD assets; no `0.99.0-rc.1` short image alias, no RC chart in OCI or the index.

Keyless signing needs real Actions OIDC, so the script suite covers the logic *around* cosign — which digest is signed, what happens when a step fails — against a stub; the fork run is what covers cosign itself.

Scoped out deliberately: `cosign attest --type=spdxjson` and `release-metadata.json`, both still open in #693. The attestation question was raised for maintainer input in [this comment](https://github.com/kubefleet-dev/kubefleet/issues/693#issuecomment-4382117838) and has not been answered; this PR proceeds without it rather than settling it.

Part of #693.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,6 +66,13 @@ jobs:
         with:
           ref: ${{ needs.setup.outputs.tag }}
 
+      # This job gates every publishing job, so failing here is the last point
+      # at which a release can be abandoned for free. Signing runs long after
+      # the images are public, so a run started from a ref that cannot produce
+      # a verifiable signature has to be stopped now or not at all.
+      - name: Check this ref can sign a release
+        run: ./hack/release/check-signing-ref.sh
+
       - name: Create or reuse the draft release
         run: ./hack/release/create-draft-release.sh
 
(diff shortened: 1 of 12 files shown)


sjwaight (approved):
LGTM

PR #827, opened 2026-08-14
feat: [FEP-0001] add accessor interface for placement policy and related APIs

michaelawyu:

This PR adds accessor interfaces for the placement policy and related APIs, which enables controllers to process namespace-scoped and cluster-scoped variants of the same API types easier.

I have:

- [NA] Associated this change with a known KubeFleet Issue (Bug, Feature, etc).

N/A

N/A

### DIFF
--- a/apis/kubefleet.dev/placement/v1alpha1/interface.go
+++ b/apis/kubefleet.dev/placement/v1alpha1/interface.go
@@ -75,9 +75,6 @@ func (p *ClusterPlacementPolicy) SetStatus(status PlacementPolicyStatus) {
 	p.Status = status
 }
 
-var _ PlacementResourceSnapshotAccessor = &PlacementResourceSnapshot{}
-var _ PlacementResourceSnapshotAccessor = &ClusterPlacementResourceSnapshot{}
-
 // PlacementResourceSnapshotAccessor provides unified access to the spec of placement resource snapshot resources,
 // namespace-scoped and cluster-scoped.
 //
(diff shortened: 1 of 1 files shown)


apis/kubefleet.dev/placement/v1alpha1/interface.go
```diff
@@ -0,0 +1,150 @@
...
+// namespace-scoped and cluster-scoped.
+//
+// +kubebuilder:object:generate=false
+type PlacementPolicyAccessor interface {
+	client.Object
+
+	GetSpec() *PlacementPolicySpec
+	GetStatus() *PlacementPolicyStatus
+
+	SetSpec(PlacementPolicySpec)
```
ytimocin, line 42: Naming check before more of us build on this: v1beta1 uses `PlacementObj` / `BindingObj` with `GetPlacementSpec()` and the `*GetterSetter` split. Is `*Accessor` + bare `GetSpec()` the convention for v1 going forward? Fine either way, I just want to switch #829 over once.

ytimocin (approved):

apis/kubefleet.dev/placement/v1alpha1/interface.go
michaelawyu: Hi Yetkin! I've gotten no preference on this either; it's just copilot suggesting that having the API name within might feel a bit redundant 😂 Would it be OK if we keep it as it is for now just to unblock progress? This is an area that we can still (relatively easily) change later I think

PR #829, opened 2026-08-18
feat: [FEP-0001] support annotation-based placement

### LINKED ISSUE #785: [FEP-0001] Implement control loops in resource watcher for supporting annotation-based placement
Implement support for annotation-based placement per specs in [FEP-0001](https://github.com/kubefleet-dev/enhancements/tree/main/enhancements/0001-placement-policy).

ytimocin:
## Description of your changes

Implements FEP-0001 annotation-based placement (#785): setting

```yaml
metadata:
  annotations:
    kubefleet.dev/cluster-selectors: "env=staging,count=All;env=canary,region=eastus,count=2"
```

on any resource the hub agent watches keeps a `PlacementPolicy` (namespaced source) or `ClusterPlacementPolicy` (cluster-scoped source) in sync with the annotation for as long as it is present. The annotation is named after the `spec.clusterSelectors` field it compiles into, per the discussion on #785; `region` expands to `topology.kubernetes.io/region` and `alias` to `kubefleet.dev/cluster-alias`.

> **Status: lands unwired (deferred integration, per review).** The controller merges as dormant code — no feature flag, no chart value, no hub-agent setup, no resource-watcher hooks — so nothing changes for an existing installation until the rest of FEP-0001 makes the experience usable. The integration change that connects it (and the kind e2e suite that drives it, parked on `parked/fep0001-annotation-placement-e2e`) follows once the chain works end to end.

### How it works

- **Parser** (`pkg/v1/controllers/annotationplacement/parser.go`): pure translation of the annotation grammar into `[]ClusterSelector`, mirroring the API's own limits so a bad annotation is reported against the annotation rather than surfacing later as a failed policy write.
- **Policy derivation** (`policy.go`): deterministic name `<kind>-<name>-<16-hex hash of apiGroup/Kind/namespace/name>`, provenance labels (`placement.kubefleet.dev/parent-api-group|parent-kind|parent-name`), and a **non-controller owner reference** to the annotated resource (setting `controller`/`blockOwnerDeletion` would need `update` on the owner's `finalizers` subresource for arbitrary kinds). All API-defaulted fields are set explicitly so reconciliation converges instead of issuing no-op updates forever.
- **Reconciler** (`controller.go`): a controller-runtime controller, generic over a request that carries the source's kind alongside its namespaced name, since it reconciles resources of any kind the hub agent watches. It creates/updates/deletes the generated policy through `controllerutil.CreateOrUpdate` and a resource-version-guarded delete; records `PlacementPolicyCreated/Updated/Deleted` and `InvalidClusterSelectorsAnnotation` events on the annotated resource. An invalid annotation is not retried (no retry makes it parse) and the policy from the last valid annotation is left standing. The merge overwrites only what this controller generates — foreign labels and owner references are preserved. Annotated resources are read straight from the API server (a cached read of an unseen kind would start an informer for it, one that never syncs for a kind without list/watch); generated policies are read through the manager's cache, which the policy watch is served from too.
- **Drift repair**: `SetupWithManager` watches `PlacementPolicy`/`ClusterPlacementPolicy` through the manager's cache and maps every policy event back to the resource it was generated from (owner reference whose identity reproduces the policy's own name), so an edited or deleted policy is restored even for a source that never changes again. The watch admits only what this controller repairs (generation, labels, owner references), so status writes by the scheduling controllers do not fan out into uncached reads of the annotated resources. Annotated resources themselves arrive through a `SourceQueue`, which the resource watcher will feed in the integration change: `Add` writes straight into the controller's own workqueue, so the watcher is never blocked and a burst of reports for one resource costs one reconcile.
- **Cleanup**: annotation removed → policy deleted; source deleted → policy deleted explicitly (GC would wait for *every* owner reference, and the merge deliberately preserves foreign ones); source becomes ineligible for placement (e.g. a ReplicaSet adopted by a Deployment, reported by the watcher as a filtered delete) → policy deleted via the same eligibility test the watcher uses.
- **Not wired**: nothing constructs the reconciler in the hub agent. The flag, chart value, RBAC, `CheckCRDInstalled` gate, and the resource-watcher hooks that fed annotated resources into the controller's queue were part of earlier revisions and are deferred to the integration change. What remains in shared code is the `placement.kubefleet.dev` group joining the resource config's default skip list, so KubeFleet's own bookkeeping objects are never treated as placeable resources.
- **Alias producer**: moved to #901 per review (aliases are not specific to annotation-based placement), which has since merged; this branch reads `ClusterAliasLabel` from `placement/v1beta1` on main.
- **`pkg/utils/naming`**: shared truncate/hash/label-value helpers extracted so #820 can drop its private copies on its next rebase (verified behaviour-identical).

### Adjacent fixes that surfaced while testing

- **Tombstone deletions were silently dropped** in the resource watcher's filter: pre-existing, found here, now its own PR (#899).
- **`count` accepted out-of-range bare integers**: `Pattern` constrains only the string form of an int-or-string, so `count: 1000` was accepted while `count: "1000"` was rejected. A CEL rule now bounds the integer form to the same 1–999; proven against a real API server in envtest.
- **`minCount<=count` produced an opaque evaluation error beside a plain pattern violation** for over-long digit strings; the rule now leaves malformed strings to the field's own validation, and `count` gains `MaxLength=3` (its longest legal values are exactly 3 characters).
- **`maxUnavailable`/`maxSurge` accepted negative bare integers** on served v1/v1beta1 APIs: now its own PR (#898), so it can ship and backport independently.

### Deliberately out of scope

- The `placement.kubefleet.dev` CRDs are not added to the hub-agent chart (`crd-verify` excludes them until the implementation completes, per #781). The integration change will bring back the fail-fast `CheckCRDInstalled` gate for them.
- Wiring the ResourcePlacement webhook in (behavior change on a served API deserving its own review).
- `whenUnfulfilled` on generated selectors keeps the API's `AddClusterClaim` default deliberately: requesting a cluster that does not exist yet is the point of these APIs, and the annotation is meant to be the lowest-friction way to ask. Guarding against unsatisfiable selectors belongs to the claim fulfillment layer.

Fixes #785

## I have:

- [x] Read and followed KubeFleet's [contribution process].

## How has this code been tested

- Unit tests throughout (`annotationplacement` 98%+, `naming` 100%), including mutation-tested assertions (e.g. removing the old-object annotation check fails the removal test).
- envtest suites against a real API server: the full annotation round trip (set → change → invalidate → remove, both scopes), drift repair, and count bounds in both int-or-string forms — plus the controller running under a real manager, fed through the source queue, proving the watch on the generated policies (edit and delete of a generated policy are repaired with no event on the source).
- Manually on a Kind cluster (on the earlier, wired revision of this branch) running the hub agent image with the flag on: fail-fast without CRDs, recovery with them, the full annotation lifecycle, live drift repair through the informers (edit and delete of the generated policy), skipped-namespace eligibility, and source-deletion cleanup. That run also caught a double-reported deletion (fixed in its own commit) that no unit test had surfaced.
- Manually on real AKS (same wired revision; 3 clusters, 3 hub-agent replicas, webhook enabled, `denyModifyMemberClusterLabels=true`): the same lifecycle matrix, plus alias seeding on MemberCluster join under the live guard rail, an admin's alias rename surviving reconciles, the guard rail still denying non-reserved label changes, the `alias=` shorthand selecting the renamed alias, the count CEL bound rejecting an out-of-range integer at the API server, and leader failover with reconciliation resuming under the new leader.

## Special notes for your reviewer

- Commits are sliced for review: parser → derivation → reconciler → wiring → drift/cleanup hardening → validation fixes → a commit that takes the wiring back out (kept as its own commit so the removal is reviewable and the integration change can revert it) → the port to controller-runtime under `pkg/v1/`, per review → a commit that moves the rolling-update and tombstone fixes out to #898/#899 → the merge that stacks this on #901 → the swap of the private spec accessor for #827's `GetSpec()` → the final review pass (watch predicate, test gaps). Squash-merge flattens all of it.
- **For the integration change, not this PR**: annotation-based placement produces `PlacementPolicy` objects; the controllers that consume them are gated by `--enable-placement-policy-apis` on #820's branch. Enabling annotation-based placement without that flag would leave generated policies sitting inert with no error anywhere, so the integration change should add a startup guard (refuse to start with one flag and not the other). #820's rebase also owes the migration of its private naming helpers to `pkg/utils/naming` introduced here.
- Open question parked in #785's discussion: the `whenUnfulfilled` default noted above. The alias-uniqueness check landed as a webhook warning in #901.


ytimocin:
Ran this branch on real AKS to cover what Kind can't (live admission webhook, guard rail, multi-replica leader election).

**Setup**: 3 AKS clusters (`Standard_B2s_v2`, eastus2euap) + ACR; hub-agent built from this branch (`make TARGET_ARCH=amd64 …` — note `Makefile:22` overrides the env var with the host arch, separate issue); chart installed with `enableAnnotationBasedPlacement=true`, webhook on via cert-manager, `denyModifyMemberClusterLabels=true`, 3 replicas; `placement.kubefleet.dev` CRDs applied from `config/crd/bases/`.

| Scenario | Result |
| --- | --- |
| Annotate ConfigMap → `PlacementPolicy` + `PlacementPolicyCreated` event | ✅ |
| Edit generated policy → restored by the policy watch | ✅ |
| Delete generated policy → recreated | ✅ |
| Invalid annotation → warning event, previous policy left standing | ✅ |
| Remove annotation → policy deleted | ✅ |
| `count: 1000` (bare int) → rejected by the new CEL rule | ✅ |
| MemberCluster created → `kubefleet.dev/cluster-alias` seeded under the live guard rail | ✅ |
| Admin renames alias → allowed, never reverted by later reconciles | ✅ |
| Non-reserved label change → still denied (guard rail active) | ✅ |
| `alias=<renamed>` annotation selects the renamed alias | ✅ |
| Leader pod deleted → new leader elected, annotation reconciled (~60s informer sync on 2-vCPU nodes) | ✅ |

Also verified startup fail-fast: with the flag on and the CRDs absent, the agent exits with `unable to find the CRD that annotation based placement requires` instead of starting degraded. Rig torn down afterwards.

charts/hub-agent/README.md
```diff
@@ -109,6 +109,7 @@ _See [helm install](https://helm.sh/docs/helm/helm_install/) for command documen
 | `enableClusterInventoryAPI` | Enable cluster inventory APIs | `true` |
 | `enableStagedUpdateRunAPIs` | Enable staged update run APIs | `true` |
 | `enableEvictionAPIs` | Enable eviction APIs | `true` |
+| `enableAnnotationBasedPlacement` | Keep a placement policy in sync with the `kubefleet.dev/cluster-selectors` annotation on a resource. Requires the `placement.kubefleet.dev/v1alpha1` CRDs, which this chart does not install. | `false` |
```
ytimocin: We should mention which chart installs the necessary CRDs.

cmd/hubagent/main.go
```diff
@@ -171,6 +173,7 @@ func main() {
 			NetworkingAgentsEnabled: opts.ClusterMgmtOpts.NetworkingAgentsEnabled,
 			MaxConcurrentReconciles: int(math.Ceil(float64(opts.PlacementMgmtOpts.MaxFleetSize) / 100)), //one member cluster reconciler routine per 100 member clusters
 			ForceDeleteWaitTime:     opts.ClusterMgmtOpts.ForceDeleteWaitTime.Duration,
+			SeedClusterAliasLabel:   opts.FeatureFlags.EnableAnnotationBasedPlacement,
```
ytimocin: Is this name aligned with the repository patterns?

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,364 @@
...
+	// EventReasonPolicyUpdated is recorded when an annotation change reaches its generated policy.
+	EventReasonPolicyUpdated = "PlacementPolicyUpdated"
+	// EventReasonPolicyDeleted is recorded when the policy generated for a resource is deleted --
+	// because the annotation was removed, or because the resource stopped being eligible for
+	// placement. The event's message names which.
+	EventReasonPolicyDeleted = "PlacementPolicyDeleted"
+	// EventReasonInvalidAnnotation is recorded when an annotation cannot be parsed. It is a warning
+	// rather than an error on the queue: no amount of retrying makes a malformed annotation parse,
+	// so the event is the only way the user learns that the placement they asked for is not running.
+	EventReasonInvalidAnnotation = "InvalidClusterSelectorsAnnotation"
```
ytimocin, line 62: Do these apply for both **PlacementPolicy** and **ClusterPlacementPolicy**?

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,364 @@
...
+	Recorder record.EventRecorder
+
+	// ShouldPlace reports whether a resource is one KubeFleet places at all, mirroring the filter
+	// the resource watcher applies to its events. The reconciler applies it again because it can be
+	// reached for a resource the watcher would filter out: the watcher reports a resource that
+	// stops passing its filter as a deletion, and the generated policy watch enqueues whatever a
+	// policy names as its owner. A resource that fails the check has its generated policy deleted.
+	//
+	// Left nil, every resource is eligible.
+	ShouldPlace func(source *unstructured.Unstructured) (bool, error)
```
ytimocin, line 149: We should be more clear here. A simple example would help for both cases: **should place** or **should not place**.

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,364 @@
...
+		return ctrl.Result{}, controller.NewUnexpectedBehaviorError(err)
+	}
+	klog.V(2).InfoS("Reconciling annotation-based placement", "obj", clusterWideKey)
+	defer func() {
+		klog.V(2).InfoS("Annotation-based placement reconciliation loop ends", "obj", clusterWideKey, "latency", time.Since(startTime).Milliseconds())
+	}()
+
+	source, err := r.sourceObject(clusterWideKey)
+	switch {
+	case apierrors.IsNotFound(err):
```
ytimocin: Could there be a race condition here where the resource could be manually deleted but would be recreated by another controller?

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,364 @@
...
+		// would keep the policy standing indefinitely. Deleting explicitly is idempotent, so at
+		// worst it beats the collector to an object that was doomed anyway.
+		deleted, err := r.deleteGeneratedPolicy(ctx, clusterWideKey.GroupVersionKind(), clusterWideKey.Namespace, clusterWideKey.Name)
+		switch {
+		case err != nil:
+			klog.ErrorS(err, "Failed to delete the policy generated for a resource that is gone", "obj", clusterWideKey)
+		case deleted:
+			klog.V(2).InfoS("Deleted the policy generated for a resource that is gone", "obj", clusterWideKey)
+		}
+		return ctrl.Result{}, err
```
ytimocin: The error messages must be aligned with the repository standards.

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,364 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if r.ShouldPlace != nil {
+		eligible, err := r.ShouldPlace(source)
+		if err != nil {
+			klog.ErrorS(err, "Failed to decide whether the resource is eligible for placement", "obj", clusterWideKey)
+			return ctrl.Result{}, err
+		}
+		if !eligible {
```
ytimocin, line 188: Could there be a scenario where an object could be ineligible at `time x` but then eligible at `time y`?

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,364 @@
...
+	return nil
+}
+
+// deletePolicy removes the policy generated for a resource that should not have one -- because the
+// annotation was removed, or because the resource is not eligible for placement -- and tells the
+// user which through an event carrying the given cause.
+//
+// The cause is a plain string, never a format: keeping the only format string in the Eventf call
+// below constant is what lets go vet check it.
+func (r *Reconciler) deletePolicy(ctx context.Context, source *unstructured.Unstructured, cause string) error {
```
ytimocin, line 314: Why do we pass the cause to this function? This function deletes the policy and prints a message. That is not aligned with single responsibility principle.

pkg/controllers/annotationplacement/watch.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		// The owner's own namespace is the policy's: a generated policy always lives in the
+		// namespace of the resource it came from, and a cluster-scoped one has none, matching a
+		// cluster-scoped owner.
+		source := &unstructured.Unstructured{}
+		source.SetGroupVersionKind(gv.WithKind(owner.Kind))
+		source.SetNamespace(accessor.GetNamespace())
+		source.SetName(owner.Name)
+		enqueue(source)
+	}
```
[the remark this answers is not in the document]
ytimocin: Fixed — `enqueueGeneratingResources` now enqueues only the owner whose `generatedPolicyName` reproduces this policy's own name, so a foreign owner (or an owner of a kind outside resource discovery) is skipped rather than lazily creating an informer. Added a foreign-owner-not-followed test (mutation-verified).

charts/hub-agent/README.md
ytimocin: Updated — no chart installs the `placement.kubefleet.dev` CRDs yet, so the row now points at `config/crd/bases/` and notes the hub agent fails fast if the flag is on without them.

cmd/hubagent/main.go
ytimocin: `SeedClusterAliasLabel` follows the existing boolean-option field naming on the reconciler; the `"membercluster-controller"` registration string is pre-existing and unchanged. Happy to rename if you had a specific convention in mind.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Yes — the reasons are scope-neutral and apply to both `PlacementPolicy` and `ClusterPlacementPolicy`; the event message names the concrete kind. Clarified in the doc comment.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Added examples to the `ShouldPlace` doc: a Deployment in a user namespace places; a ConfigMap in a skipped namespace like kube-system, or a ReplicaSet a Deployment already owns, does not.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: No corruption. A delete-then-recreate by another controller yields a distinct object (new UID): the delete reconcile removes the policy, the recreate fires a fresh event and generates a policy under the same deterministic, name-derived name via create-or-update, carrying the new owner reference. The name is keyed on identity, not UID, so it converges.

pkg/controllers/annotationplacement/controller.go
ytimocin: These follow the repo convention (lowercase, no trailing punctuation, `%w`-wrapped, `%+v` for keys) — same as rollout/workgenerator. If you meant a specific message, point me at it and I'll align.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Yes — e.g. a ReplicaSet becomes eligible again once orphaned. It is handled: the transitions that matter (owner references, labels) are edits to the resource, so each fires an event that re-runs `ShouldPlace`, and a resource that becomes eligible again while still annotated has its policy regenerated on that event. Noted in the `ShouldPlace` doc.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: The `cause` is the event's human-readable reason. `deletePolicy` deletes the generated policy and records why in one place — the single spot that already derives the policy's generated name. Moving the event emission to the callers would spread that name derivation and the event format across each call site. I kept it cohesive; happy to refactor to caller-emitted events if you'd prefer.

ytimocin:
Pushed a round of fixes addressing review feedback. Each ships with a regression test that fails without the fix:

- **Generated-name sanitization** — characters legal in a resource's own name but not in a Kubernetes object name (e.g. an RBAC name like `system:aggregate-to-admin`) are sanitized before they reach a generated policy's name, which the API server would otherwise reject and hot-loop the reconciler. The identity hash still covers the full, unsanitized identity, so uniqueness is preserved.
- **Ineligible / config-excluded sources** — a source of a disabled API, or in a skipped namespace, no longer keeps a generated policy.
- **Ownership by either provenance marker** — a generated policy is recognized as ours by its owner reference *or* its provenance labels, so editing away one marker no longer leaves it neither repairable nor deletable.
- **Uncached reads** — generated policies are read straight from the API server rather than from a cache fed by a different informer, so an edit or deletion the policy watch delivers first isn't met with a stale object (a missed deletion would otherwise persist indefinitely).
- **Requeue on name conflict** — when a policy authored elsewhere already occupies a resource's generated name, the source is requeued so the requested placement is created once the name is free.
- **Scope from the resolved REST mapping** — a cluster-scoped source whose served API version was retired is no longer misread as namespaced and taken for deleted.
- **Guarded delete** — the generated-policy delete carries a resourceVersion precondition, so a policy replaced between the read and the delete isn't removed.

Also ran a local end-to-end smoke pass (kind hub, feature flag enabled) across 13 scenarios covering all of the above — basic generation, cluster-scoped sources, RBAC-name sanitization, provenance-drift repair, direct policy deletion/recreation, the name-conflict flow, config-excluded and skipped-namespace sources, and source delete/recreate — all green.

cmd/hubagent/options/featureflags.go
```diff
@@ -51,6 +51,17 @@ type FeatureFlags struct {
 	// ResourcePlacement APIs are a set of KubeFleet APIs for processing namespace scoped resource placements.
```
michaelawyu, line 51: Hi Yetkin! Given our current state of development I wonder if it would make sense to have deferred integration rather than enabling the full-range setup right now.

michaelawyu: I feel a bit like this as currently the experience isn't working, but all the extra arguments/charts/default value changes added to the codebase might incur confusion or unexpected side effects, esp. considering that we backport changes and/or cut releases.

michaelawyu: (Not to mention that annotation-based placement is also dependent on the availability of the placement policy APIs)

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -75,6 +76,12 @@ type Reconciler struct {
 	MaxConcurrentReconciles int
 	// the wait time in minutes before we force delete a member cluster.
 	ForceDeleteWaitTime time.Duration
+
+	// SeedClusterAliasLabel controls whether joining member clusters are given the cluster alias
+	// label, seeded from the cluster name. It follows the annotation-based placement feature flag:
+	// the alias exists for that feature's alias= shorthand, and seeding it on fleets that do not
+	// run the feature would relabel every member cluster for nothing.
+	SeedClusterAliasLabel bool
```
michaelawyu: Hi Yetkin! For part, I fear that cluster alias might not necessarily be dependent on the annotation-based placement. Placement policies refer to it as well, plus any future APIs (e.g. migration).

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kfplacementv1alpha1 "github.com/kubefleet-dev/kubefleet/apis/kubefleet.dev/placement/v1alpha1"
+	"github.com/kubefleet-dev/kubefleet/pkg/utils/controller"
```
michaelawyu: Hi Yetkin! I couldn't help noticing that currently the annotation-based placement controller uses the in-house controller framework; this is, obviously, functionally OK, but we do have received feedbacks/suggestions before that we should use the controller runtime controller framework as much as possible (see also the comments by Ahmet from LinkedIn).

These feedbacks/suggestions are actually part of the reasons why currently in the codebase most controllers do not use the in-house controller framework; only the CRP/RP controllers use them IIRC and they do so for legacy reasons.

michaelawyu: The controller runtime controllers are preferred not only because they are (much) more widely used and battle tested, but also due to the fact that they have many built-in capabilities, incl. out-of-box support for a number of metrics.

They support custom work queue as well (see the work applier for an example), which should enable us to enqueue arbitrary information (e.g., keys) for processing when the resource watcher detects changes.

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+	// fails the check has its generated policy deleted.
+	//
+	// Eligibility can change over a resource's life -- an owned ReplicaSet becomes eligible again
+	// once orphaned, a resource stays ineligible while in a skipped namespace. The transitions that
+	// matter are edits to the resource itself (its owner references, its labels), so each fires an
+	// event that re-runs this check; a resource that becomes eligible again and still carries the
+	// annotation has its policy regenerated on that event.
+	//
+	// Left nil, every resource is eligible.
+	ShouldPlace func(source *unstructured.Unstructured) (bool, error)
```
michaelawyu, line 149: Hi Yetkin! We have `ResourceConfig` component available in `utils/apiresources.go` which controls which resources are placeable -> this is the logic that CRP/RP resource selectors are using to check whether a resource can be selected or not. Maybe we should use that here as well.

(Plus, the placement resource snapshot manager needs to re-use the same logic).

michaelawyu: With this being said that component might need some refactoring; we also need to have different built-in rules for CRPs/RPs and PlacementPolicies. Let me create an issue and submit a PR.

michaelawyu: (That component does not check for namespaces either)

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+// logs or records an event about the deletion need not derive it a second time.
+//
+// The policy is read through the uncached reader before it is deleted, so a deletion the policy watch
+// delivered is seen even if a cache has yet to catch up; the read is confined to annotated resources,
+// which are all that reach this controller's queue.
+func (r *Reconciler) deleteGeneratedPolicy(ctx context.Context, gvk schema.GroupVersionKind, namespace, name string) (string, bool, error) {
+	actual := emptyPolicyForScope(namespace)
+	policyName := generatedPolicyName(gvk, namespace, name)
+
+	err := r.UncachedReader.Get(ctx, client.ObjectKey{Namespace: namespace, Name: policyName}, actual)
```
michaelawyu: Hi Yetkin! Is there any particular reason why we have to use the uncached reader for this part? Since the later part is already checking for resource versions, cached client should suffice I think. The system would either fail upon conflicts or become guaranteed to received the next event for processing.

michaelawyu: Also... since owner refs have been correctly configured here, why not let Kubernetes GC handle the cleanup automatically for us?

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+	klog.V(2).InfoS("Reconciling annotation-based placement", "obj", clusterWideKey)
+	defer func() {
+		klog.V(2).InfoS("Annotation-based placement reconciliation loop ends", "obj", clusterWideKey, "latency", time.Since(startTime).Milliseconds())
+	}()
+
+	source, err := r.sourceObject(clusterWideKey)
+	switch {
+	case apierrors.IsNotFound(err):
+		// The resource is gone, and the delete is issued from here rather than left to garbage
+		// collection. The generated policy does carry an owner reference back to the resource, but
```
michaelawyu: Hi Yetkin! I really think we should let Kubernetes GC handle the clean up here. We typically do not guard against byzantine type of errors, and if the users do have other owners set up (assuming that users know what they are doing), should we really be deleting the resource?

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+		// worst it beats the collector to an object that was doomed anyway.
+		_, deleted, err := r.deleteGeneratedPolicy(ctx, clusterWideKey.GroupVersionKind(), clusterWideKey.Namespace, clusterWideKey.Name)
+		switch {
+		case err != nil:
+			klog.ErrorS(err, "Failed to delete the policy generated for a resource that is gone", "obj", clusterWideKey)
+		case deleted:
+			klog.V(2).InfoS("Deleted the policy generated for a resource that is gone", "obj", clusterWideKey)
+		}
+		return ctrl.Result{}, err
+	case meta.IsNoMatchError(err):
```
michaelawyu: The GC can also help cover this case.

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+	switch {
+	case apierrors.IsNotFound(err):
+		return policyName, false, nil
+	case err != nil:
+		return policyName, false, controller.NewAPIServerError(true, err)
```
michaelawyu: Hi Yetkin! A nit: if OK, would you mind using the new error handling package we added earlier?

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+	if err != nil {
+		if meta.IsNoMatchError(err) {
+			// Returned unwrapped: the caller distinguishes a kind that does not exist, which no
+			// retry can fix, and the wrapping below would flatten the error to a string.
+			return nil, err
+		}
+		return nil, controller.NewUnexpectedBehaviorError(fmt.Errorf("failed to get the resource of object %+v: %w", key, err))
+	}
+	gvr := restMapping.Resource
+	if !r.InformerManager.IsInformerSynced(gvr) {
```
michaelawyu: Hi Yetkin! I think this part might have some issue:

michaelawyu: Calling `Lister()` would create a new informer for the resource, however, the informer uninitialized by default. It requires `Start()` to be called on the manager to have the cache populated otherwise all ops would be in vain.

Now, our informer manager is indeed set up to automatically start newly configured informers periodically; however, not all resources can be configured so -> if the resource does not support the LIST/WATCH verb, there is no way for the informer to function.

In fact, in the resource watcher we have specific guard set up to make sure that when discovering new API resources to watch, the watcher must filter out resources that do not support the LIST/WATCH verbs. The call we have here effectively bypasses the guard.

The better way to handle this, I think, might be to fall back to dynamic client to retrieve the resource directly without using cache if the resource to place is not yet registered in the resource watcher.

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+	return source, nil
+}
+
+// syncPolicy creates or updates the policy generated from a resource's annotation, reporting whether
+// the source needs to be looked at again later.
+func (r *Reconciler) syncPolicy(ctx context.Context, source *unstructured.Unstructured, selectors []kfplacementv1alpha1.ClusterSelector) (ctrl.Result, error) {
+	desired := desiredPolicy(source, selectors)
+	actual := emptyPolicyForScope(source.GetNamespace())
+
+	err := r.UncachedReader.Get(ctx, client.ObjectKeyFromObject(desired), actual)
```
michaelawyu: Hi Yetkin! I think that this part does not need to use the uncached client either. If the cached result is stale, the controller will either requeue upon conflicts, or is guaranteed to receive a new event.

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+		// name was chosen deliberately.
+		//
+		// The source is requeued rather than dropped: removing the blocking policy fires no event that
+		// would reach this source (the policy carries no owner reference to it), and the annotation
+		// does not change, so nothing else would ever create the requested policy once the conflict is
+		// cleared. See conflictRequeueAfter.
+		klog.V(2).InfoS("A policy at the generated name was not generated by this controller; leaving it untouched", "obj", klog.KObj(source), "policy", klog.KObj(actual))
+		r.Recorder.Eventf(source, corev1.EventTypeWarning, EventReasonPolicyConflict,
+			"A %s named %s already exists and was not generated from the %s annotation; it was left unchanged", generatedPolicyKind(source.GetNamespace()), actual.GetName(), kfplacementv1alpha1.ClusterSelectorsAnnotation)
+		return ctrl.Result{RequeueAfter: conflictRequeueAfter}, nil
```
michaelawyu: Hi Yetkin! I agree as well that this scenario is extremely unlikely to happen, but when this happens the controller will just requeue forever given the code, which feels a bit tricky. Exponential backoff with errors would be better I guess?

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+
+	if !applyDesiredPolicy(actual, desired) {
+		klog.V(3).InfoS("The generated placement policy is already up to date", "obj", klog.KObj(source), "policy", klog.KObj(actual))
+		return ctrl.Result{}, nil
+	}
+	if err := r.Client.Update(ctx, actual); err != nil {
```
michaelawyu: Hi Yetkin! `controllerutil` has a `CreateOrUpdate` utility that comes with built-in equality check, which does the same logic here, which might be helpful.

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+}
+
+// applyDesiredPolicy brings a live generated policy in line with the desired one, reporting whether
+// anything changed.
+//
+// Only what this controller generates is overwritten: the spec, the provenance labels, and the owner
+// reference to the annotated resource. Labels and annotations that something else added are left
+// alone, so that a generated policy can be labelled by an operator or a GitOps tool without this
+// controller and that tool taking turns undoing each other.
+func applyDesiredPolicy(actual, desired client.Object) bool {
```
michaelawyu: Hi Yetkin! This part of logic can also be simplified with `CreateOrUpdate` utility, if applicable.

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+		if ensureOwnerReference(actual, owner) {
+			changed = true
+		}
+	}
+	return changed
+}
+
+// ensureOwnerReference adds the owner reference to the annotated resource if it is missing or has
+// drifted, leaving any other owner reference in place, and reports whether it changed anything.
+func ensureOwnerReference(policy client.Object, want metav1.OwnerReference) bool {
```
michaelawyu: Hi Yetkin! It seems that `controllerutil.SetOwnerReference` can be helpful here.

apis/kubefleet.dev/placement/v1alpha1/common.go
```diff
@@ -16,6 +16,64 @@ limitations under the License.
...
+//
+// These keys carry the API group rather than the bare KubeFleet domain, since unlike the keys
+// above they are set on objects of this group.
+//
+// The labels exist so that a policy can be found with a List when its generated name is not known;
+// the owner reference on the policy remains the authoritative record of where it came from. Note
+// that ParentNameLabel is lossy: the name of a resource can run to 253 bytes while a label value
+// stops at 63, so for a longer name the label holds a prefix and a hash instead, and selecting on
+// it with the resource's own name matches nothing.
+const (
```
michaelawyu, line 59: Hi Yetkin! For this part, it appears that we are using 1:1 name-based mapping between an annotated resource and its derived placement policy, and the labels are never used for listing -> in that case, would it be simpler if we switch to annotations?

pkg/v1/controllers/annotationplacement/policy.go
```diff
@@ -0,0 +1,268 @@
...
+//
+// whenUnfulfilled deliberately follows the API default of requesting a cluster: asking the platform
+// for a cluster that does not exist yet is the point of these APIs, and the annotation is meant to
+// be the lowest-friction way to do so. The grammar has no slot for the field, so guarding against
+// a selector nothing can ever satisfy (a mistyped region, say) is the claim fulfillment layer's
+// job -- quota, approval, expiry -- not a softer default here.
+const (
+	defaultResourceRevisionHistoryLimit int32 = 3
+
+	defaultWhenUnfulfilled = kfplacementv1alpha1.WhenUnfulfilledOptionAddClusterClaim
```
michaelawyu, line 60: Hi Yetkin! I am not 100% sure about this default; its availability depends on whether cluster claims are enabled on the system I believe.

pkg/v1/controllers/annotationplacement/policy.go
```diff
@@ -0,0 +1,268 @@
...
+// listing. Neither property may come at the cost of the third: the hash is taken over the
+// resource's whole identity, including the parts that do not appear in the name at all, so two
+// resources cannot generate one name even when truncation erases what distinguishes them.
+func generatedPolicyName(gvk schema.GroupVersionKind, namespace, name string) string {
+	// The kind and the name are sanitized before they appear in the name: a resource legal for its
+	// own API can carry characters a DNS-1123 object name forbids (an RBAC name like
+	// system:aggregate-to-admin), and copying them through would produce a policy the API server
+	// rejects, hot-looping the reconciler. The hash below is taken over the unsanitized identity, so
+	// sanitizing the visible parts costs no uniqueness.
+	kindSegment := naming.Truncate(naming.Sanitize(gvk.Kind), maxKindSegmentLength)
```
michaelawyu, line 76: Hi Yetkin! Just a nit: does `Kind` need sanitization? I recall it's alphanumeric only. Of course there's no hurt running the name through the function

pkg/controllers/annotationplacement/policy.go
```diff
@@ -0,0 +1,268 @@
...
+//
+// The one residual gap is a policy that loses both markers at once, which is then indistinguishable
+// from a foreign object and abandoned. It is reachable without malice -- a tool that rewrites the
+// whole object, such as a plain kubectl apply of a hand-kept manifest or a non-server-side-apply
+// Update, drops metadata.labels and metadata.ownerReferences together -- but a policy this controller
+// generated is not one a user is expected to manage that way, and requiring both markers to survive
+// is the accepted price of never touching a policy that is genuinely someone else's.
+func isGeneratedFor(policy client.Object, gvk schema.GroupVersionKind, name string) bool {
+	sourceRef := metav1.OwnerReference{APIVersion: gvk.GroupVersion().String(), Kind: gvk.Kind, Name: name}
+	for _, ref := range policy.GetOwnerReferences() {
```
michaelawyu: Hi Yetkin! This feels like `controllerutil` can help, too (`HasOwnerReference`).

pkg/v1/controllers/annotationplacement/policy.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return &kfplacementv1alpha1.PlacementPolicy{}
+}
+
+// generatedPolicyKind returns the kind of the policy that emptyPolicyForScope produces for the
+// given namespace, spelled as the kind itself is, so that a message carrying it can be pasted
+// straight into a kubectl command against the right resource. It asks emptyPolicyForScope rather
+// than repeating its namespace check, which keeps the scope decision in the one place that
+// function's contract promises.
+func generatedPolicyKind(namespace string) string {
```
michaelawyu, line 250: Hi Yetkin! Just a nit: would this just be `if namespace == "" return "ClusterPlacementPolicy" else return "PlacementPolicy"`?

pkg/v1/controllers/annotationplacement/policy.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return "PlacementPolicy"
+}
+
+// policySpec returns a pointer to the spec of a generated policy, whichever scope it has, so that
+// callers can read and write the spec without repeating the scope distinction.
+//
+// It returns nil for any other type. Callers within this package only ever pass objects that
+// emptyPolicyForScope produced, so a nil here means the two have fallen out of step.
+func policySpec(policy client.Object) *kfplacementv1alpha1.PlacementPolicySpec {
```
michaelawyu: Hi Yetkin! Maybe the accessor/adapter PR might be of help here.

pkg/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
...
+		return ctrl.Result{}, nil
+	}
+	if err := r.Client.Update(ctx, actual); err != nil {
+		klog.ErrorS(err, "Failed to update the generated placement policy", "obj", klog.KObj(source), "policy", klog.KObj(actual))
+		return ctrl.Result{}, controller.NewAPIServerError(false, err)
+	}
+	klog.V(2).InfoS("Updated the generated placement policy", "obj", klog.KObj(source), "policy", klog.KObj(actual))
+	r.Recorder.Eventf(source, corev1.EventTypeNormal, EventReasonPolicyUpdated,
+		"Updated the %s %s from the %s annotation", generatedPolicyKind(source.GetNamespace()), actual.GetName(), kfplacementv1alpha1.ClusterSelectorsAnnotation)
+	return ctrl.Result{}, nil
```
michaelawyu: Hi Yetkin! I noticed that the name of the generated placement policy is nowhere for the users to find. The proposal mentioned about users being able to discover the name via annotation; we can take a different approach if there is a better way as well.

pkg/resourcewatcher/event_handlers.go
```diff
@@ -60,6 +62,9 @@ func (d *ChangeDetector) onResourceAdded(obj interface{}) {
 	klog.V(3).InfoS("A resource is added", "obj", klog.KObj(metaInfo),
 		"gvk", runtimeObject.GetObjectKind().GroupVersionKind().String())
 	d.ResourceChangeController.Enqueue(obj)
+	if annotationplacement.HasClusterSelectorsAnnotation(metaInfo) {
```
michaelawyu: Hi Yetkin! For this part, we might also need to check if the resource can be placed.

pkg/webhook/validation/uservalidation.go
```diff
@@ -94,7 +95,13 @@ func ValidateFleetMemberClusterUpdate(currentMC, oldMC clusterv1beta1.MemberClus
 	}
 
 	isLabelUpdated := isMapFieldUpdated(currentMC.GetLabels(), oldMC.GetLabels())
-	if isLabelUpdated && !isUserInGroup(userInfo, mastersGroup) && shouldDenyLabelModification(currentMC.GetLabels(), oldMC.GetLabels(), denyModifyMemberClusterLabels) {
+	// A whitelisted identity (the hub agent seeds the alias) or a cluster admin may modify the
+	// reserved kubefleet.dev/ labels; ordinary users may not. The distinction matters because,
+	// unlike the member-name label the controller reasserts every reconcile, the cluster alias is
+	// left as it is set -- so a non-admin edit to it would persist and redirect alias-based
+	// placements. (system:masters is already exempted before this runs, at the guard below.)
+	isFleetController := isAdminGroupUserOrWhiteListedUser(whiteListedUsers, userInfo)
```
michaelawyu: Hi Yetkin! Due to the deferred integration discussions, maybe it's best if we skip this part for now and make sure that all the added code are not connected until we are ready to do so.

michaelawyu: P.S., this method concerns only the managed solution, the hub agent of which already uses a master identity IIRC?

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,480 @@
+/*
```
michaelawyu, line 1: Hi Yetkin! Do you think it would be OK if we put all the code related to the new experience in its own directory (e.g., `pkg/v1`) just to keep a clean distinction? We can do some further refactoring as the experience stablizes.

michaelawyu (commented):
Hi Yetkin! I've added some comments, PTAL 🙏

cmd/hubagent/options/featureflags.go
ytimocin: Makes sense, done in e2f5caa3. The flag, chart value/RBAC, setup.go wiring and the resource watcher hooks are all out; the controller merges as dormant code with its unit/envtest coverage. Nothing is reachable in a running hub agent now. The kind e2e suite needs the flag so I parked it on `parked/fep0001-annotation-placement-e2e` for the integration PR once the chain works end to end.

pkg/webhook/validation/uservalidation.go
ytimocin: Dropped along with the rest of the wiring. Good point on the master identity — with that in place the `kubefleet.dev/` exemption may not be needed at all; I'll revisit when alias seeding moves to its own PR and only bring it back if seeding actually gets denied.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
ytimocin: Agreed. With the flag gone nothing sets `SeedClusterAliasLabel` anymore, so it's inert here. I'll pull the seeding and the alias-collision warning out into their own ungated PR so policies and future APIs get the alias regardless of this feature — fine if that lands ahead of this one.

pkg/controllers/annotationplacement/controller.go
ytimocin: Agreed, will port it. One note on shape: the work applier is a single-kind controller with a priority queue, which is not quite our problem — this one reconciles arbitrary discovered kinds, so the plan is a generic `TypedController` over a small GVK+namespace/name request type, the two generated-policy kinds watched normally through the manager cache with a map func back to the source, and (in the integration PR) a `source.TypedChannel` that the resource watcher feeds. Source reads go through the API reader, which also takes care of the `Lister()` issue below.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Sure — I see your PRs already use `pkg/v1/...`, so I will move this to `pkg/v1/controllers/annotationplacement` in the same pass as the controller-runtime port.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Sounds good. I'll keep the eligibility hook as the injected `ShouldPlace` func and switch it to `ResourceConfig` once your refactor lands.

pkg/controllers/annotationplacement/controller.go
ytimocin: Uncached reads are gone with the controller-runtime port (50264305); policies go through the manager cache now. On GC: kept the explicit delete. GC only covers the source being deleted, not the annotation being removed or the source turning ineligible, and a foreign owner reference we deliberately preserve would keep the policy alive. The delete is precondition-guarded and only fires on a policy carrying our provenance.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Kept, same reasoning as on L357: GC doesn't cover annotation removal or ineligibility. A user-added owner is preserved; the delete still only fires when the annotation is gone or the resource is ineligible, and only on a policy we generated.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Same as above. GC is the backstop for the resource-deleted case; the explicit delete covers the other two.

pkg/controllers/annotationplacement/controller.go
ytimocin: Done in 50264305.

pkg/controllers/annotationplacement/controller.go
ytimocin: Right, thanks. Fixed in 50264305: sources are read through `mgr.GetAPIReader()`, policies through the manager cache.

pkg/controllers/annotationplacement/controller.go
ytimocin: Done, cached client in 50264305.

pkg/controllers/annotationplacement/controller.go
ytimocin: Done in 50264305: the conflict returns an error now and gets the queue's exponential backoff.

pkg/controllers/annotationplacement/controller.go
ytimocin: Done, `CreateOrUpdate` in 50264305.

pkg/controllers/annotationplacement/controller.go
ytimocin: Done in 50264305.

pkg/controllers/annotationplacement/controller.go
ytimocin: Done, `SetOwnerReference` in 50264305.

pkg/controllers/annotationplacement/policy.go
ytimocin: Done, `HasOwnerReference` in 50264305.

pkg/v1/controllers/annotationplacement/policy.go
ytimocin: Done in 50264305.

pkg/v1/controllers/annotationplacement/policy.go
ytimocin: `Sanitize` also lowercases, and a Kind is CamelCase, so it has to go through it for a DNS-1123 name.

pkg/v1/controllers/annotationplacement/policy.go
ytimocin: Will do once #827 merges; I'd rather not stack on it.

pkg/v1/controllers/annotationplacement/policy.go
ytimocin: It mirrors the API's `+kubebuilder:default=AddClusterClaim`. Without it `CreateOrUpdate` sees a diff on every pass and the policy watch loops. If claims availability should drive this, the API default is the place to change; this follows it.

apis/kubefleet.dev/placement/v1alpha1/common.go
ytimocin: Kept as labels for now: they're the discovery path, `kubectl get placementpolicy -l placement.kubefleet.dev/parent-name=<name>`. The watch mapping goes by owner reference since the port. Happy to switch if you'd rather not have that listing surface.

pkg/controllers/annotationplacement/controller.go
ytimocin: Two ways today: the `PlacementPolicyCreated` event on the annotated resource names the policy, and the parent labels select it. Writing the name back onto the user's resource is a write surface on arbitrary kinds; I'd defer that to the integration PR if we still want it.

pkg/resourcewatcher/event_handlers.go
ytimocin: Moot for now: the resource watcher hooks came out with the de-wiring. The integration PR will run the eligibility check there too.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
ytimocin: Moved to #901 (seeding always on, plus the collision warning). This PR is stacked on it now.

pkg/webhook/validation/uservalidation.go
ytimocin: Turned out to be needed after all: the label guard only exempts system:masters, and the hub agent SA isn't in it, so seeding would be denied wherever denyModifyMemberClusterLabels is on. #901 makes it a one-line rule instead: kubefleet.dev/ labels follow the same treatment as kubernetes-fleet.io/ ones.

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,405 @@
...
+//
+// A not-found error and a no-match error (the kind itself is unknown) are returned as they are, so
+// that the caller can tell the two apart from a read that failed.
+func (r *Reconciler) sourceObject(ctx context.Context, req Request) (*unstructured.Unstructured, error) {
+	restMapping, err := r.RESTMapper.RESTMapping(req.GroupKind(), req.Version)
+	if meta.IsNoMatchError(err) && req.Version != "" {
+		// The version the request recorded may have been removed while the kind lives on under a
+		// newer served version -- a policy generated for it is still valid, so the served mapping
+		// is tried before concluding the kind is gone.
+		restMapping, err = r.RESTMapper.RESTMapping(req.GroupKind())
```
michaelawyu: Hi Yetkin! This part is really interesting IMO actually. So there are basically 3 scenarios (for now I assume the discussion concerns primarily custom resources; built-in K8s resource version change requires K8s API server upgrades, which may trigger system restarts, so in-between changes are normally non-issues):

michaelawyu: 1) the user updates CRDs to enable a new version, but the old version is still available (via conversion webhook, or via no-op conversion) -> this branch will run normally as the GVK is still present.
2) the user updates CRDs and have an old version disabled (`served=false`); if we happen to read the object in the old version, the mapping might have been gone and the code here will fall back to the currently available version (by using a GK-based check). Now here are two outcomes:
    2a) There is a conversion webhook or no-op conversion is possible; the read will be fine;
    2b) There is no conversion available; with the new version the read will be erred.
3) the user updates CRDs and have an old version dropped directly; the code will fall back to the currently available version, but the data is no longer readable even with the new version.

michaelawyu: For 2a) the code here does guard against errors; however there is another complication about whether it is the user's intention for us to use the new version ->

For example, the new version might not be available on member clusters at all. If we auto-fallback to the new version, and start the placement, it will err out and the outcome might be a bit confusing. It feels a bit like a case where it's best to leave the errors and let user decide?

michaelawyu: And there is another complication -> we run API type discovery periodically and we explicitly request server perferred versions. So, if we do see an older version, it is guaranteed that a new informer would be set up and populated and the object will be re-processed in the new version any way 😂

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,405 @@
...
+	source := &unstructured.Unstructured{}
+	source.SetGroupVersionKind(restMapping.GroupVersionKind)
+	// Scope is read from the mapping just resolved, not from the request: the request's version may
+	// be one the fallback above replaced, and the mapping carries the authoritative scope for the
+	// resource the object is actually read through. A cluster-scoped read must not name a namespace.
+	key := client.ObjectKey{Name: req.Name}
+	if restMapping.Scope.Name() != meta.RESTScopeNameRoot {
+		key.Namespace = req.Namespace
+	}
+	if err := r.APIReader.Get(ctx, key, source); err != nil {
```
michaelawyu, line 261: Hi Yetkin! For this part, is the uncached read a must? If there're a large number of annotated placements, re-syncs/restarts might trigger a large volume of quorum reads.

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,405 @@
...
+	source, err := r.sourceObject(ctx, req)
+	switch {
+	case apierrors.IsNotFound(err), meta.IsNoMatchError(err):
+		// The resource is gone, or its whole kind is (the CRD was removed; the generated policy
+		// watch enqueues a stale policy's owner). The delete is issued from here rather than left
+		// to garbage collection: the collector removes a dependent only once every owner is gone,
+		// and the merge deliberately preserves owner references that other parties added -- any
+		// live one of which would keep the policy standing indefinitely. Deleting explicitly is
+		// idempotent, so at worst it beats the collector to an object that was doomed anyway.
+		_, deleted, err := r.deleteGeneratedPolicy(ctx, req.GroupVersionKind, req.Namespace, req.Name)
```
michaelawyu: Hi Yetkin! For this part, there is a corner case I fear:

Since we do not add finalizers to the annotated object, the deletion now relies on the delivery of Delete events. In normal ops this is fine, but now let's say:

a) the user annotates an object for placement -> b) KubeFleet creates the placement policy -> c) the user deletes the object -> d) before KubeFleet gets the Delete event, our agent goes down (an upgrade occurs, or the system errs) -> e) the agent reboots, but it will never see any Delete events; re-sync cannot help either as the object is already gone from the system.

michaelawyu: This is partly the reason why we sometimes need to rely on owner refs, as it will do free cleanup for us in a (relatively speaking) reliable, eventually consistent manner.

pkg/v1/controllers/annotationplacement/controller.go
```diff
@@ -0,0 +1,405 @@
...
+		return ctrl.Result{}, r.deletePolicy(ctx, source, "the "+kfplacementv1alpha1.ClusterSelectorsAnnotation+" annotation was removed")
+	}
+
+	selectors, err := parseClusterSelectors(value)
+	if err != nil {
+		// The annotation is the user's to fix, so the failure is reported to the user and the
+		// request is dropped. Any policy generated from an earlier, valid annotation is deliberately
+		// left standing: the desired state is now unknown, and tearing down a running placement is
+		// a worse answer to a typo than leaving the last one the user did express.
+		klog.V(2).InfoS("The annotation cannot be parsed", "obj", req, "err", err)
```
michaelawyu, line 218: Hi Yetkin! This one should be an error log I assume?

pkg/v1/controllers/annotationplacement/parser.go
```diff
@@ -0,0 +1,226 @@
+/*
```
michaelawyu, line 1: Hi Yetkin! Just a nit: the error handling part of this file might need some categorization? It's OK if you'd like to address this separately

pkg/v1/controllers/annotationplacement/controller.go
michaelawyu, line 180: Yeah, we should probably just say that it's a KubeFleet managed object and we will always delete it. 😂

pkg/v1/controllers/annotationplacement/policy.go
```diff
@@ -0,0 +1,255 @@
...
+	namespace := source.GetNamespace()
+
+	policy := emptyPolicyForScope(namespace)
+	policy.SetName(generatedPolicyName(gvk, namespace, source.GetName()))
+	policy.SetNamespace(namespace)
+	policy.SetLabels(parentLabels(gvk, source.GetName()))
+	policy.SetOwnerReferences([]metav1.OwnerReference{parentOwnerReference(source)})
+	*policy.GetSpec() = kfplacementv1alpha1.PlacementPolicySpec{
+		ClusterSelectors:             withAPIDefaults(selectors),
+		ResourceSelectors:            []kfplacementv1alpha1.ResourceSelector{parentResourceSelector(source)},
```
michaelawyu, line 230: Hi Yetkin! In the proposal we mentioned about supporting select dependent resources; it's OK to do that separately if needed though to unblock progress -> I guess we just need a comment or an issue?

pkg/v1/controllers/annotationplacement/watch.go
```diff
@@ -0,0 +1,140 @@
...
+// that never changes again the policy would stay missing or wrong forever.
+//
+// The caller owns what the controller cannot decide for itself: the placement.kubefleet.dev types
+// must be registered in the manager's scheme, the hub agent must hold RBAC to read, create, update,
+// and delete the two policy kinds, and the channel's sender is expected to buffer it to the event
+// rate it produces -- the channel source blocks its dispatcher on a full channel.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, sourceEvents <-chan event.TypedGenericEvent[client.Object]) error {
+	return builder.TypedControllerManagedBy[Request](mgr).
+		Named(controllerName).
+		WatchesRawSource(source.TypedChannel(sourceEvents, handler.TypedEnqueueRequestsFromMapFunc(mapSourceToRequest))).
```
michaelawyu: Hi Yetkin! This part could use a workqueue for de-dup actually; if we use a channel there might be many wasted runs under heavy traffic.

michaelawyu: The channel buffering is also a concern. Sync chan would block resource watcher.

michaelawyu: Of course if you plan to use a workqueue + chan producer it is also good, but maybe a workqueue would be easier.

pkg/webhook/membercluster/membercluster_validating_webhook.go
michaelawyu, line 92: Hi Yetkin! This part might have overlapped with another PR?

michaelawyu (commented):
Thanks, Yetkin! I've added a few comments but I think we are super close. Apologies for the delay in response. Let me know if you would like to sync about anything 🙏

And I will have the resource guarding component done ASAP.

michaelawyu (commented):
Hi Yetkin! I've gotten no further comments and I think the PR is good to go -> but merging is blocked now as it appears, feel free to ping me when it's ready 👌 Thanks 🙏

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Agreed, and your discovery point settles it. Removed the fallback in 140929ac, so the resource is read only under the version the request names. On a no-match it now tells a retired version from a gone kind: if the kind still resolves, the request is dropped and the generated policy kept, since discovery re-reports it under the preferred version; only a gone kind deletes the policy.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Kept, but not for freshness. A cached read of a kind the cache has not seen starts an informer for it, one that never syncs for a kind that cannot be listed or watched. The volume is bounded by annotated resources, and the policy watch now ignores status writes, so steady state does not re-read.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Kept at V(2) on purpose: it is the user's typo, not a controller fault. It already surfaces as a Warning event on the annotated resource and the request is dropped, not retried. At error level every malformed annotation would look like a hub-agent failure.

pkg/v1/controllers/annotationplacement/parser.go
ytimocin: Done in 140929ac: every error leaving the parser is a NewUserError. Messages are unchanged, since they are what the event shows the user.

pkg/v1/controllers/annotationplacement/controller.go
ytimocin: Done, reworded along those lines.

pkg/v1/controllers/annotationplacement/policy.go
ytimocin: Filed #927 and noted it at the selector. Open questions there: which references count, whether following them is opt-in, since the annotator may not be allowed to read a Secret they would cause to be copied, and how the set stays fresh.

pkg/v1/controllers/annotationplacement/watch.go
ytimocin: Done in 140929ac: the channel is gone. SourceQueue.Add writes straight into the controller's own workqueue, so the resource watcher never blocks and repeat reports of one resource collapse into one reconcile. The controller's queue was already deduping downstream of the channel; the real fix was the blocking you pointed out.

pkg/webhook/membercluster/membercluster_validating_webhook.go
ytimocin: Yes. This branch still carried #901's changes because it had not been re-synced since #901 merged. Merged main in; it no longer touches the webhook, label guard or alias constant.

ytimocin:
On sequencing: happy to land this before the resource-guard integration. It is unwired, so the guard's absence changes nothing at runtime, and the integration can follow as its own PR once #931 settles. All your comments from the 18th are addressed above. PTAL.

michaelawyu (approved):
LGTM ;) Thanks, Yetkin!

PR #830, opened 2026-08-18
fix: notify security team for scheduled Trivy findings

britaniar:
The scheduled Trivy workflow can fail after finding vulnerabilities because issue creation assigns the invalid REST assignee `copilot`.

## Changes

- Remove the invalid `copilot` issue assignee.
- Create a new daily Trivy issue or refresh the existing same-day issue with current findings.
- Mention `@kubefleet-dev/kubefleet-secops` so security owners are notified.
- Preserve existing non-scheduled scan failure behavior and `security,trivy` labels.

This applies the same fix pattern as Azure/fleet-networking#401.

## Validation

- `actionlint` v1.7.12 against `.github/workflows/trivy.yml`


sjwaight (approved):
LGTM

Issue #787, opened 2026-08-03
[FEP-0001] Implement the Placement Resource Snapshot Manager


michaelawyu:
Implement placement resource snapshot manager per specs in [FEP-0001](https://github.com/kubefleet-dev/enhancements/tree/main/enhancements/0001-placement-policy).

michaelawyu:
Assigning myself to this one.

PR #844, opened 2026-08-20
feat: [FEP-0001] add placement resource snapshot manager (1/)

michaelawyu:

This pull request introduces a new Placement Resource Snapshot Manager for KubeFleet, providing a robust foundation for managing placement resource snapshots. It also adds new accessor interfaces for placement-related resources, constants for snapshot labeling, and improvements to the informer manager utility. The changes enhance modularity, maintainability, and clarity in handling placement resource snapshots and their indices.

The most important changes are:

**Placement Resource Snapshot Manager:**

* Added a new `Manager` in `pkg/v1/managers/placementresourcesnapshot/manager.go` responsible for managing placement resource snapshots, including setup of custom field indices for efficient querying and per-placement-policy locking for concurrency control.
* Added snapshot creation helpers (`primaryPlacementResourceSnapshot` and `secondaryPlacementResourceSnapshot`) in `pkg/v1/managers/placementresourcesnapshot/snapshots.go` for constructing and labeling snapshot resources with proper controller references.

**Placement Resource Snapshot Labeling and Accessors:**

* Introduced new constants in `apis/kubefleet.dev/placement/v1alpha1/placementresourcesnapshot_types.go` for standardized snapshot labeling and annotation, enabling consistent identification and management of placement resource snapshots and their indices/sub-indices.
* Added accessor interfaces and implementations in `apis/kubefleet.dev/placement/v1alpha1/interface.go` to provide unified access to the spec and status of placement policies, resource snapshots, and bindings for both namespace-scoped and cluster-scoped resources.

**Informer Manager Utility Improvements:**

* Enhanced the informer manager in `pkg/utils/informer/informermanager.go` by documenting side effects of implicit informer creation, adding an `IsInformerSet` method to check if an informer is set up for a resource, and implementing the method for safer informer usage. [[1]](diffhunk://#diff-d6d90ec151044a67c800606e241838c4dd1ca17759b6d34a7916f6ff4c7f80daR33-R40) [[2]](diffhunk://#diff-d6d90ec151044a67c800606e241838c4dd1ca17759b6d34a7916f6ff4c7f80daR53-R55) [[3]](diffhunk://#diff-d6d90ec151044a67c800606e241838c4dd1ca17759b6d34a7916f6ff4c7f80daR167-R174)

Related to #787 

I have:

For size control reasons, tests have been skipped in this PR. Will add in subsequent PRs.

### DIFF
--- a/pkg/utils/informer/informermanager.go
+++ b/pkg/utils/informer/informermanager.go
@@ -30,6 +30,14 @@ import (
 	ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
 )
 
+// Note (chenyu1): many methods in this utility, such as IsInformerSynced and Lister, will implicitly create an informer
+// for the queried resource if one does not exist already. This might have side effects as such informers
+// will not start until the manager's Start() method is called, provided that such resources have support
+// for LIST/WATCH ops. Normally this is fine as the resource watcher is configured to periodically register
+// all applicable resources in the informer manager, but the gaps between the synchronization might lead to
+// unexpected behaviors (hopefully temporary). For newer code that needs to integrate with the informer manager,
+// consider calling IsInformerSet first to check if an informer has been set up, before calling other methods.
+
 // InformerManager manages dynamic shared informer for all resources, include Kubernetes resource and
 // custom resources defined by CustomResourceDefinition.
 type Manager interface {
@@ -41,6 +49,9 @@ type Manager interface {
 
 	// IsInformerSynced checks if the resource's informer is synced.
 	IsInformerSynced(resource schema.GroupVersionResource) bool
+
+	// IsInformerSet returns if an informer has been set up for the given resource.
+	IsInformerSet(gvk schema.GroupVersionKind) bool
 
 	// Start will run all informers, the informers will keep running until the channel closed.
 	// It is intended to be called after create new informer(s), and it's safe to call multi times.
@@ -151,6 +162,14 @@ func (s *informerManagerImpl) AddStaticResource(resource APIResourceMeta, handle
 func (s *informerManagerImpl) IsInformerSynced(resource schema.GroupVersionResource) bool {
 	// TODO: use a lazy initialized sync map to reduce the number of informer sync look ups
 	return s.informerFactory.ForResource(resource).Informer().HasSynced()
+}
+
+func (s *informerManagerImpl) IsInformerSet(gvk schema.GroupVersionKind) bool {
+	s.resourcesLock.RLock()
+	defer s.resourcesLock.RUnlock()
+
+	_, ok := s.apiResources[gvk]
+	return ok
 }
 
 func (s *informerManagerImpl) Lister(resource schema.GroupVersionResource) cache.GenericLister {
--- a/pkg/v1/managers/placementresourcesnapshot/manager.go
+++ b/pkg/v1/managers/placementresourcesnapshot/manager.go
@@ -17,14 +17,12 @@ limitations under the License.
 package placementresourcesnapshot
 
 import (
-	"context"
 	"fmt"
 	"hash/fnv"
 	"sync"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -35,34 +33,6 @@ import (
 
 const (
 	managerName = "placementresourcesnapshot"
-)
-
-const (
-	// The two custom fields that are used to index placement resource snapshots in the cache.
-	//
-	// Important: the placement resource snapshot manager runs under the assumption that proper custom fields
-	// have been added and indexed in the cache when running. Failure to complete such prior setup **before
-	// the manager starts** will result in unexpected behaviors. Make sure that the manager uses the client
-	// provided by the hub controller manager, and `SetupWithManager` is called before the manager starts.
-
-	// ownedByAndSubIndexedCustomFieldName is the name of the custom field that indexes placement resource
-	// snapshots by their owner placement policies and their sub-indices.
-	//
-	// This is added to help the manager retrieve all primary placement resource snapshots (i.e., those
-	// with a sub-index of 0) associated with a placement policy.
-	ownedByAndSubIndexedCustomFieldName = "ownedByWithSubIndex"
-	// ownedByAndIndexedCustomFieldName is the name of the custom field that indexes placement resource snapshots by
-	// their owner placement policies and their indices.
-	//
-	// This is added to help the manager retrieve all placement resource snapshots of a specific index
-	// associated with a placement policy.
-	ownedByAndIndexedCustomFieldName = "ownedByWithIndex"
-
-	// The format of the custom field values for the two custom fields above.
-	//
-	// Note that slashes are used to avoid unexpected collisions.
-	ownedByAndSubIndexedCustomFieldFmt = "%s/%s"
-	ownedByAndIndexedCustomFieldFmt    = "%s/%s"
 )
 
 const (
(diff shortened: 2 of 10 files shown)


michaelawyu:
Important:

* The PR depends on the accessor interface PR to be merged (#827); will rebase when that PR is merged.
* Tests are omitted from this PR due to size control reasons; they will be submitted separately.
* Certain logic (block reserved namespaces/resources, strip fields from specific resource types) is omitted form this PR; will submit them separately.

pkg/v1/managers/placementresourcesnapshot/manager.go
```diff
@@ -0,0 +1,98 @@
...
+	placementPolicyKey := fmt.Sprintf(placementPolicyKeyFmt, placementPolicy.GetNamespace(), placementPolicy.GetName())
+
+	hasher := fnv.New32a()
+	hasher.Write([]byte(placementPolicyKey))
+
+	slot := int(hasher.Sum32() % m.muSlotCnt)
+	m.mus[slot].Lock()
+}
+
+func (m *Manager) releaseLock(placementPolicy placementv1alpha1.PlacementPolicyAccessor) {
```
ytimocin, line 90: No race here, the slot is a pure function of the key. Two things worth a comment though: the mutex is held across every API call in a pass, and unrelated policies that hash to the same slot serialize. A single `slot(policy) *sync.Mutex` helper would also drop the duplicated hashing.

pkg/v1/managers/placementresourcesnapshot/ops.go
```diff
@@ -0,0 +1,557 @@
...
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	placementv1alpha1 "github.com/kubefleet-dev/kubefleet/apis/kubefleet.dev/placement/v1alpha1"
+	errors "github.com/kubefleet-dev/kubefleet/pkg/utils/errors"
+	"github.com/kubefleet-dev/kubefleet/pkg/v1/utils/fieldindexers"
+)
+
+const (
+	dryRunAnnotationKey = "kubefleet.dev/dry-run"
```
ytimocin: This can be a const of higher level... As in more project wide...

pkg/v1/managers/placementresourcesnapshot/ops.go
```diff
@@ -0,0 +1,557 @@
...
+	return createdSnapshots, true, nil
+}
+
+// SnapshotResourcesIfLatestSnapshotIsNotUpToDate checks for the latest placement resource snapshot(s) associated
+// with a placement policy. If the latest snapshot is not up-to-date or no snapshots are present, it creates a
+// new placement resource snapshot; otherwise, it returns the current latest placement resource snapshot(s).
+//
+// This method is exposed for rollout purposes, where a controller may need to create a new placement resource
+// snapshot per user requests to complete a rollout.
+func (m *Manager) SnapshotResourcesIfLatestSnapshotIsNotUpToDate(
```
ytimocin, line 99: These two are ~90% identical. One `snapshot(ctx, policy, createIfStale bool)` would cover both.

pkg/v1/managers/placementresourcesnapshot/ops.go
```diff
@@ -0,0 +1,557 @@
...
+			return nil, errors.NewAPIServerError(err, "failed to list placement resource snapshots", true)
+		}
+		subIndexedSnapshots = make([]placementv1alpha1.PlacementResourceSnapshotAccessor, len(placementResourceSnapshotList.Items))
+		for i := range placementResourceSnapshotList.Items {
+			subIndexedSnapshots[i] = &placementResourceSnapshotList.Items[i]
+		}
+	}
+	// Sort the sub-indexed snapshots by their sub-indices.
+	sortErrs = nil
+	sort.Slice(subIndexedSnapshots, func(i, j int) bool {
```
ytimocin, line 264: Collecting errors inside a sort comparator is awkward, and this block appears twice. Parse the index labels into ints first, then `slices.SortFunc`.

pkg/v1/managers/placementresourcesnapshot/resources.go
```diff
@@ -0,0 +1,419 @@
...
+	// The format is `[API-GROUP]/[API-VERSION]/[KIND]/[NAMESPACE]/[NAME]`, where
+	// `[API-GROUP]` is the API group of the resource, `[API-VERSION]` is the API version of the resource,
+	// `[KIND]` is the kind of the resource, `[NAMESPACE]` is the namespace of the resource,
+	// and `[NAME]` is the name of the resource.
+	//
+	// Note that for cluster-scoped resources, the `[NAMESPACE]` segment will be empty.
+	resourceUniqueIdStrFmt = "%s/%s/%s/%s/%s"
+)
+
+func (m *Manager) retrieveAndHashSelectedResources(
```
ytimocin, line 62: The dedupe block is copy-pasted into both selector arms. Collect all candidates first and dedupe once after the loop. `resource.Identifier` is comparable, so it can be the set key directly instead of the formatted id string.

pkg/v1/managers/placementresourcesnapshot/resources.go
```diff
@@ -0,0 +1,419 @@
...
+	snapshottedResource, err := snapshotResource(resource)
+	if err != nil {
+		return placementv1alpha1.SnapshottedResource{},
+			errors.Wraps(err, "failed to snapshot selected resource", "manager", managerName,
+				"gvr", gvr, "namespace", namespace, "name", resourceSelector.Name)
+	}
+	return snapshottedResource, nil
+}
+
+func (m *Manager) retrieveResourcesByLabelSelector(
```
ytimocin, line 213: These two share ~40 lines (GVK→GVR, namespace pick, informer-or-direct). One helper that branches only on the final Get/List would halve them.

pkg/v1/managers/placementresourcesnapshot/resources.go
```diff
@@ -0,0 +1,419 @@
...
+
+func splitResourcesIntoSizeControlledGroups(resources []placementv1alpha1.SnapshottedResource) ([][]placementv1alpha1.SnapshottedResource, error) {
+	if len(resources) == 0 {
+		// Return one single empty group.
+		return [][]placementv1alpha1.SnapshottedResource{{}}, nil
+	}
+
+	var groups [][]placementv1alpha1.SnapshottedResource
+	// Pre-allocate with a reasonably guessed initial capacity.
+	currentGroup := make([]placementv1alpha1.SnapshottedResource, 0, 10)
```
ytimocin, line 385: Capacity guess. `var currentGroup []placementv1alpha1.SnapshottedResource` is enough.

pkg/v1/managers/placementresourcesnapshot/uniquename.go
```diff
@@ -0,0 +1,146 @@
...
+
+	truncatedIdxStr := strconv.Itoa(idx)
+	if len(truncatedIdxStr) > availablePerSeg {
+		truncatedIdxStr = truncatedIdxStr[:availablePerSeg]
+	}
+
+	return fmt.Sprintf(PrimaryPlacementResourceSnapshotNameWithHashFmt, truncatedPlacementPolicyName, truncatedIdxStr, hash), nil
+}
+
+func uniqueNameForSecondaryPlacementResourceSnapshot(placementPolicyName string, idx int, subIdx int) (string, error) {
```
ytimocin, line 111: Identical to the primary version except for the index segment. One `uniqueName(policyName, idxSeg string) string` covers both. Neither can actually return an error. And the even split truncates a policy name to ~110 chars while the index needs at most 4; give the name whatever is left.

pkg/v1/utils/fieldindexers/indexers.go
```diff
@@ -0,0 +1,213 @@
...
+		return []string{v}
+	}); err != nil {
+		return errors.NewUnexpectedError(err, "failed to index cluster placement resource snapshots by owner and sub-index")
+	}
+	return nil
+}
+
+// IndexClusterPlacementResourceSnapshotOwnedByAndIndexedField indexes cluster placement resource snapshots
+// by their owner label value and index label value.
+func IndexClusterPlacementResourceSnapshotOwnedByAndIndexedField(ctx context.Context, fieldIdxer client.FieldIndexer) error {
```
ytimocin, line 160: Four copies of the same indexer. One helper over `client.Object` taking the field name and the label keys, plus a loop in `SetupWithManager`.

pkg/v1/managers/placementresourcesnapshot/ops.go
```diff
@@ -0,0 +1,557 @@
...
+		nextSnapshotIdx = lastSeenSnapshotIdx + 1
+	}
+
+	// Clean up orphaned secondary placement resource snapshots (if any).
+	//
+	// Due to the inability to create multiple placement resource snapshots with the same index in a
+	// transactional manner, it is possible that the manager has already created a few secondary placement
+	// resource snapshot in a previous pass. In this case, the manager should delete the existing snapshots
+	// (its content might be outdated, and the object spec is immutable anyway) before creating new ones.
+	acted, err := m.cleanUpOrphanedSecondarySnapshots(ctx, placementPolicy, nextSnapshotIdx)
```
ytimocin, line 421: This can delete live secondaries. They are created before the primary (L446, then L485), so the cache has a window where index N's secondaries are visible and its primary is not. A reconcile in that window (typically the very next one) still sees N-1 as latest, computes `nextSnapshotIdx = N`, passes the "no sub-index 0" check, and deletes the secondaries the just-created primary N refers to. `retrieveLatestSnapshot` then fails permanently with "fewer than the count label indicates", and nothing repairs N because the next index is N+1. The retry can also recreate secondaries at N with the new hash while primary N keeps the old one.

The dry-run in `isSnapshotUpToDate` guards the hash-match path against exactly this staleness; the create path needs the same guard. Cheapest: before deleting anything at N, confirm through the API reader (a `PartialObjectMetadata` Get) that no primary N exists, and treat "exists" as the same transient error.

pkg/v1/managers/placementresourcesnapshot/resources.go
```diff
@@ -0,0 +1,419 @@
...
+		Kind:    resourceSelector.Kind,
+	}
+
+	// Convert the GVK to a GVR using the REST mapper.
+	mapping, err := m.restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
+		return placementv1alpha1.SnapshottedResource{}, errors.NewUnexpectedError(err, "failed to map GVK to GVR",
+			"manager", managerName, "gvk", gvk)
+	}
+	gvr := mapping.Resource
```
ytimocin, line 156: `mapping.Scope` is computed and never used. Namespace comes from the policy or the selector, so a `ClusterPlacementPolicy` selecting a namespaced kind with no `namespace` does `Lister(gvr).Get(name)` on a namespaced informer and gets NotFound forever (the label path lists across all namespaces), while a `PlacementPolicy` selecting a cluster-scoped kind does `ByNamespace(ns).Get(name)` and misses it. Route on `mapping.Scope.Name()`: a cluster-scoped kind ignores the namespace; a namespaced kind with no namespace on a cluster policy is a user error, not a retry.

pkg/v1/managers/placementresourcesnapshot/resources.go
```diff
@@ -0,0 +1,419 @@
...
+	if m.hubDynamicInformerManager.IsInformerSet(gvk) && m.hubDynamicInformerManager.IsInformerSynced(gvr) {
+		// An informer for the selected resource has been set up and synced; proceed to retrieve the resource from the cache.
+		var obj runtime.Object
+		if namespace == "" {
+			obj, err = m.hubDynamicInformerManager.Lister(gvr).Get(resourceSelector.Name)
+		} else {
+			obj, err = m.hubDynamicInformerManager.Lister(gvr).ByNamespace(namespace).Get(resourceSelector.Name)
+		}
+		if err != nil {
+			return placementv1alpha1.SnapshottedResource{}, errors.NewAPIServerError(err, "failed to get selected resource", true,
```
ytimocin, line 178: NotFound is fatal here. v1beta1 uses `IgnoreNotFound` on a name selector so one absent object does not block the whole placement. Here the policy never gets a snapshot and requeues forever with nothing visible to the user. If the change is intentional it needs a condition.

pkg/v1/managers/placementresourcesnapshot/resources.go
```diff
@@ -0,0 +1,419 @@
...
+			return nil, errors.Wraps(err, "failed to snapshot selected resource", "manager", managerName,
+				"gvr", gvr, "namespace", namespace, "name", resource.GetName())
+		}
+	}
+	return snapshottedResources, nil
+}
+
+// snapshotResource removes fields that are not needed in a snapshot from an unstructured resource and converts it
+// into a SnapshottedResource.
+func snapshotResource(resource *unstructured.Unstructured) (placementv1alpha1.SnapshottedResource, error) {
```
ytimocin, line 312: This is `generateRawContent` from `pkg/utils/controller/resource_selector_resolver.go` minus its special cases: `Service.spec.clusterIP(s)`, `healthCheckNodePort`, per-port `nodePort`, and the Job `controller-uid` selector/labels. Without those a snapshotted Service or Job cannot be applied on a member. Same for `ShouldPropagateObj` (SA token Secrets, `kube-root-ca.crt`, the `default` ServiceAccount, owned ControllerRevisions, managed EndpointSlices). Reuse those, or lift them into `pkg/v1/utils` if the v1beta1 package is off-limits.

pkg/v1/managers/placementresourcesnapshot/resources.go
```diff
@@ -0,0 +1,419 @@
...
+				}
+			}
+		default:
+			return nil, "", errors.NewUserError(nil, "invalid resource selector: neither name nor label selector is specified",
+				"manager", managerName, "resourceSelectorIndex", idx)
+		}
+	}
+
+	// Sort the selected resources to ensure deterministic outcomes.
+	sort.Slice(resources, func(i, j int) bool {
```
ytimocin, line 127: Sorted by the `group/version/kind/ns/name` string. Core group sorts first, but a CR whose group sorts before `apiextensions.k8s.io` lands ahead of its CRD, and Namespaces are not guaranteed first. v1beta1 has `resourceSortOrder` for this. Does the v1 work applier still depend on manifest order?

pkg/v1/managers/placementresourcesnapshot/ops.go
```diff
@@ -0,0 +1,557 @@
...
+		}
+	} else {
+		nextSnapshot = &placementv1alpha1.PlacementResourceSnapshot{
+			ObjectMeta: metav1.ObjectMeta{Name: nextName, Namespace: placementPolicy.GetNamespace()},
+		}
+	}
+
+	// Send a dry-run JSON merge patch that adds the dry-run annotation to the next-index snapshot.
+	patchData := fmt.Appendf(nil, `{"metadata":{"annotations":{%q:%q}}}`, dryRunAnnotationKey, "true")
+	err = m.hubClient.Patch(ctx, nextSnapshot, client.RawPatch(types.MergePatchType, patchData), client.DryRunAll)
```
ytimocin, line 370: A `metav1.PartialObjectMetadata` Get through `mgr.GetAPIReader()` answers the same question without `patch` RBAC on snapshots or a trip through admission, and the dry-run annotation const goes away with it.

ytimocin (commented):
Went through the whole PR. Six inline notes; the two on `ops.go:421` (orphan cleanup can delete live secondaries) and `resources.go:156` (REST mapping scope ignored) are the ones I would fix before merging. Also:

- No tests on the create/cleanup state machine. Both issues above are what an envtest of "create, then reconcile again immediately" catches, so I would rather not defer those.
- The same snapshot label constants land in #848; whichever merges second drops them.
- `retrieveLatestSnapshot` lists every primary a policy has ever had. Is GC / a history limit a later PR in this series?

pkg/v1/managers/placementresourcesnapshot/ops.go
michaelawyu: Hi Yetkin! Yeah, I too agree that this part feels a bit awkward (the pattern is used in the existing implementation as well IIRC), though the error handling part is mostly cosmetic -> in normal ops it should never err. It is true that we can do the validation beforehand or build a wrapper first for each snapshot with parsed index before doing the sort, but that feels a bit redundant considering the extremely unlikely errors. Let me know what you think 🙏

pkg/v1/managers/placementresourcesnapshot/manager.go
michaelawyu: Hi Yetkin! Yeah, we used slots here instead of a per-policy mutex here mostly to avoid GC complications -> Go does not GC used space in maps (even if the keys have been deleted); so if we set up a per-policy mutex, as the agents continue to run, the map in use would constantly grow. To counter this we need to periodically rebuild the map, but this feels a bit over-kill for such a lite implementation. Let me clarify things a bit in a comment.

pkg/v1/managers/placementresourcesnapshot/resources.go
michaelawyu: Hi Yetkin! The work applier has always done its own sorting to support waved manifest processing (this is true in the current implementation as well; we had a plan to let users customize the order, but this has not been implemented yet). The sorting here is just for snapshotting consistency reasons; it has no effect on the actual apply order. Let me add a comment to clarify things.

pkg/v1/managers/placementresourcesnapshot/resources.go
michaelawyu: Hi Yetkin! I took that part out to control the PR size (similarly, the manager does not have resource guard to ignore unsupported resources either) -> let me add a comment here to be clear and it will be added in the subsequent PR.

pkg/v1/managers/placementresourcesnapshot/ops.go
```diff
@@ -0,0 +1,556 @@
...
+	subIndexedSnapshotCntStr := latestPrimarySnapshot.GetLabels()[placementv1alpha1.SubIndexedPlacementResourceSnapshotCountLabelKey]
+	if subIndexedSnapshotCntStr == "1" {
+		// The primary placement resource snapshot is the only snapshot with the latest index; return it.
+		return []placementv1alpha1.PlacementResourceSnapshotAccessor{latestPrimarySnapshot}, nil
+	}
+	subIndexedSnapshotCnt, err := strconv.Atoi(subIndexedSnapshotCntStr)
+	if err != nil {
+		return nil, errors.NewUnexpectedError(err, "failed to convert sub-indexed placement resource snapshot count label to integer",
+			"placementResourceSnapshot", klog.KObj(latestPrimarySnapshot))
+	}
```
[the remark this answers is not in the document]
michaelawyu: N/A; required label.

pkg/v1/managers/placementresourcesnapshot/uniquename.go
```diff
@@ -0,0 +1,145 @@
...
+	// `[SNAPSHOT-INDEX]` is the monotonically increasing index of the snapshot, and
+	// `[SNAPSHOT-SUB-INDEX]` is the monotonically increasing sub-index of the snapshot.
+	//
+	// If the name becomes too long (> 251 characters), KubeFleet will truncate the placement policy name segment
+	// and the `[SNAPSHOT-INDEX]-[SNAPSHOT-SUB-INDEX]` segment as appropriate and add a hash suffix to the name, i.e.,
+	//
+	// `[PLACEMENT-POLICY-NAME-TRUNCATED]-resource-snapshot-[INDEX-TRUNCATED]-[HASH]`,
+	//
+	// where `[HASH]` is the first few characters of the hash of the value
+	// `[PLACEMENT-POLICY-NAME-TRUNCATED]-resource-snapshot-[SNAPSHOT-INDEX]-[SNAPSHOT-SUB-INDEX]`.
```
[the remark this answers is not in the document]
michaelawyu: Fixed.

pkg/utils/informer/informermanager.go
```diff
@@ -42,6 +50,9 @@ type Manager interface {
 	// IsInformerSynced checks if the resource's informer is synced.
 	IsInformerSynced(resource schema.GroupVersionResource) bool
 
+	// IsInformerSet returns if an informer has been set up for the given resource.
+	IsInformerSet(gvk schema.GroupVersionKind) bool
+
```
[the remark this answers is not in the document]
michaelawyu: N/A; tests will be submitted separately.

pkg/v1/managers/placementresourcesnapshot/ops.go
michaelawyu: Hi Yetkin! I merged the two into `SnapshotResources(ctx context.Context, placementPolicy placementv1alpha1.PlacementPolicyAccessor, createOnlyWhenMissing bool)`; let me know if it makes sense.

pkg/v1/managers/placementresourcesnapshot/resources.go
michaelawyu: Hi Yetkin! I made some adjustments. For the identifier part, since the de-dup must ignore the API version input, the string key might still be necessary.

pkg/v1/managers/placementresourcesnapshot/resources.go
michaelawyu: Made some adjustments, PTAL.

pkg/v1/managers/placementresourcesnapshot/resources.go
michaelawyu: Hi Yetkin! Our current codebase tends to pre-allocate whenever possible, usu. with a capacity; it can save a few copies when appending, though the benefits can be relatively limited when the number of items is small. I do not particularly have a strong opinion on this pattern; though for this one, considering that the snapshotted data can be large, some pre-allocation might not necessary be a bad idea.

pkg/v1/managers/placementresourcesnapshot/uniquename.go
michaelawyu: Hi Yetkin! Merging these two into the same method might not read very well IMO. But I've dropped the error part and adjusted the truncation rules -> PTAL.

pkg/v1/utils/fieldindexers/indexers.go
michaelawyu: Hi Yetkin! Yeah, let me tweak this part a bit.

pkg/v1/managers/placementresourcesnapshot/ops.go
michaelawyu: Hi Yetkin! Thanks so much for this; great catch 🙏 Let me fix that.

pkg/v1/managers/placementresourcesnapshot/resources.go
michaelawyu: Hi Yetkin! Yeah, categorization should be better here; let me fix that.

pkg/v1/managers/placementresourcesnapshot/resources.go
michaelawyu: Hi Yetkin! It is an intentional change (primarily because we always require explicit selector for security reasons). The error here is returned to the caller (PlacementPolicy controller, or the StagedUpdateRun controller, as applicable) so that it can be surfaced to the user. We already have a `ResourceCollected` condition defined on the PlacementPolicy API.

pkg/v1/managers/placementresourcesnapshot/ops.go
michaelawyu: Hi Yetkin! Yeah, you are definitely right here; `PartialObjectMetadata` would be more lightweighted. Let me fix that.

ytimocin (approved):

Issue #676, opened 2026-04-27
test: improve coverage for TestPickBindingsToRoll in rollout controller


weng271190436:
## Summary

There is a `// TODO: add more tests` at `pkg/controllers/rollout/controller_test.go:1238` in `TestPickBindingsToRoll`. The existing 33 test cases cover the main happy paths but leave several code paths in `pickBindingsToRoll` and `determineBindingsToUpdate` untested.

### Scheduled → Bound transition gaps
- Scheduled binding with **latest resources** (same snapshot, no overrides) — all current tests use outdated resources
- Scheduled binding with **PickFixed** placement type — only PickAll and PickN are tested
- Multiple scheduled bindings constrained by **MaxSurge=0** — rollout should be blocked
- Scheduled + ready bound mix with **MaxSurge constraining** partial rollout
- Multiple scheduled bindings with **percentage-based MaxSurge** (rounding behavior)

### Untested code paths in `pickBindingsToRoll`
1. **`IsBindingDiffReported` path** (lines ~400, ~449) — no test creates a binding where `IsBindingDiffReported()` returns true, affecting `canBeReadyBindings` membership
2. **Scheduled binding with DeletionTimestamp** — no special handling exists; should verify behavior is intentional
3. **Failed unscheduled binding** (line ~391) — code logs but takes no action; no test validates this path
4. **Negative `maxNumberToRemove`** — possible with specific ratio calculations but never tested
5. **`maxNumberToAdd` exactly zero or negative** — no explicit test
6. **Stale unselected bindings only from `boundingCandidates`** — the append logic in `determineBindingsToUpdate` isn't tested in isolation
7. **Mixed all-state non-ready combinations** — unscheduled + scheduled + bound all non-ready simultaneously
8. **Empty or malformed binding conditions** — edge cases for `HasBindingFailed`, `IsBindingDiffReported`, `isBindingReady`

## Location

- File: `pkg/controllers/rollout/controller_test.go` (line 1238)
- Function under test: `pickBindingsToRoll` in `pkg/controllers/rollout/controller.go`

junnhwan:
Hi! I'd like to work on this issue. I've been studying \pickBindingsToRoll\ and \determineBindingsToUpdate\ in the rollout controller, and I'd like to add tests covering the missing paths (PickFixed placement, Scheduled-to-Bound transitions, MaxSurge constraints, the IsBindingDiffReported path, and a few more). I'll open a PR once the tests are ready.

[...]

PR #847, opened 2026-08-23
test: add scheduled-to-bound coverage for TestPickBindingsToRoll

akshita317:
## Description

`TestPickBindingsToRoll` carries a `// TODO: add more tests`. This adds five table-driven cases for the scheduled-to-bound transition, which the existing cases cover only thinly, and which left one branch of `calculateRealTarget` untested entirely.

Cases added:

| Case | What it pins |
|---|---|
| scheduled binding already at the latest snapshot | every existing case starts from an outdated snapshot, so this path was untested |
| **PickFixed placement type** | `PickFixedPlacementType` appeared **zero** times in this package's tests, so the `ClusterNames` branch of `calculateRealTarget` was never taken |
| multiple scheduled bindings, `maxSurge: 0`, nothing ready | rollout is allowed |
| scheduled binding, `maxSurge: 0`, target already met by ready bound bindings | rollout is held back |
| multiple scheduled bindings, percentage `maxSurge` | the surge budget rounds **up** (50% of 3 → 2, not 1) |

The `maxSurge: 0` pair is deliberately split into two cases. Scheduled bindings are not added to `canBeReadyBindings` (`controller.go:421-431`), so `calculateMaxToAdd` returns the full target number when nothing is ready yet — `maxSurge: 0` only holds a rollout back once ready bindings already meet the target. The two cases document that distinction, which I found surprising when reading the code.

## Coverage

```
                        before    after
calculateRealTarget      70.0%    80.0%
```

The PickFixed case is the one that moves the needle. `pickBindingsToRoll` (98.7%), `determineBindingsToUpdate` (100%) and `calculateMaxToAdd` (100%) were already covered, so the remaining four cases add behavioural assertions — particularly the percentage rounding — rather than new lines. I would rather state that plainly than imply more than the numbers support.

Expected values were derived from reading `pickBindingsToRoll`, `determineBindingsToUpdate`, `calculateMaxToAdd` and `calculateRealTarget` before running anything, not adjusted afterwards to match observed output.

## Testing

```
go test ./pkg/controllers/rollout/ -run TestPickBindingsToRoll -count=1 -v
```

All five new cases pass. All other unit tests in the package pass.

**One pre-existing failure, unrelated to this change:**

```
--- FAIL: TestPickBindingsToRoll/test_unscheduled_bindings_with_different_waitTimes_and_check_the_wait_time_is_correct
    pickBindingsToRoll() = waitTime 0s, want 25s
```

This reproduces identically on unmodified `main` (I stashed this change and re-ran to confirm), and the CI run for `e160c1cd`, the commit this branch is based on, is also red. The case builds bindings with `time.Now()`-relative transition times, so it looks timing-sensitive rather than caused by any recent change. Flagging it here so it is not misattributed to this PR — happy to open a separate issue if that would be useful.

`TestAPIs` was not run locally as it needs an envtest control plane.

## Related

Part of #676. That issue also lists a set of untested paths in `pickBindingsToRoll` itself — `IsBindingDiffReported`, deletion timestamps on scheduled bindings, negative `maxNumberToRemove`, and others. Those are a separate concern from the scheduled-to-bound group and are left for follow-up rather than bundled here, so the `// TODO` comment stays in place.

### DIFF
--- a/pkg/controllers/rollout/controller_test.go
+++ b/pkg/controllers/rollout/controller_test.go
@@ -1236,6 +1236,204 @@ func TestPickBindingsToRoll(t *testing.T) {
 		wantErr                     error
 	}{
 		// TODO: add more tests
+		"test scheduled binding to bound, latest resources and nil overrides - rollout allowed": {
+			allBindingsFunc: func() []*placementv1beta1.ClusterResourceBinding {
+				return []*placementv1beta1.ClusterResourceBinding{
+					generateClusterResourceBinding(placementv1beta1.BindingStateScheduled, "snapshot-1", cluster1),
+				}
+			},
+			latestResourceSnapshotName: "snapshot-1",
+			crp: clusterResourcePlacementForTest("test",
+				createPlacementPolicyForTest(placementv1beta1.PickAllPlacementType, 0),
+				createPlacementRolloutStrategyForTest(placementv1beta1.RollingUpdateRolloutStrategyType, generateDefaultRollingUpdateConfig(), nil)),
+			wantTobeUpdatedBindings: []int{0},
+			wantDesiredBindingsSpec: []placementv1beta1.ResourceBindingSpec{
+				{
+					State:                placementv1beta1.BindingStateBound,
+					TargetCluster:        cluster1,
+					ResourceSnapshotName: "snapshot-1",
+				},
+			},
+			wantNeedRoll: true,
+			wantWaitTime: 0,
+		},
+		"test scheduled binding to bound with PickFixed placement type - rollout allowed": {
+			allBindingsFunc: func() []*placementv1beta1.ClusterResourceBinding {
+				return []*placementv1beta1.ClusterResourceBinding{
+					generateClusterResourceBinding(placementv1beta1.BindingStateScheduled, "snapshot-1", cluster1),
+				}
+			},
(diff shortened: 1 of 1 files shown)


akshita317:
CI note for reviewers: `unit-and-integration-tests` is red, but not from this change.

```
ok    github.com/kubefleet-dev/kubefleet/pkg/controllers/rollout           571.040s
FAIL  github.com/kubefleet-dev/kubefleet/pkg/controllers/bindingwatcher    79.814s
```

`pkg/controllers/rollout` — the only package this PR touches — passes, including the five new cases. The failure is `TestAPIs` in `pkg/controllers/bindingwatcher`, at `watcher_integration_test.go:751`, which this PR does not touch.

Happy to rebase once that settles if you would rather see a green run before reviewing.

sjwaight:
@weng271190436 can you please review given original issue was opened by you. Thanks.

pkg/controllers/rollout/controller_test.go
```diff
@@ -1236,6 +1236,204 @@ func TestPickBindingsToRoll(t *testing.T) {
 		wantErr                     error
 	}{
 		// TODO: add more tests
```
ytimocin, line 1238: this can be removed

weng271190436 (approved):

ytimocin (approved):

Issue #789, opened 2026-08-03
[FEP-0001] Implement the work generator controller for placement policies


michaelawyu:
Implement the work generator per specs in [FEP-0001](https://github.com/kubefleet-dev/enhancements/tree/main/enhancements/0001-placement-policy).

michaelawyu:
Assigning myself to this one.

PR #848, opened 2026-08-25
feat: [FEP-0001] add work generator (1/)

michaelawyu:

This PR adds the work generator implementation as specified in FEP-0001 for the placement policy APIs.

Related to #789 

I have:

N/A; please see the notes.

To control the PR size, tests + some logs/events/monitoring setup have not been included in this PR; they will be submitted later.


michaelawyu:
Note also that this PR depends on #827. Will rebase after it is merged.

pkg/v1/controllers/workgenerator/retrieval.go
```diff
@@ -0,0 +1,181 @@
...
+
+	workList := &placementv1alpha1.WorkList{}
+	listOptions := []client.ListOption{
+		client.InNamespace(memberClusterNamespace),
+		client.MatchingLabels{placementv1alpha1.WorkOwnedByPlacementBindingLabelKey: ownerBindingName},
+	}
+	if err := r.hubClient.List(ctx, workList, listOptions...); err != nil {
+		return nil, errors.NewAPIServerError(err, "failed to list work objects", true)
+	}
+	return workList.Items, nil
```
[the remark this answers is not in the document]
michaelawyu: Fixed.

pkg/v1/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,241 @@
...
+		latency := time.Since(startTime).Milliseconds()
+		klog.V(2).InfoS("Reconciliation ends", "placementBinding", req.NamespacedName, "controller", controllerName, "latency", latency)
+	}()
+
+	// Retrieve the PlacementBinding object.
+	placementBinding, err := r.retrievePlacementBinding(ctx, req.NamespacedName)
+	if err != nil {
+		klog.ErrorS(err, "", "namespacedName", req.NamespacedName, "controller", controllerName)
+		return ctrl.Result{}, errors.Wraps(err, "", "namespacedName", req.NamespacedName, "controller", controllerName)
+	}
```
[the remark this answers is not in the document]
michaelawyu: Fixed.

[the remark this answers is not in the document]
michaelawyu: Fixed.

pkg/v1/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,241 @@
...
+	return &Reconciler{
+		hubClient:    hubClient,
+		parallelizer: parallelizer,
+	}
+}
+
+// TO-DO (chenyu1): switch to field-based indexes for better performance when listing objects.
+
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	startTime := time.Now()
```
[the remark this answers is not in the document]
michaelawyu: N/A -> WIP; tests will be submitted in separate PRs.

pkg/v1/controllers/workgenerator/status.go
```diff
@@ -0,0 +1,219 @@
...
+
+	// Set a false Synchronized condition with the WaitingForSynchronization reason on the placement binding.
+	meta.SetStatusCondition(&placementBindingStatus.Conditions, metav1.Condition{
+		Type:               placementv1alpha1.PlacementBindingCondTypeSynchronized,
+		Status:             metav1.ConditionFalse,
+		ObservedGeneration: placementBinding.GetGeneration(),
+		Reason:             placementv1alpha1.PlacementBindingSynchronizedCondReasonWaitingForSynchronization,
+		Message:            "Waiting for the resources to be synchronized to the target cluster",
+	})
+
```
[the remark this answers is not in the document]
michaelawyu: Note: the available condition would become stale in this situation (`observedGeneration`), consistent with Kubernetes standard condition practices. With that being said I do not mind setting this condition in the `False`/`Unknown` status if we believe it does yield better clarity.

pkg/v1/controllers/workgenerator/works.go
```diff
@@ -0,0 +1,308 @@
...
+	if placementBinding.GetStatus().LastProcessedResourceSnapshotName != nil {
+		lastProcessedSnapshotName = *placementBinding.GetStatus().LastProcessedResourceSnapshotName
+	}
+	primarySnapshotName := placementBinding.GetSpec().ResourceSnapshotName
+
+	if lastProcessedSnapshotName != primarySnapshotName {
+		return false
+	}
+
+	// Check if the sync strategy of the placement binding still matches that on the work objects.
```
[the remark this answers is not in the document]
michaelawyu: `Work` objects in general are considered KubeFleet internal API objects; and for simplicity and performance reasons we typically do not guard against byzantine type of errors (e.g., unexpected mishandling, such as users editing/deleting the work objects manually, leading to system inconsistencies).

This branch is added as a shortcut to save the controller from constantly spending cycles cross-referencing/validating work objects (which includes all the manifests to place). Will add a comment on the code for better clarity.

ytimocin (commented):
One real bug (suspend/resume), lint is red for this PR, and a status-reason question. The unit-job failure is the workapplier AfterSuite teardown timeout, unrelated. Same five snapshot constants as #844; second to merge drops them.

pkg/v1/controllers/workgenerator/works.go
```diff
@@ -0,0 +1,412 @@
...
+		lastProcessedSnapshotName = *placementBinding.GetStatus().LastProcessedResourceSnapshotName
+	}
+	primarySnapshotName := placementBinding.GetSpec().ResourceSnapshotName
+
+	if lastProcessedSnapshotName != primarySnapshotName {
+		return false, nil
+	}
+
+	// Do some sanity checks, just to make sure that the cache is up to date.
+	if len(works) == 0 {
```
ytimocin, line 72: A suspended binding never resumes. Cleanup deletes the works but leaves `status.lastProcessedResourceSnapshotName`, so this branch sees zero works with a matching name and returns the transient error forever. Clear the field in `cleanupWorks`.

pkg/v1/controllers/workgenerator/status.go
```diff
@@ -0,0 +1,251 @@
...
+		Reason:             placementv1alpha1.PlacementBindingAvailableCondReasonWaitingForAvailabilityCheck,
+		Message:            "Waiting for the resources to be checked for availability in the target cluster",
+	})
+
+	// Count the number of manifests in all created/updated work objects.
+	total := 0
+	for idx := range worksToCreateOrUpdate {
+		total += len(worksToCreateOrUpdate[idx].Spec.Manifests)
+	}
+	placementBindingStatus.SelectedResources = ptr.To(int32(total))
```
ytimocin: Lint: G115 here and at L192, `unconvert` at L44-46. Count on `int32` directly.

pkg/v1/controllers/workgenerator/status.go
```diff
@@ -0,0 +1,251 @@
...
+	return nil
+}
+
+func refreshPlacementBindingSyncCond(placementBinding placementv1alpha1.PlacementBindingAccessor, works []placementv1alpha1.Work) {
+	// The binding is synchronized only if every work has been applied and its applied condition is up-to-date.
+	synchronized := true
+	for idx := range works {
+		work := &works[idx]
+		appliedCond := meta.FindStatusCondition(work.Status.Conditions, placementv1alpha1.WorkCondTypeApplied)
+		if !condition.IsConditionStatusTrue(appliedCond, work.GetGeneration()) {
```
ytimocin, line 138: A missing or stale-generation condition is reported as Failed here and at L159, overwriting the Waiting reasons before the agent has run. Keep a "not yet" branch like the v1beta1 incomplete case.

pkg/v1/controllers/workgenerator/status.go
```diff
@@ -0,0 +1,251 @@
...
+				// The manifest has been applied.
+				synced++
+			}
+
+			availableCond := meta.FindStatusCondition(manifest.Conditions, placementv1alpha1.ManifestCondTypeAvailable)
+			switch {
+			case availableCond == nil:
+				// The Available condition has not been set yet; the manifest has not been processed.
+				continue
+			case availableCond.Status != metav1.ConditionTrue:
```
ytimocin, line 238: Any `Available != True` lands in `failedResources`, including a Deployment mid-rollout. Same as v1beta1; if intended, say so in the field doc.

pkg/v1/controllers/workgenerator/cleanup.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return errors.Wraps(err, "failed to generate work name for primary placement resource snapshot")
+	}
+	workForPrimaryResSnapshot := &placementv1alpha1.Work{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: fmt.Sprintf(utils.NamespaceNameFormat, placementBinding.GetSpec().ClusterName),
+			Name:      workName,
+		},
+	}
+	if err := r.hubClient.Delete(ctx, workForPrimaryResSnapshot); err != nil && !apierrors.IsNotFound(err) {
```
ytimocin, line 62: Work names hash policy and cluster, not the binding, and this delete has no ownership check. A replacement binding for the same policy adopts the works, then the old binding's finalizer deletes them. Check the owned-by annotation first, or is that reschedule impossible?

pkg/v1/controllers/workgenerator/derivedfrom.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Verify that all formatter implements the derivedFromSourceFormatter interface.
+var _ derivedFromSourceFormatter = &placementResourceSnapshotDerivedFromSourceFormatter{}
+
+// derivedFromSourceFormatter is an interface that helps format the ID of a source that derives a work object for
+// various use cases, primarily for preparing unique names for work objects.
+type derivedFromSourceFormatter interface {
```
ytimocin, line 24: One implementation, constant `SourceType`, `snapshotNamespacedName` never read. Pass the sub-index string.

pkg/v1/controllers/workgenerator/uniquename.go
```diff
@@ -0,0 +1,133 @@
...
+	workDerivedFromOtherSourcesNameFmt          = "%s-work-%s-%s"
+)
+
+// uniqueNameForWorkDerivedFromPlacementResourceSnapshot generates a unique name for a work object derived from a
+// placement resource snapshot, given the owner placement binding and the snapshot sub-index (0 = primary).
+func uniqueNameForWorkDerivedFromPlacementResourceSnapshot(
+	placementBinding placementv1alpha1.PlacementBindingAccessor,
+	isFromPrimarySnapshot bool,
+	derivedFromSrcFormatter derivedFromSourceFormatter,
+) (string, error) {
```
ytimocin: Always returns a nil error. Return `string`, drop the dead branches at `works.go:248` and `cleanup.go:61`.

pkg/v1/controllers/workgenerator/retrieval.go
```diff
@@ -0,0 +1,184 @@
...
+			return nil, errors.NewAPIServerError(err, "failed to list placement resource snapshots", true)
+		}
+		snapshots = make([]placementv1alpha1.PlacementResourceSnapshotAccessor, len(snapshotList.Items))
+		for i := range snapshotList.Items {
+			snapshots[i] = &snapshotList.Items[i]
+		}
+	}
+
+	// Sort the snapshots by their sub-indices in ascending order.
+	var sortErrs []error
```
ytimocin, line 139: Same error-collecting sort as #844. Parse first, then `slices.SortFunc`.

pkg/v1/controllers/workgenerator/works.go
michaelawyu: Thanks, Yetkin! Great catch; let me fix that.

pkg/v1/controllers/workgenerator/status.go
michaelawyu: Yeah, linter error fixed.

pkg/v1/controllers/workgenerator/status.go
michaelawyu: Hi Yetkin! Yeah, the behavior is consistent with the existing system. I patched the comment a bit to be more clear.

pkg/v1/controllers/workgenerator/status.go
michaelawyu: Hi Yetkin! Yeah, let me fix that.

pkg/v1/controllers/workgenerator/cleanup.go
michaelawyu: Hi Yetkin! The same policy selecting the same cluster is expected to always generate the same binding -> this part might have an implicit coupling on the placement policy controller logic; if there's other plan, let's discuss it.

The current system names bindings with a randomized suffix and takes a (if there are two bindings from the same placement selecting the same cluster, backoff until one of the bindings is gone) stance; but with the hash-based suffix this backoff does not seem to be necessary any more.

pkg/v1/controllers/workgenerator/derivedfrom.go
michaelawyu: Done ✔️

pkg/v1/controllers/workgenerator/uniquename.go
michaelawyu: Done ✔️

pkg/v1/controllers/workgenerator/retrieval.go
michaelawyu: Hi Yetkin! Yeah, I too agree that this part feels a bit awkward (the pattern is used in the existing implementation as well IIRC), though the error handling part is mostly cosmetic -> in normal ops it should never err. It is true that we can do the validation beforehand or build a wrapper first for each snapshot with parsed index before doing the sort, but that feels a bit redundant considering the extremely unlikely errors. Let me know what you think 🙏

ytimocin (approved):

PR #849, opened 2026-08-25
fix: skip access provider population in cluster profiles if no endpoint/CA data exists

### LINKED ISSUE #761: [BUG] ClusterProfile publishes an AccessProvider with an empty cluster.server
### **Describe the bug**
 
Once a `MemberCluster` is joined and property collection succeeds, the ClusterProfile controller always adds a `KubeFleet` entry to `status.accessProviders`. It does this even when the collected properties don't include `k8s.io/cluster-entrypoint`, and in that case the entry goes out with an empty server:
 
```yaml
status:
  accessProviders:
    - name: KubeFleet
      cluster:
        server: ""
```
 
This isn't a timing race. The built-in Azure property provider (currently the only provider the member agent can select) never sets `k8s.io/cluster-entrypoint`, so every cluster using it ends up in this state by default.
 
An empty `server` is arguably worse than publishing no access provider at all. cluster-inventory-api's `BuildConfigFromCP` copies `cluster.server` straight into `rest.Config.Host` without complaint ([config.go](https://github.com/kubernetes-sigs/cluster-inventory-api/blob/85a8c75f5861c2d1f8ed7429dfd7b4b4f916e977/pkg/access/config.go#L124-L200)), and client-go falls back to `localhost` when the host is empty ([url_utils.go](https://github.com/kubernetes/client-go/blob/v0.35.3/rest/url_utils.go#L80-L96)). Since the profile carries CA data, the resulting base URL is `https://localhost`. I verified this locally: `BuildConfigFromCP` returns `host=""` with a nil error, and `DefaultServerUrlFor` resolves it to `https://localhost`. A consumer that trusts the AccessProvider ends up sending requests to its own machine instead of the member cluster.
 
### **Environment**
 
- KubeFleet `main` @ `42a1df03b2597ad55ca771c668d0db64aac1aae0`
- kind clusters `hub` and `cluster-1` (`kindest/node:v1.33.4`), member joined
- Property provider: `azure`
The status-writing code came in with #321 and first shipped in `v0.1.1`; everything from there through `v0.3.1` and current `main` has it.
 
### **To Reproduce**
 
1. From the KubeFleet repository, create one hub and one member cluster, install the locally built agents, and join the member:
```shell
   JOIN_MEMBERS=true MEMBER_CLUSTER_COUNT=1 make setup-clusters
```
 
2. Wait for property collection to succeed:
```shell
   kubectl --context kind-hub wait \
     --for='jsonpath={.status.conditions[?(@.type=="ClusterPropertyCollectionSucceeded")].status}=True' \
     membercluster/kind-cluster-1 \
     --timeout=5m
```
 
3. Confirm the collected properties contain no cluster entry point:
```shell
   kubectl --context kind-hub get membercluster kind-cluster-1 -o json |
     jq '.status.properties["k8s.io/cluster-entrypoint"]'
```
 
   Returns `null`.
 
4. Read the generated ClusterProfile:
```shell
   kubectl --context kind-hub \
     --namespace fleet-system \
     get clusterprofile kind-cluster-1 \
     -o yaml
```
 
   The status contains the empty `accessProviders` entry shown above.
 
### **Root cause**
 
In the reconcile path, the controller appends the AccessProvider first and only afterwards looks up `ClusterEntryPointProperty`; if the property is missing, `Server` just keeps its zero value ([controller.go#L185-L220](https://github.com/kubefleet-dev/kubefleet/blob/42a1df03b2597ad55ca771c668d0db64aac1aae0/pkg/controllers/clusterinventory/clusterprofile/controller.go#L185-L220)).
 
Nothing in the repo actually writes that property. The constant is declared in `pkg/propertyprovider/commons.go`, but the Azure provider's `Collect` never sets it, and the member agent has no other provider to select. On top of that, the installed CRD requires the `server` key to be present but doesn't reject an empty string, so the status update is accepted.
 
One more thing: the existing unit test explicitly asserts that an AccessProvider is created even with an empty property map ([controller_test.go#L68-L88](https://github.com/kubefleet-dev/kubefleet/blob/42a1df03b2597ad55ca771c668d0db64aac1aae0/pkg/controllers/clusterinventory/clusterprofile/controller_test.go#L68-L88)), so the fix will need to update that test too.
 
### **Expected behavior**
 
Don't publish an AccessProvider unless there's a usable endpoint. If `k8s.io/cluster-entrypoint` is missing, empty, relative, or has no host, `status.accessProviders` should be omitted entirely. When a provider does supply a valid endpoint:
 
```yaml
status:
  accessProviders:
    - name: KubeFleet
      cluster:
        server: https://api.example.com:6443
```
 
### **Screenshots**
 
N/A
 
### **Additional context**
 
[cluster-inventory-api #80](https://github.com/kubernetes-sigs/cluster-inventory-api/pull/80) (still open as of this writing) adds CEL validation requiring `server` to be an absolute URL with a host. Once that lands, the API server will reject KubeFleet's current status update outright. So the controller needs to stop writing the empty entry either way. With the new CRD it breaks loudly, and with the old one it silently points consumers at localhost.

michaelawyu:

This pull request primarily refactors import statements for consistency and improves the logic for populating the `AccessProviders` field in the `ClusterProfile` controller. It ensures that the cluster access provider is only set when both the cluster entry point and CA data are available, and updates related tests accordingly.

Fixes #761 

I have:

- [x] UTs
- [x] ITs 

N/A

### DIFF
--- a/pkg/controllers/clusterinventory/clusterprofile/controller.go
+++ b/pkg/controllers/clusterinventory/clusterprofile/controller.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	clientcmdv1 "k8s.io/client-go/tools/clientcmd/api/v1"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	clusterinventory "sigs.k8s.io/cluster-inventory-api/apis/v1alpha1"
@@ -180,8 +181,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return ctrl.Result{}, nil
 }
 
-// fillInClusterStatus fills in the ClusterProfile status fields from the MemberCluster status.
-// Currently, it only fills in the Kubernetes version field.
+// fillInClusterStatus fills in the ClusterProfile status fields from the MemberCluster status,
+// including the Kubernetes version and, when available, the cluster access provider.
 func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *clusterinventory.ClusterProfile) {
 	clusterPropertyCondition := meta.FindStatusCondition(mc.Status.Conditions, string(clusterv1beta1.ConditionTypeClusterPropertyCollectionSucceeded))
 	if !condition.IsConditionStatusTrue(clusterPropertyCondition, mc.Generation) {
@@ -196,27 +197,25 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
 			Kubernetes: k8sversion.Value,
 		}
 	}
-	// Add the class access provider, we only have one so far
-	cp.Status.AccessProviders = []clusterinventory.AccessProvider{
-		{
-			Name: controller.ClusterManagerName,
-		},
-	}
-	// TODO throw and unexpected error if clusterEntryPoint is not found
-	// We don't have a way to get it yet
-	clusterEntry, exists := mc.Status.Properties[propertyprovider.ClusterEntryPointProperty]
-	if exists {
-		klog.V(3).InfoS("Get Kubernetes cluster entry point from member cluster status", "clusterEntryPoint", clusterEntry.Value, "clusterProfile", klog.KObj(cp))
-		cp.Status.AccessProviders[0].Cluster.Server = clusterEntry.Value
-	}
-	// Get the CA Data
-	certificateAuthorityData, exists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
-	if exists {
-		klog.V(3).InfoS("Get Kubernetes cluster certificate authority data from member cluster status", "clusterProfile", klog.KObj(cp))
-		cp.Status.AccessProviders[0].Cluster.CertificateAuthorityData = []byte(certificateAuthorityData.Value)
+
+	// Add cluster access provider, if and only if a cluster entry point and the CA data exist as part of the
+	// cluster properties.
+	clusterEntrypoint, entryPtExists := mc.Status.Properties[propertyprovider.ClusterEntryPointProperty]
+	caData, caDataExists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
+	if entryPtExists && caDataExists {
+		cp.Status.AccessProviders = []clusterinventory.AccessProvider{
+			{
+				Name: controller.ClusterManagerName,
+				Cluster: clientcmdv1.Cluster{
+					Server:                   clusterEntrypoint.Value,
+					CertificateAuthorityData: []byte(caData.Value),
+				},
+			},
+		}
 	} else {
-		// throw an alert
-		_ = controller.NewUnexpectedBehaviorError(fmt.Errorf("cluster certificate authority data not found in member cluster %s status", mc.Name))
+		klog.V(2).InfoS("Cluster entry point and/or CA data is missing; skip adding cluster access provider to cluster profile status",
+			"memberCluster", klog.KObj(mc), "clusterProfile", klog.KObj(cp),
+			"clusterEntryPointExists", entryPtExists, "caDataExists", caDataExists)
 	}
 }
 
--- a/pkg/controllers/clusterinventory/clusterprofile/controller.go
+++ b/pkg/controllers/clusterinventory/clusterprofile/controller.go
@@ -202,7 +202,7 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
 	// cluster properties.
 	clusterEntrypoint, entryPtExists := mc.Status.Properties[propertyprovider.ClusterEntryPointProperty]
 	caData, caDataExists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
-	if entryPtExists && caDataExists {
+	if entryPtExists && caDataExists && len(clusterEntrypoint.Value) > 0 && len(caData.Value) > 0 {
 		cp.Status.AccessProviders = []clusterinventory.AccessProvider{
 			{
 				Name: controller.ClusterManagerName,
@@ -213,7 +213,8 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
 			},
 		}
 	} else {
-		klog.V(2).InfoS("Cluster entry point and/or CA data is missing; skip adding cluster access provider to cluster profile status",
+		cp.Status.AccessProviders = nil
+		klog.V(2).InfoS("Cluster entry point and/or CA data is missing; reset cluster access provider to cluster profile status",
 			"memberCluster", klog.KObj(mc), "clusterProfile", klog.KObj(cp),
 			"clusterEntryPointExists", entryPtExists, "caDataExists", caDataExists)
 	}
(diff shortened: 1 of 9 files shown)


pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -196,27 +197,25 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
...
-	certificateAuthorityData, exists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
-	if exists {
-		klog.V(3).InfoS("Get Kubernetes cluster certificate authority data from member cluster status", "clusterProfile", klog.KObj(cp))
-		cp.Status.AccessProviders[0].Cluster.CertificateAuthorityData = []byte(certificateAuthorityData.Value)
+
+	// Add cluster access provider, if and only if a cluster entry point and the CA data exist as part of the
+	// cluster properties.
+	clusterEntrypoint, entryPtExists := mc.Status.Properties[propertyprovider.ClusterEntryPointProperty]
+	caData, caDataExists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
+	if entryPtExists && caDataExists {
```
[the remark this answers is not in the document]
michaelawyu: Fixed.

pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -196,27 +197,26 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
...
+			},
+		}
 	} else {
-		// throw an alert
-		_ = controller.NewUnexpectedBehaviorError(fmt.Errorf("cluster certificate authority data not found in member cluster %s status", mc.Name))
+		cp.Status.AccessProviders = nil
+		klog.V(2).InfoS("Cluster entry point and/or CA data is missing; reset cluster access provider to cluster profile status",
+			"memberCluster", klog.KObj(mc), "clusterProfile", klog.KObj(cp),
+			"clusterEntryPointExists", entryPtExists, "caDataExists", caDataExists)
 	}
```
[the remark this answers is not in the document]
michaelawyu: Fixed.

pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -196,27 +197,26 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
...
-	certificateAuthorityData, exists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
-	if exists {
-		klog.V(3).InfoS("Get Kubernetes cluster certificate authority data from member cluster status", "clusterProfile", klog.KObj(cp))
-		cp.Status.AccessProviders[0].Cluster.CertificateAuthorityData = []byte(certificateAuthorityData.Value)
+
+	// Add cluster access provider, if and only if a cluster entry point and the CA data exist as part of the
+	// cluster properties.
+	clusterEntrypoint, entryPtExists := mc.Status.Properties[propertyprovider.ClusterEntryPointProperty]
+	caData, caDataExists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
+	if entryPtExists && caDataExists && len(clusterEntrypoint.Value) > 0 && len(caData.Value) > 0 {
```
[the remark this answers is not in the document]
michaelawyu: This behavior is consistent with the original setup; for simplicity reasons this controller passes the data as it is (same for the CA data). It might make sense to add CEL-based check on the upstream API definition, but not here.

sjwaight (dismissed):
Assuming E2Es pass, LGTM.

britaniar (approved):

pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -196,27 +197,26 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
...
-	}
-	// Get the CA Data
-	certificateAuthorityData, exists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
-	if exists {
-		klog.V(3).InfoS("Get Kubernetes cluster certificate authority data from member cluster status", "clusterProfile", klog.KObj(cp))
-		cp.Status.AccessProviders[0].Cluster.CertificateAuthorityData = []byte(certificateAuthorityData.Value)
+
+	// Add cluster access provider, if and only if a cluster entry point and the CA data exist as part of the
+	// cluster properties.
+	clusterEntrypoint, entryPtExists := mc.Status.Properties[propertyprovider.ClusterEntryPointProperty]
```
ytimocin, line 203: Can `mc.Status.Properties[propertyprovider.ClusterEntryPointProperty]` ever throw nil pointer exception?

pkg/controllers/clusterinventory/clusterprofile/controller.go
```diff
@@ -196,27 +197,26 @@ func (r *Reconciler) fillInClusterStatus(mc *clusterv1beta1.MemberCluster, cp *c
...
+	caData, caDataExists := mc.Status.Properties[propertyprovider.ClusterCertificateAuthorityProperty]
```
ytimocin, line 204: Same question for this one too.

test/e2e/join_and_leave_test.go
```diff
@@ -462,14 +461,10 @@ var _ = Describe("Test member cluster join and leave with clusterProfile", Label
...
-						return fmt.Errorf("cluster profile %s has no access providers %+v", cp.Name, cp.Status.AccessProviders)
-					}
-					if cp.Status.AccessProviders[0].Name != controller.ClusterManagerName {
-						return fmt.Errorf("cluster profile %s access provider name %s doesn't match expected %s", cp.Name, cp.Status.AccessProviders[0].Name, controller.ClusterManagerName)
-					}
-					if len(cp.Status.AccessProviders[0].Cluster.CertificateAuthorityData) == 0 {
-						return fmt.Errorf("cluster profile %s access provider certificate authority data should not be empty", allMemberClusterNames[idx])
+					if len(cp.Status.AccessProviders) != 0 {
+						// Note: at this moment the Azure property provider does not expose cluster FQDNs, and
+						// as a result no access provider will be populated.
```
ytimocin, line 466: If case checks if **AccessProviders** array has items, right? The comment sounds a bit off to me. Am I missing something?

michaelawyu: Hi Yetkin! Yeah, about this one: the access provider, as the issuer contributor mentioned, needs to have a valid cluster entrypoint (API server URL) set. However, at this moment, even though we do have the cluster entrypoint property listed, it is never written/used; as a result, all the cluster profiles we produce have invalid access providers with no entrypoints.

With this PR we have added the check as requested, i.e., if there is no cluster entrypoint, we do not populate access providers; until we add support for the cluster entrypoint property, the expected behavior onwards is that all the cluster profiles hub agent produces will have no access providers, hence the updated check.

pkg/controllers/clusterinventory/clusterprofile/controller.go
michaelawyu: Hi Yetkin! `mc.Status` is a struct (empty if not written yet, but never nil), and nil maps (`.status.properties`) can be read as empty maps with no panic errors, so we should be good.

pkg/controllers/clusterinventory/clusterprofile/controller.go
michaelawyu: Hi Yetkin! mc.Status is a struct (empty if not written yet, but never nil), and nil maps (.status.properties) can be read as empty maps with no panic errors, so we should be good.

michaelawyu:
Hi all! I am merging this to unblock progress; if there is any question/concern, please let me know. Will address them in separate PRs.

PR #864, opened 2026-09-01
feat: [FEP-0001] add utilities for manipulating binding manager

michaelawyu:

This pull request introduces a new `bindingmanager` utility package for managing the binding manager role in placement policies, adds integration tests for this logic, and tightens validation on the `objectRefs` field in the `BindingManager` struct and CRDs.

Related to #786 .

I have:

- [x] ITs

Depends on #827 -> will rebase when it is merged.


pkg/v1/utils/bindingmanager/manager.go
```diff
@@ -0,0 +1,200 @@
...
+		if err := hubClient.Status().Update(ctx, placementPolicy); err != nil {
+			// Reset the binding manager to its previous state.
+			placementPolicy.GetStatus().BindingManager = bindingManagerCopy
+			return false, errors.NewAPIServerError(err, "failed to update placement policy status", false)
+		}
+		klog.V(2).InfoS("Successfully claimed the binding manager role", "placementPolicy", klog.KObj(placementPolicy), "controllerName", controllerName, "objectRef", objectRef)
+		return true, nil
+	}
+
+	if bindingManager.ControllerName != controllerName {
```
ytimocin, line 74: Only return in the file without an API call. A stale cache delays the claim a cycle. Add the dry-run like `RelinquishRoleFor` does, plus a case in the stale-view Context.

pkg/v1/utils/bindingmanager/manager.go
```diff
@@ -0,0 +1,200 @@
...
+		klog.V(2).InfoS("The binding manager role has already been claimed by another controller; should retry later",
+			"placementPolicy", klog.KObj(placementPolicy),
+			"currentControllerName", bindingManager.ControllerName,
+			"applicantControllerName", controllerName, "applicantObjectRef", objectRef)
+		return false, nil
+	}
+
+	found := false
+	for idx := range bindingManager.ObjectRefs {
+		if reflect.DeepEqual(bindingManager.ObjectRefs[idx], objectRef) {
```
ytimocin: `ObjectReference` is five strings; `==` here and at L151.

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -564,8 +564,9 @@ type BindingManager struct {
 	// A list of references to the objects that are currently managing the bindings for this placement,
 	// under the reconciliation of the specified controller.
 	//
-	// +kubebuilder:validation:Optional
-	ObjectRefs []ObjectReference `json:"objectRefs,omitempty"`
+	// +kubebuilder:validation:Required
```
ytimocin, line 567: Merges clean against main and #829. Whoever lands second re-runs `make manifests`.

ytimocin (dismissed):

apis/kubefleet.dev/placement/v1alpha1/placementpolicy_types.go
```diff
@@ -564,8 +564,9 @@ type BindingManager struct {
...
+	// +kubebuilder:validation:MinItems=1
+	ObjectRefs []ObjectReference `json:"objectRefs"`
```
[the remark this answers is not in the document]
michaelawyu: N/A -> tests will be added at a later stage. New APIs; no risk of regression.

pkg/v1/utils/bindingmanager/suite_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	placementv1alpha1 "github.com/kubefleet-dev/kubefleet/apis/kubefleet.dev/placement/v1alpha1"
+)
+
+// Note (chenyu1): this package uses envtest-based environment for testing purposes as some the ops in the logic,
```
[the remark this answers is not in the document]
michaelawyu: Done.

pkg/v1/utils/bindingmanager/manager.go
michaelawyu: Hi Yetkin! So far we have been following the if stale, requeue pattern pretty consistently in our controller code; through refreshing the state itself is not that complicated, handling this properly is a bit tricky, esp. when in the middle of a control loop (as there might already have been some states being calculated based on the stale view; the caller might be in a better position to decide what to do).

The rest of the returns are the same actually; the writes would fail (RV Conflict) if the view is stale, and the dry-runs are added to check if the view has become stale, rather than to correct the stale view.

pkg/v1/utils/bindingmanager/manager.go
michaelawyu: Sure, let me update that 👌 Thanks 🙏

michaelawyu:
Hi Yetkin! I've addressed the comments, PTAL. Thanks 🙏

ytimocin (approved):

PR #866, opened 2026-09-01
fix: fail closed for namespace guardrail webhook

weng271190436:

Changes only `fleet.namespace.guardrail.validating` from `failurePolicy: Ignore` to `failurePolicy: Fail`. This prevents reserved Namespace CREATE, UPDATE, and DELETE requests from being admitted when the guardrail webhook is unavailable or times out.

The regression test verifies the namespace guardrail is fail-closed and all five other guardrail webhooks remain fail-open.

I have:

- `go test ./pkg/webhook -count=1`
- `make reviewable`: formatting, vet, and fast lint passed. Staticcheck initially used a stale Go 1.25-built cached binary and failed against the Go 1.26.6 module; after rebuilding Staticcheck with Go 1.26.6, `make staticcheck` passed.
- `git diff --check`

The scope is intentionally narrow: no other webhook failure policy, timeout, rule, or admission behavior changes.

### DIFF
--- a/.github/.copilot/breadcrumbs/2026-09-01-1819-fail-closed-reserved-namespace-webhook.md
+++ b/.github/.copilot/breadcrumbs/2026-09-01-1819-fail-closed-reserved-namespace-webhook.md
@@ -1,0 +1,72 @@
+# Fail closed reserved namespace webhook
+
+## Requirements
+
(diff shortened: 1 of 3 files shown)


michaelawyu:
Hi Wei! Just a note about this: the reason why it was set to fail-open before is that we had seen some alerts caused by this due to webhook connectivity issues (as the webhook also inspects heartbeats). Probably needs a shoutout in the notes just to let admins/adopters be prepared.

PR #867, opened 2026-09-01
fix: scope namespace-accessible placement status

weng271190436:

This change makes the namespaced `ClusterResourcePlacementStatus` a namespace-scoped projection instead of a near-verbatim copy of cluster-scoped placement status.

It:
- includes only the selected Namespace and resources contained in that namespace
- retains only same-namespace ResourceOverride and ResourceEnvelope references
- omits cluster-scoped override and envelope references
- preserves retained drift/diff values and condition messages unchanged
- keeps the full cluster-scoped ClusterResourcePlacement status unchanged
- updates the served API documentation and generated CRD descriptions

Value redaction is intentionally handled separately in independent PR #886.

- `go test ./pkg/controllers/placement -run TestBuildNamespaceAccessiblePlacementStatus -count=1`
- `go test ./pkg/controllers/placement -run TestSyncClusterResourcePlacementStatusSanitizesNamespaceView -count=1`
- `go test ./test/utils/crpstatussync -count=1`
- `make manifests`
- `git diff --check`

Tests used the Go 1.26.6 toolchain required by `go.mod`. The full placement envtest suite was not run because `/usr/local/kubebuilder/bin/etcd` is unavailable in this environment.

The namespace projection is intentionally applied only at the CRP-to-CRPS trust boundary. Privileged cluster-scoped status remains unchanged. The shared integration helper independently constructs and compares the expected projection so retained in-scope data remains covered.

### DIFF
--- a/pkg/controllers/placement/namespace_status_sync.go
+++ b/pkg/controllers/placement/namespace_status_sync.go
@@ -79,13 +79,38 @@ func isNamespaceAccessibleCRP(placementObj placementv1beta1.PlacementObj) bool {
 	return true
 }
 
-// filterStatusSyncedCondition creates a copy of the placement status with the
-// ClusterResourcePlacementStatusSynced condition removed. This is used when
-// creating ClusterResourcePlacementStatus objects because the StatusSynced
-// condition is only relevant for the main CRP object, not the CRPS object.
-// Returns a filtered copy of the status with all conditions except StatusSynced.
-func filterStatusSyncedCondition(status *placementv1beta1.PlacementStatus) *placementv1beta1.PlacementStatus {
+// buildNamespaceAccessiblePlacementStatus creates a namespace-scoped copy of the placement status.
+// It excludes resources and references outside the target namespace.
+func buildNamespaceAccessiblePlacementStatus(status *placementv1beta1.PlacementStatus, targetNamespace string) *placementv1beta1.PlacementStatus {
 	filteredStatus := status.DeepCopy()
+
+	filteredStatus.SelectedResources = filterResourceIdentifiersByNamespace(filteredStatus.SelectedResources, targetNamespace)
+	for idx := range filteredStatus.PerClusterPlacementStatuses {
+		clusterStatus := &filteredStatus.PerClusterPlacementStatuses[idx]
+		clusterStatus.ApplicableResourceOverrides = filterNamespacedNamesByNamespace(clusterStatus.ApplicableResourceOverrides, targetNamespace)
+		clusterStatus.ApplicableClusterResourceOverrides = nil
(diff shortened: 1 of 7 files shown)


pkg/controllers/placement/namespace_status_sync.go
```diff
@@ -99,6 +108,119 @@ func filterStatusSyncedCondition(status *placementv1beta1.PlacementStatus) *plac
...
+			clearPatchDetailValues(placement.ObservedDiffs)
+			filtered = append(filtered, placement)
+		}
+	}
+	return filtered
+}
+
+func clearPatchDetailValues(details []placementv1beta1.PatchDetail) {
+	for idx := range details {
+		details[idx].ValueInMember = ""
```
michaelawyu: Hi Wei! For this part, just out of clarity reasons, maybe it's better to say `(redacted)` or any message with that can explain the reason why the info is unset?

michaelawyu (commented):
Just a minor comment. Other parts LGTM ;)

pkg/controllers/placement/namespace_status_sync.go
weng271190436: Thanks Michael. I moved value redaction into independent PR #886 and used the existing explanatory marker `(redacted for security reasons)` there. PR #867 now focuses only on the namespace-to-cluster boundary.

pkg/controllers/placement/namespace_status_sync.go
```diff
@@ -99,6 +106,103 @@ func filterStatusSyncedCondition(status *placementv1beta1.PlacementStatus) *plac
...
+// otherwise namespace-scoped resource identifier.
+func sanitizeResourceIdentifier(identifier placementv1beta1.ResourceIdentifier, targetNamespace string) placementv1beta1.ResourceIdentifier {
+	if identifier.Envelope != nil &&
+		(identifier.Envelope.Type == placementv1beta1.ClusterResourceEnvelopeType || identifier.Envelope.Namespace != targetNamespace) {
+		identifier.Envelope = nil
+	}
+	return identifier
+}
+
+func filterResourceIdentifiersByNamespace(identifiers []placementv1beta1.ResourceIdentifier, targetNamespace string) []placementv1beta1.ResourceIdentifier {
```
ytimocin, line 151: Not sure if we can create a generic function instead multiple functions that differ only by the first argument...

weng271190436: Did this for 3 of them `clusterStatus.FailedPlacements` `clusterStatus.DriftedPlacements` `clusterStatus.DiffedPlacements`.

The other 2 are a bit different.

PR #898, opened 2026-09-08
fix: bound both forms of the rolling update settings

ytimocin:

`maxUnavailable` and `maxSurge` are int-or-strings. The `Pattern` marker only constrains the string form, so a negative integer passed CRD validation; and the digit branch of the pattern was unbounded, so a digit string longer than an int32 passed too and failed in whatever consumed it later.

- Adds a CEL rule that keeps the integer form non-negative (`type(self) == int ? self >= 0 : true`).
- Caps the digit branch of the pattern at nine digits.
- Applies to CRP and RP, v1 and v1beta1; CRDs regenerated.

Split out of #829, where it rode along; it is a v1beta1 fix that should ship and backport on its own.

I have:

envtest table in `test/apis/placement/v1beta1/api_validation_integration_test.go`: integer 0 and 5 accepted, `-1` rejected by the CEL rule with the field named, `"-1"` and `"101%"` rejected by the pattern, nine digits accepted, ten rejected, and an unrelated update on an existing CRP does not re-validate the bounds. RP has no webhook, so it gets its own smaller table.


michaelawyu (approved):
LGTM ;)

PR #899, opened 2026-09-08
fix: judge the object a tombstone wraps rather than the tombstone

ytimocin:

A deletion the watch missed arrives at the resource watcher as a `cache.DeletedFinalStateUnknown` wrapping the object's final state. `dynamicResourceFilter` derived a key from the tombstone itself, which fails, so every relist-detected deletion was silently dropped before the delete handler (which does unwrap tombstones) ever saw it. Unwrap first, then filter on the object.

Split out of #829.

I have:

Three table cases in `change_detector_test.go`: a tombstone wrapping a watched object passes, one wrapping an object in a skipped namespace is filtered, one wrapping garbage is filtered. The first fails without the fix.


michaelawyu (approved):
LGTM ;)

PR #901, opened 2026-09-08
feat: seed the cluster alias label on member clusters

ytimocin:

Adds `kubefleet.dev/cluster-alias`, a label on `MemberCluster` that names a cluster for placement by role rather than by name.

- The member cluster controller seeds it from the cluster name when the label is absent, and never reasserts it, so an admin can move an alias to another cluster and a selector following the alias follows the role.
- The member cluster validating webhook returns an admission **warning** when another cluster already holds the alias. Not a denial: two holders is the normal middle state of a handoff. The check fails open if the list fails.
- Constants `KubeFleetPrefix` and `ClusterAliasLabel` next to `FleetPrefix` and `MemberNameLabel` in `apis/placement/v1beta1`, where the other MemberCluster labels live.
- The member cluster label guard (`shouldDenyLabelModification`) exempts `kubefleet.dev/` labels for service accounts. The hub agent is one and is not in `system:masters`, so without this its own seeding update would be denied wherever `denyModifyMemberClusterLabels` is on. The exemption is narrower than the existing any-user one for `kubernetes-fleet.io/` on purpose: the alias is a scheduling label the controller never reasserts, so a plain user under the guard must not be able to move it.

Split out of #829 per review: placement policies and future APIs select on this label, so it should not depend on annotation-based placement. #829 is stacked on this branch.

Upgrade note: every existing MemberCluster gets the label once. The scheduler's member cluster watcher treats a label change as a scheduling trigger, so expect one sweep of re-evaluation proportional to clusters times placements on the first reconcile after upgrade.

I have:

Unit tables: name and alias labels present (no update), alias renamed by an admin is left alone, no labels, name label missing, name label wrong, name correct and alias absent (alias alone drives the update), update error. Webhook: no alias, unique alias, alias held by another cluster warns, the holder does not warn about itself, many holders list three plus a count, a list error admits without a warning. Label guard: a service account may create a `kubefleet.dev/` label under the guard; a plain user may not change one.


pkg/webhook/membercluster/membercluster_validating_webhook.go
```diff
@@ -87,5 +89,55 @@ func (v *memberClusterValidator) Handle(ctx context.Context, req admission.Reque
 		klog.V(2).ErrorS(err, "Member cluster has invalid fields, request is denied", "operation", req.Operation, "memberCluster", mcObjectName)
 		return admission.Denied(err.Error())
 	}
-	return admission.Allowed("Member cluster has valid fields")
+
+	response := admission.Allowed("Member cluster has valid fields")
+	if warning := v.clusterAliasCollisionWarning(ctx, &mc); warning != "" {
+		response = response.WithWarnings(warning)
```
michaelawyu, line 95: Hi Yetkin! Just a note for this on the managed solution -> users won't be able to see this warning (as direct label ops are blocked; they must be set up via tags). The check might be best made optional in this scenario.

pkg/webhook/membercluster/membercluster_validating_webhook.go
```diff
@@ -87,5 +89,55 @@ func (v *memberClusterValidator) Handle(ctx context.Context, req admission.Reque
...
+// handoff passes through a state where two clusters share the alias. For the same reason a failure
+// to list the member clusters does not block the request -- an advisory check must not stand
+// between an admin and the cluster they are registering.
+func (v *memberClusterValidator) clusterAliasCollisionWarning(ctx context.Context, mc *clusterv1beta1.MemberCluster) string {
+	alias, ok := mc.Labels[placementv1beta1.ClusterAliasLabel]
+	if !ok || alias == "" {
+		return ""
+	}
+
+	memberClusterList := &clusterv1beta1.MemberClusterList{}
```
michaelawyu, line 116: Hi Yetkin! I know it's inevitable, but I just couldn't get it out of head that in larger fleets (our scalability goal is 1K) this can be expensive.

michaelawyu: Maybe we can set up a field-based index? It does not have to be done right now (not a blocker for this PR) -> a TO-DO would suffice.

Plus the index might be of use in the scheduling framework as well.

pkg/webhook/membercluster/membercluster_validating_webhook.go
```diff
@@ -87,5 +89,55 @@ func (v *memberClusterValidator) Handle(ctx context.Context, req admission.Reque
...
+		if other.Labels[placementv1beta1.ClusterAliasLabel] == alias {
+			holders = append(holders, other.Name)
+		}
+	}
+	if len(holders) == 0 {
+		return ""
+	}
+	// The message lists at most a few holders: it is read on a terminal, and the actionable half is
+	// the alias value and that someone else holds it, not an exhaustive roll call.
+	const maxListedHolders = 3
```
michaelawyu, line 137: Just a nit: how about breaking out of the for-loop when there are more than 3 matches? Not a blocker; I know the chances for this branch to trigger is low.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -283,16 +283,29 @@ func (r *Reconciler) ensureFinalizer(ctx context.Context, mc *clusterv1beta1.Mem
...
 	}
-	mc.Labels[placementv1beta1.MemberNameLabel] = mc.Name
 
-	klog.InfoS("Ensured the member cluster name label", "memberCluster", klog.KObj(mc))
+	if mc.Labels[placementv1beta1.MemberNameLabel] != mc.Name {
+		mc.Labels[placementv1beta1.MemberNameLabel] = mc.Name
+		changed = true
+	}
+
+	// The alias label is seeded from the cluster name, but only when it is absent. Unlike the name
```
michaelawyu, line 296: Hi Yetkin! This PR would take effect as soon as it checks in, right? Though I guess it should be fine 😂 It's just an additional label.

michaelawyu (commented):
Added a few comments, PTAL 🙏

apis/placement/v1beta1/commons.go
```diff
@@ -79,6 +79,16 @@ const (
 	// by name in ResourceOverride and ClusterResourceOverride via labelSelector.
 	MemberNameLabel = FleetPrefix + "member-name"
 
+	// KubeFleetPrefix is the prefix used for the labels/annotations of the kubefleet.dev APIs. Like
+	// FleetPrefix, it is reserved: users may not set unprefixed keys, and the member cluster label
+	// guard treats both prefixes alike.
```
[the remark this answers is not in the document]
ytimocin: Fixed. The comment now says prefixed keys are reserved for KubeFleet (unprefixed ones are the user's, per the Kubernetes convention), and spells out that the two prefixes are not exempted alike by the label guard: `kubernetes-fleet.io/` for any user, `kubefleet.dev/` for service accounts only, with `system:masters` bypassing the guard entirely.

pkg/webhook/membercluster/membercluster_validating_webhook.go
ytimocin: Fixed here rather than deferred. `v.client` is `mgr.GetClient()`, so the List is cache-served and costs no API call; the part that grows with the fleet is the copy, since a cached List deep-copies what it returns. controller-runtime applies the label selector *before* that copy, so selecting on the alias copies only the colliding clusters instead of the whole inventory on every member cluster admission. One line, no indexer needed for this call site. Still agree an alias index could be useful for the scheduling framework later.

pkg/webhook/membercluster/membercluster_validating_webhook.go
ytimocin: Done, slightly differently so the message stays exact: it now keeps at most three names while still counting every holder, so `and N more` is accurate even if the whole fleet carries the alias. Also dropped the preallocation that sized the slice to the inventory. Tests cover one, three, four and six holders.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
ytimocin: Yes, one label per existing MemberCluster on the first reconcile after upgrade. The one thing worth naming: the scheduler's member cluster watcher treats any label change as a scheduling trigger, so expect a single sweep of re-evaluation proportional to clusters x placements, then quiet. Nothing selects on the alias yet, so that sweep re-derives identical decisions. It's called out as an upgrade note in the PR description.

pkg/webhook/membercluster/membercluster_validating_webhook.go
ytimocin: Good point on the visibility. Leaving it always-on for now: the check is advisory and fail-open, so where users cannot set labels directly nobody sees the warning and nothing is blocked. The one caller it does reach there is the hub agent's own seeding update, which passes back through this webhook and discards warnings, and seeding is one-shot per cluster. A collision is still possible in that environment (tags eventually become labels), so an off switch would only guarantee silence rather than prevent it. Happy to put it behind a flag if you would rather have it off there; say the word and I will add it in this PR.

michaelawyu: Hi Yetkin! Yeah, a flag would be nice -> but to unblock progress maybe it's not that urgent. We can do that later I guess.

michaelawyu (approved):
LGTM ;)

Issue #924, opened 2026-09-17
[FEP-0001] Add staged update rollout support for placement policies


michaelawyu:
Enable staged update rollout support for the new placement APIs.

See FEP-0001 for more information.

PR #925, opened 2026-09-17
interface: [FEP-0001] add APIs for stage update run APIs for placement policies

michaelawyu:

Related to #924.

I have:

N/A

Largely mirroring the existing APIs, with some adjustments.


britaniar:
Looks like the CRDs are failing linter?

apis/kubefleet.dev/rollout/v1alpha1/stagedupdaterun_types.go
```diff
@@ -0,0 +1,331 @@
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="stagedUpdateStrategyName is immutable"
+	StagedUpdateStrategyName string `json:"stagedUpdateStrategyName"`
+
+	// Whether the staged update run is suspended. If set to true, KubeFleet will pause the rollout.
+	//
+	// If the staged update run is created with this field set to true, KubeFleet will initialize the staged update run,
+	// i.e., decide on the resource snapshot to roll out and determine the stages and the clusters in each stage,
+	// but will not start rolling out resource changes until the field is set back to false.
```
britaniar, line 137: Q: So it will be true while it is initialized until a user updates to false to start it. Then be a toggle for start/stop? (checking my understanding)

britaniar: Or, is it false during initialize and then be true when done initializing? Then from there user has to start by setting it to false.

apis/kubefleet.dev/rollout/v1alpha1/stagedupdaterun_types.go
```diff
@@ -0,0 +1,331 @@
...
+
+type StagedUpdateRunFailurePolicy struct {
+	// The maximum number of failures (e.g., clusters where the rollout fails) before KubeFleet stops the staged update run.
+	//
+	// The default value is 1.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
+	MaxFailureCount int32 `json:"maxFailureCount,omitempty"`
```
britaniar, line 159: if the MaxFailureCount is >1 and there are multiple clusters in different stages, what happens when 1 cluster fails in a stage? Do we continue to the next stage and won't fail until we hit the maxFailureCount? What is the behavior between stages? 
I am confused if this is cluster based, staged based, or any failure in staged update run?

britaniar (commented):
Clarifying questions left.

apis/kubefleet.dev/rollout/v1alpha1/stagedupdaterun_types.go
michaelawyu: Hi Britania!

Yeah, the default value for this field is `false`. And it's fully controlled by the user:

* If the run is created with `suspended=true`, we will do the initialization but no update is performed.
    * The user can later flip this to `false` to have the update started.
* If the run is created with `suspended=false`, we will do the initialization and have the update started right away.
    * The user can flip this to `true`; initialization will be completed anyway but update will be stopped, leaving the current state as it is.

apis/kubefleet.dev/rollout/v1alpha1/stagedupdaterun_types.go
michaelawyu: Hi Britania! Ah, you are right; this should be set on the stage level as a per-stage configuration. Let me edit it to correct things.

michaelawyu: Hi Britania! Currently this is set on the run-level, i.e., it tallies all failed clusters across all (completed/running) stages. Until the threshold is reached, we will keep progressing from one stage to the next (the before-, after- tasks are still honored of course).

We register cluster-level failures only; I have updated the name to be a bit more specific (`maxAllowedClusterFailures`).

It concerns only update failures, i.e., resources fail to be applied/fail to become available within the wait time. For simplicity we haven't added any guard against constant flipping yet.

michaelawyu: This kind of deviates from our current behavior, where:

a) we will keep retrying upon apply/availability check failures indefinitely (a stuck condition will appear after a fixed period of time);
b) a run is cancelled only when validation fails for the cluster (the binding is updated unexpectedly) or the run itself (list of target clusters changed, etc.).

w/ the new placement policy APIs the constant validation is no longer needed, and the failure policy is added to avoid the indefinite retries. Of course, upon unexpected situations (bindings are missing, etc.) things will still cancel.

michaelawyu: And Britania, do you think there's anything we should change about this? Is it better as a strategy level field, and/or something that should be set on individual stages? I keep the field here on the run object as it appears to be simpler to use/understand (i.e., different runs may share the same stage configuration but depending on their intention the failure policy might differ), but if there's anything you are feeling off about, please let me know 🙏

britaniar: I think it makes sense to be here. Thank you for clarifying and answering my questions!

This behavior is better in my opinion than the previous behavior of "stuck" and keep it from progressing.

britaniar (approved):
LGTM! I am liking the new approaches :)

PR #930, opened 2026-09-18
docs: make the SECURITY.md support window definitive

ytimocin:

Related: #693

The cadence and support-window questions tracked on #693 are settled: minors roughly every three months, and the two most recent minors (`N`/`N-1`) receive security patches. `VERSIONING.md` already states this as policy; `SECURITY.md` still said the project *targets* that window and carried a TODO pointing back at the open question.

- Rewrite the "Supported versions" paragraph as policy rather than intent.
- Narrow the top-of-file TODO to the two sections that are genuinely still provisional (response SLO, coordinated disclosure).

No table or SLO values change.

I have:

Docs only. Wording cross-checked against `VERSIONING.md` ("Release cadence and supported versions") so the two files say the same thing.

One-paragraph change. The response-SLO and disclosure sections are deliberately left as-is; they have their own bullet on #693.


SECURITY.md
```diff
@@ -14,10 +13,10 @@ below to report it to the KubeFleet team.
 
 ## Supported versions
 
-KubeFleet is pre-1.0 and *targets* an `N`/`N-1` support window: the latest minor release and
-the one immediately preceding it receive security patches. The project has maintained a roughly
-2–3 month minor-release cadence since `v0.2`, giving approximately four to six months of patch
-coverage from the GA of any given minor. Minor cadence slippage is possible while we are pre-1.0.
+KubeFleet supports an `N`/`N-1` window: the latest minor release and the one immediately
+preceding it receive security patches. Minor releases are cut roughly every three months, so a
+minor receives patches for approximately six months from its GA. Exact dates may slip while the
```
sjwaight: I would consider defining what "GA" is, or use a different term.

sjwaight (changes requested):
Make the suggested change, and good to go.

